PR #3, opened 2025-05-14
fix: drift in BaseAgentExecutor and uses it in examples

codefromthecrypt:
This fixes drift in BaseAgentExecutor and uses it in examples. Ideally, this can centralize how cancel is to be marked unimplemented in a consistent way.


codefromthecrypt:
@holtskinner this has your change in #2, but has some lint fail unrelated. do you want to correct that in main? or should I add a commit to do that here?

holtskinner:
I'll fix the lint issues in main. Don't worry about them here.

> @holtskinner this has your change in #2, but has some lint fail unrelated. do you want to correct that in main? or should I add a commit to do that here?

kthota-g:
BaseAgentExecutor is a leftover class from previous iterations, this class will be removed.

codefromthecrypt:
Thanks for the tips

examples/google_adk/calendar_agent/adk_agent_executor.py
```diff
@@ -221,10 +218,6 @@ async def execute(
         )
         logger.debug('[Calendar] execute exiting')
 
-    async def cancel(self, context: RequestContext, event_queue: EventQueue):
-        # Ideally: kill any ongoing tasks.
-        raise ServerError(error=UnsupportedOperationError())
```
codefromthecrypt, line 226: @kthota-g is this the correct way to handle UnsupportedOperationError vs the event_queue?

kthota-g: yes, the latest version would handle throwing exceptions from agent executor

PR #31, opened 2025-05-15
feat: Add functionality for extended agent card. 

dmandar:
Does not support authentication yet. The helloworld sample adds an extended skill that is only available in the extended agent card. Note that the extended card endpoint is expected to be authenticated (once auth changes are added).

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -158,7 +158,7 @@ class TestA2ACardResolver:
         )
         
         # Fetch the extended card by providing its relative path and example auth
-        auth_kwargs = {"headers": {"Authorization": "Bearer testtoken"}}
+        auth_kwargs = {"headers": {"Authorization": "Bearer test token"}}
         agent_card_result = await resolver.get_agent_card(
             relative_card_path=self.EXTENDED_AGENT_CARD_PATH,
             http_kwargs=auth_kwargs


src/a2a/client/client.py
```diff
@@ -34,30 +33,92 @@ def __init__(
...
+            ) from e
         except httpx.RequestError as e:
             raise A2AClientHTTPError(
-                503, f'Network communication error: {e}'
+                503,
+                f'Network communication error fetching public agent card from {public_card_url}: {e}',
             ) from e
 
+        # Check for supportsAuthenticatedExtendedCard
+        if agent_card.supportsAuthenticatedExtendedCard:
```
swapydapy: can we add tests for this new functionality

src/a2a/types.py
```diff
@@ -710,7 +710,10 @@ class AgentCard(BaseModel):
     """
     The version of the agent - format is up to the provider.
     """
-
+    supportsAuthenticatedExtendedCard: Optional[bool] = Field(default=None)
```
kthota-g: This is a generated file, type changes should be made here.

src/a2a/server/apps/starlette_app.py
```diff
@@ -247,25 +245,74 @@ async def event_generator(
 
     async def _handle_get_agent_card(self, request: Request) -> JSONResponse:
         """Handles GET requests for the agent card."""
+        # Construct the public view of the agent card.
+        public_card_data = {
```
kthota-g: How is agent card different from a public card?

src/a2a/types.py
```diff
@@ -1408,7 +1408,7 @@ class AgentCard(BaseModel):
     """
     The version of the agent - format is up to the provider.
     """
-    supportsAuthenticatedExtendedCard: Optional[bool] = Field(default=None)
+    supportsAuthenticatedExtendedCard: bool | None = Field(default=None)
```
swapydapy, line 1419: The types changes need to be made in types.ts, then spec.json to be updated. Then the SDK repo should be updated with latest types. We should refer from there.

swapydapy (changes requested):

src/a2a/client/client.py
dmandar: Added.

src/a2a/types.py
dmandar: Done.

src/a2a/server/apps/starlette_app.py
dmandar: I am now passing a separate authenticated card directly to the starlette app.

src/a2a/client/client.py
```diff
@@ -66,20 +64,90 @@ async def get_agent_card(
...
```
kthota-g: The auth credentials are only required for authenticated card, but the implementation requires sending the credentials when invoking get_agent_card so it can internal invoke the authenticated card url in the same request. Since both the methods are anyway returning AgentCard, do we need the inner second request?

I am thinking the client should be able to do this check, change the card url and invoke with appropriate http headers. Probably the default card_url can be moved from the constructor to the get_agent_card method.

dmandar: Not quite, the thought is that agent cards will be public (unless an org needs them behind auth for its own private suite of agents). Based on the securitySchemes in the agentCard, the client will authenticate and then fetch the authenticated extended agent card. This will be handled as part of the overall auth changes for the SDK.

kthota-g: I understood the flow, but can't we use the same get_agent_card method, something like this?

```
resolver = A2ACardResolver()
//fetch the public card
public_card: AgentCard = resolver.get_agent_card()

if public_card.supportsAuthenticatedExtendedCard:
  //look at securitySchema and generate the auth credentials

  authenticated_card: AgentCard = resolver.get_agent_card(card_url="/agent/authenticatedExtendedCard", http_kwargs={"Authorization":"Bearer foo"})
```

dmandar: (Discussed offline with kthota-g) Have made the recommended changes. The client now needs to explicitly make a call for the extended agent card if one is available. This is no longer handled withing the A2ACardResolver

src/a2a/client/client.py
```diff
@@ -35,6 +29,7 @@ def __init__(
         httpx_client: httpx.AsyncClient,
         base_url: str,
         agent_card_path: str = '/.well-known/agent.json',
+        extended_agent_card_path: str = '/agent/authenticatedExtendedCard',
```
kthota-g: I don't see any usage for extended_agent_card_path, this attribute can be removed

src/a2a/client/client.py
```diff
@@ -119,14 +149,17 @@ async def get_client_from_agent_card_url(
         agent_card_path: str = '/.well-known/agent.json',
         http_kwargs: dict[str, Any] | None = None,
     ) -> 'A2AClient':
-        """Fetches the AgentCard and initializes an A2A client.
+        """Fetches the public AgentCard and initializes an A2A client.
+
+        This method will always fetch the public agent card. If an authenticated
+        or extended agent card is required, the A2ACardResolver should be used
+        directly to fetch the specific card, and then the A2AClient should be
+        instantiated with it.
```
kthota-g, line 155: http_kwargs seems to be accidentally removed from docstring

dmandar: done

src/a2a/client/client.py
dmandar: done.

src/a2a/types.py
dmandar: Done.

examples/helloworld/__main__.py
```diff
@@ -17,25 +21,49 @@
...
         defaultOutputModes=['text'],
         capabilities=AgentCapabilities(streaming=True),
-        skills=[skill],
+        skills=[skill],  # Only the basic skill for the public card
+        supportsAuthenticatedExtendedCard=True,
+    )
+
+    # This will be the authenticated extended agent card
+    # It includes the additional 'extended_skill'
+    specific_extended_agent_card = AgentCard(
```
swapydapy: Nit: Instead of manually copying, can we use model_copy feature of Pydantic: https://docs.pydantic.dev/latest/concepts/serialization/#model_copy

dmandar: sg

src/a2a/client/client.py
```diff
@@ -48,11 +42,19 @@ def __init__(
         self.httpx_client = httpx_client
 
     async def get_agent_card(
-        self, http_kwargs: dict[str, Any] | None = None
+        self,
+        relative_card_path: str | None = None,
```
swapydapy, line 46: This param is aimed at supporting the authenticated Card.

Should we just change it to get_authenticated_card: bool = False. I feel this would increase the readability by quickly conveying the expected behaviour.

examples/helloworld/test_client.py
```diff
@@ -1,19 +1,77 @@
...
+                    http_kwargs={"headers": auth_headers_dict}
+                )
+                logger.info("Successfully fetched authenticated extended agent card:")
+                logger.info(extended_agent_card.model_dump_json(indent=2, exclude_none=True))
+            else:
+                logger.info("\nPublic card does not support authenticated extended card, or public card not fetched.")
+
+        except Exception as e:
+            logger.error(f"Error during agent card fetching: {e}", exc_info=True)
+            # If public card fetching failed, or extended card fetching failed after public card indicated support,
```
swapydapy: we should raise an exception and error out, right?

examples/helloworld/test_client.py
```diff
@@ -1,19 +1,77 @@
...
+            # we might not have a card to use.
+
+        # Determine which card to use and Initialize Client
+        if extended_agent_card:
```
swapydapy: This whole block can be avoided. Just assign final_agent_card_to_use to which card was fetched. Makes it simple to follow control flow without so many if-else branches.

src/a2a/client/client.py
dmandar: it could also be used for fetching the public agent card from a non-default URL.

src/a2a/server/apps/starlette_app.py
```diff
@@ -320,25 +322,52 @@ async def _handle_get_agent_card(self, request: Request) -> JSONResponse:
...
+        # If an explicit extended_agent_card is provided, serve that.
+        if self.extended_agent_card:
+            return JSONResponse(
+                self.extended_agent_card.model_dump(
+                    mode='json', exclude_none=True
+                )
+            )
+        # Otherwise, if supportsAuthenticatedExtendedCard is true but no specific
+        # extended card is set, serve the main agent_card.
         return JSONResponse(
```
swapydapy, line 346: I feel this should be a 404 error. My vote would be to raise errors instead of implicit default behaviours, unless its trivial.

src/a2a/server/apps/starlette_app.py
```diff
@@ -70,11 +69,14 @@ def __init__(
...
             http_handler: The handler instance responsible for processing A2A
               requests via http.
+            extended_agent_card: An optional, distinct AgentCard to be served
+              at the authenticated extended card endpoint.
             context_builder: The CallContextBuilder used to construct the
               ServerCallContext passed to the http_handler. If None, no
               ServerCallContext is passed.
         """
         self.agent_card = agent_card
+        self.extended_agent_card = extended_agent_card
```
swapydapy, line 67: Lets validate, if extended_agent_card is not None, when supportsAuthenticatedExtendedCard = True

examples/helloworld/test_client.py
dmandar: yeah..i was ambivalent on this one. There could be cases where a client is only authorized to use the public agent card and use skills approved in it. I think let's not throw for now.

examples/helloworld/test_client.py
dmandar: This is intended for logging and making it explicit for users. Since we don't throw, the extended agent card might not be valid in which case we default to the public card (which could be a valid case for some clients where they are only allowed to access a public skill).

src/a2a/server/apps/starlette_app.py
dmandar: sg

src/a2a/server/apps/starlette_app.py
dmandar: good point.

examples/helloworld/test_client.py
swapydapy: Nit: Yeah we can assign final_agent_card_to_use in the previous if-elses. Also, we have already put a lot of logging in the previous if-elses.

examples/helloworld/test_client.py
swapydapy: Got it, in that case we should handle the exception when we fetch extendedCard and ignore it. This exception should then be thrown.

examples/helloworld/test_client.py
dmandar: Done.

examples/helloworld/test_client.py
dmandar: done.

src/a2a/server/apps/starlette_app.py
```diff
@@ -320,25 +317,54 @@ async def _handle_get_agent_card(self, request: Request) -> JSONResponse:
...
+                self.extended_agent_card.model_dump(
+                    mode='json', exclude_none=True
+                )
+            )
+        # If supportsAuthenticatedExtendedCard is true, but no specific
+        # extended_agent_card was provided during server initialization,
+        # return a 404
+        return JSONResponse(
+            {'error': 'Authenticated extended agent card is supported but not configured on the server.'},
+            status_code=404,
```
swapydapy, line 348: you would have to _generate_error_response() to generate error in JSON-RPC compliant way.

@kthota-g could you advise ^^

swapydapy: I think just throw Internal error for now.
https://github.com/google/a2a-python/blob/main/src/a2a/types.py#L306

We can introduce a new specific error later.

dmandar: this is not a jsonRPC endpoint. It is a standard RESTful GET endpoint for simplicity and consistency with the public card retrieval.

swapydapy: Ah makes sense.

swapydapy (approved):

PR #34, opened 2025-05-16
fix: Rename type to kind

kthota-g:
# Description

- [ ] Follow the [`CONTRIBUTING` Guide](https://github.com/google/a2a-python/blob/main/CONTRIBUTING.md).
- [ ] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format)

### DIFF
--- a/development.md
+++ b/development.md
@@ -1,0 +1,7 @@
+## Type generation from spec
+
+<!-- TODO replace spec.json with the public url so we always get the latest version-->
+
+```bash
+uv run datamodel-codegen --input ./spec.json --input-file-type jsonschema --output ./src/a2a/types.py --target-python-version 3.10 --output-model-type pydantic_v2.BaseModel --disable-timestamp --use-schema-description --use-union-operator --use-field-description --use-default --use-default-kwarg --use-one-literal-as-default --class-name A2A --use-standard-collections
+```
--- a/development.md
+++ b/development.md
@@ -1,3 +1,5 @@
+# Development
+
 ## Type generation from spec
 
 <!-- TODO replace spec.json with the public url so we always get the latest version-->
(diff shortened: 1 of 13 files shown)


src/a2a/types.py
```diff
@@ -598,10 +602,6 @@ class TextPart(BaseModel):
     """
     Text content
     """
-    type: Literal['text'] = 'text'
```
holtskinner, line 601: I know that `type` is a reserved word in python, probably the reason for this change. I've seen other libraries use `type_` to get around this. Is there another reason for using `itemType` specifically?

holtskinner (changes requested):

src/a2a/types.py
kthota-g: there is no particular standard for specifying the discriminator. The attribute name varies across frameworks such as `type_`, `kind`, we just used `itemType`. Functionally there is no difference.

development.md
mikeas1, line 1: Two linting issues found in checks:

/github/workspace/development.md:1 MD041/first-line-heading/first-line-h1 First line in a file should be a top-level heading [Context: "## Type generation from spec"]
/github/workspace/development.md:7:3 MD047/single-trailing-newline Files should end with a single newline character

src/a2a/types.py
kthota-g: I just went ahead and renamed this attribute to `kind`.

development.md
kthota-g: updated

swapydapy (approved):

src/a2a/types.py
swapydapy: Really appreciate it. `kind` is much more generic. `itemType` implies that the object is within a context of "item" object.

PR #56, opened 2025-05-20
feat: Add request context builder with referenceTasks

swapydapy:
Based on proposed referenceTaskIds https://github.com/google/A2A/pull/596


src/a2a/server/tasks/task_updater.py
```diff
@@ -96,6 +96,5 @@ def new_agent_message(
             contextId=self.context_id,
             messageId=str(uuid.uuid4()),
             metadata=metadata,
-            final=final,
```
pstephengoogle, line 139: why did you remove this?

swapydapy: 'final' has been removed from Message object.
https://github.com/google/A2A/blob/main/types/src/types.ts#L248

pstephengoogle (approved):

PR #66, opened 2025-05-20
docs: remove final and metadata fields from docstring

martimfasantos:
# Description

This pull request adds two optional parameters - `final` and `metadata` - to the utility functions `new_agent_text_message` and `new_agent_parts_message`:

* **`final`** (`bool | None`): Indicates whether the message is the final one. Useful for managing control flow in streaming agent interactions.
* **`metadata`** (`dict[str, Any] | None`): Allows attaching additional information to a message, such as token counts, response time, or other relevant metrics, which is helpful for monitoring and debugging during development.

**NOTE:** These parameters were already documented but had not been included in the function signatures. This PR ensures that the implementation is aligned with the documentation and supports enhanced message handling for streaming agents.

---

- [x] Follow the [`CONTRIBUTING` Guide](https://github.com/google/a2a-python/blob/main/CONTRIBUTING.md).

### DIFF
--- a/src/a2a/utils/message.py
+++ b/src/a2a/utils/message.py
@@ -1,7 +1,6 @@
 """Utility functions for creating and handling A2A Message objects."""
 
 import uuid
-from typing import Any
 
 from a2a.types import (
     Message,
@@ -15,8 +14,6 @@ def new_agent_text_message(
     text: str,
     context_id: str | None = None,
     task_id: str | None = None,
-    final: bool | None = None,
-    metadata: dict[str, Any] | None = None,
 ) -> Message:
     """Creates a new agent message containing a single TextPart.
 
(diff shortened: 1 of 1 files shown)


kthota-g:
@martimfasantos types are created based on the types defined in the spec here - https://github.com/google/A2A/blob/f99d96a24547b09667696feb06f2278dc51144ec/types/src/types.ts#L248

Currently when serving a streaming request, the SSE connection terminates when a `Message` type object is encountered, so `final` is redundant.

While Message doesn't have metadata, message.parts can have metadata associated for each part. Does it help?

martimfasantos:
> @martimfasantos types are created based on the types defined in the spec here - [google/A2A@`f99d96a`/types/src/types.ts#L248](https://github.com/google/A2A/blob/f99d96a24547b09667696feb06f2278dc51144ec/types/src/types.ts#L248)
> 
> Currently when serving a streaming request, the SSE connection terminates when a `Message` type object is encountered, so `final` is redundant.
> 
> While Message doesn't have metadata, message.parts can have metadata associated for each part. Does it help?

Got it — I’ll go ahead and remove the incorrect docstring arguments then

PR #70, opened 2025-05-20
fix: Throw exception for task_id mismatches

kthota-g:
# Description

- [ ] Follow the [`CONTRIBUTING` Guide](https://github.com/google/a2a-python/blob/main/CONTRIBUTING.md).
- [ ] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format)


examples/langgraph/agent_executor.py
holtskinner, line 1: Should this file change be included?

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -278,27 +286,27 @@ async def on_message_send_stream(
...
-                    )
-                    try:
-                        created_task: Task = event
-                        await self._queue_manager.add(created_task.id, queue)
-                        task_id = created_task.id
-                    except TaskQueueExists:
-                        logging.info(
-                            'Multiple Task objects created in event stream.'
+                if isinstance(event, Task):
+                    if task_id != event.id:
```
swapydapy, line 290: - For first time message, there's no task_id specified by the client.
- So request_context has task_id = None
- task_id = cast(str, request_context.task_id), this assigns task_id as "None"
- Now if agent create and returns a Task, then task_id != event.id
- Hence, we need to use the actual created task_id

The current code actually has a bug and needs to be fixed. But this PR would result in exception being raised always.

swapydapy (changes requested):

examples/langgraph/agent_executor.py
kthota-g: yes, some clean up and organizing the imports.

src/a2a/server/request_handlers/default_request_handler.py
kthota-g: There is always as task_id generated by RequestContext if task_id is not set by the client. Check this one - https://github.com/google/a2a-python/blob/4ca8eff48d622606f80151c0c051a1106a1f4621/src/a2a/server/agent_execution/context.py#L54

The expectation is that executor must use the generated task_id for returning the responses.

swapydapy: Ah, I see.

pstephengoogle (approved):

swapydapy (approved):

holtskinner (approved):

PR #85, opened 2025-05-22
feat: Update to support python 3.12

pstephengoogle:
- add case statement to use the asyncio.Queue.shutdown method for 3.13+
- add special handling to allow for similar semantics as asyncio.Queue.shutdown for 3.12

Tested on multiple samples in the a2a repo and some examples in this repo

### DIFF
--- a/src/a2a/server/events/event_queue.py
+++ b/src/a2a/server/events/event_queue.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import sys
 
 from a2a.types import (
     A2AError,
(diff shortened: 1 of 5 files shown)


src/a2a/server/events/event_queue.py
```diff
@@ -66,6 +72,11 @@ async def dequeue_event(self, no_wait: bool = False) -> Event:
             asyncio.QueueEmpty: If `no_wait` is True and the queue is empty.
             asyncio.QueueShutDown: If the queue has been closed and is empty.
         """
+        async with self._lock:
```
kthota-g, line 89: if a consumer is already created and is waiting for events on await self.queue.get() (line 89), as I understand, that won't be unblocked with this approach. queue.shutdown seems to unblock those blocked consumers as well.

pstephengoogle: as discussed offlline - added detailed description for usage in the dequeue_event method description. Goal is to keep the existing semantics and control flows, but this requires users of the dequeue_event method to understand the gotchas in the case of python<=3.12 usage

holtskinner (approved):

PR #90, opened 2025-05-22
docs: Update README for Python 3.10+ support

holtskinner:


darkhaniop:
@holtskinner, sorry for commenting on a closed PR, just wanted to leave a quick pointer.

Looks like one file did not get updated to match this version "drop": [.python-version](https://github.com/google/a2a-python/blob/e0db20ffc20aa09ee68304cc7e2a67c32ecdd6a8/.python-version)
uv might complain about a "minimal version," but I have not tested it on a lower version.

darkhaniop:
@pstephengoogle, fyi see^^^ (just realized I should've tagged you instead in this one)

PR #94, opened 2025-05-22
feat: Introduce a ServerCallContext

mikeas1:
# Description

This introduces a `ServerCallContext` class that can be used for a few purposes:

1. Configuring SDK behavior for a single request (such as timeouts).
2. Plumbing user-data through SDK layers.
3. Providing common request-level metadata (such as authentication) through SDK layers.

For now, the ServerCallContext is extremely barebones: just a holder for a `state` property for storing arbitrary key-values. I plan to extend this soon with request User details, similar to [Starlette](https://www.starlette.io/authentication/#users).

The bulk of this change is just plumbing this parameter all over the place. That plumbing is the value-add: you can put stuff in this context and we'll shuttle it everywhere for you.

Alternatives to this design are:
1. Having some kind of IoC + DI container that manages plumbing data for you. This is probably how you'd want to do this in, e.g., C# or Java, where such systems are commonplace.
2. Using ContextVars. This works, but I find ContextVars much harder to reason about for testing -- they act as hidden dependencies for a module when used this way, so discovering all the necessary ContextVars is quite annoying.

- [X] Follow the [`CONTRIBUTING` Guide](https://github.com/google/a2a-python/blob/main/CONTRIBUTING.md).
- [X] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format)


src/a2a/server/apps/starlette_app.py
```diff
@@ -41,6 +42,29 @@
...
 
+class CallContextBuilder(ABC):
+    """A class for building ServerCallContexts using the Starlette Request."""
+
+    @abstractmethod
+    def build(self, request: Request) -> ServerCallContext:
+        """Builds a ServerCallContext from a Starlette Request."""
+
+
+class DefaultCallContextBuilder(CallContextBuilder):
```
pstephengoogle: Can we comment this as an example, not a solution? E.g. if the downstream handlers are decoupled from the transport protocol, putting transport specific information into the context will require them to be recoupled somehow

mikeas1: Sure, I can remove it for now. I will be introducing a default implementation again shortly in order to extract authentication details from the Starlette request. I initially implemented this because I wanted to be able to skip a lot of "if not null use the builder, otherwise pass none" code by guaranteeing this was always present.

That said, I fully expect that there will be good reasons why someone would want to pass through transport-specific information. I don't actually need it here, though, so it makes sense to remove it.

PR #98, opened 2025-05-22
Fix: fix hello world example for python 3.12

pstephengoogle:
There was a race condition based on when the generator yielded with respect to when the queue was closed.


pstephengoogle:
This fixes #98

mikeas1 (approved):

PR #118, opened 2025-05-27
fix: Example dependency changes

kthota-g:
- Remove examples as workspace members and regenerate lock files.
- examples now use the public sdk
- Commented out the extended agent card example since it will be part of 0.2.5
- Updates the required python lib to 3.10 for all examples

### DIFF
--- a/examples/helloworld/__main__.py
+++ b/examples/helloworld/__main__.py
@@ -39,30 +39,31 @@ if __name__ == '__main__':
         defaultOutputModes=['text'],
         capabilities=AgentCapabilities(streaming=True),
         skills=[skill],  # Only the basic skill for the public card
-        supportsAuthenticatedExtendedCard=True,
+        # supportsAuthenticatedExtendedCard=True,
     )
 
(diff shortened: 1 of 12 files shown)


examples/helloworld/__main__.py
```diff
@@ -39,30 +39,31 @@
...
-        update={
-            'name': 'Hello World Agent - Extended Edition', # Different name for clarity
-            'description': 'The full-featured hello world agent for authenticated users.',
-            'version': '1.0.1', # Could even be a different version
-            # Capabilities and other fields like url, defaultInputModes, defaultOutputModes,
-            # supportsAuthenticatedExtendedCard are inherited from public_agent_card unless specified here.
-            'skills': [skill, extended_skill],  # Both skills for the extended card
-        }
-    )
+    # specific_extended_agent_card = public_agent_card.model_copy(
```
kthota-g, line 47: Extended agent card is not available as of sdk v0.2.4. This will be uncommented after the next release.

PR #133, opened 2025-05-28
fix: Correctly adapt starlette BaseUser to A2A User

mikeas1:
# Description

This fixes using the Starlette `BaseUser` class as our `auth.User`. I was previously marking `BaseUser` as a virtual subclass of `auth.User`, but `auth.User` is not compatible (it's closer to the flask-login user).

### DIFF
--- a/src/a2a/client/client.py
+++ b/src/a2a/client/client.py
@@ -1,25 +1,36 @@
 import json
 import logging
+
 from collections.abc import AsyncGenerator
 from typing import Any
 from uuid import uuid4
 
 import httpx
+
 from httpx_sse import SSEError, aconnect_sse
 from pydantic import ValidationError
 
 from a2a.client.errors import A2AClientHTTPError, A2AClientJSONError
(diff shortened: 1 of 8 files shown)


src/a2a/server/apps/starlette_app.py
```diff
@@ -45,8 +45,20 @@
 
 logger = logging.getLogger(__name__)
 
-# Register Starlette User as an implementation of a2a.auth.user.User
-A2AUser.register(BaseUser)
+
+class StarletteUserProxy(A2AUser):
```
swapydapy, line 49: Can we add a test for this: https://github.com/google-a2a/a2a-python/blob/main/tests/server/test_integration.py

swapydapy (changes requested):

src/a2a/server/apps/starlette_app.py
mikeas1: Done.

PR #162, opened 2025-06-04
feat: Add gRPC server and client support

pstephengoogle:
Introduces gRPC support at the server and client interfaces.
- Does not require changes to the AgentExecutors, simply a different wrapper at the binary server level
- Example to follow in a2a-samples
- Will introduce spec change to enable announcement of supported transport(s) by url for an agent in a following PR


holtskinner:
On the spelling errors, there's a few spots where the docstring splits lines in the middle of a word.

```
delegat
dler
Messag
```

holtskinner:
Finally figured out how to get the linter to ignore the auto-generated pb2 files. Everything else should be valid. We can also exclude the remaining errors if those files are auto-generated.

kthota-g (approved):

PR #171, opened 2025-06-10
feat: Add middleware to the client SDK

dmandar:
This PR introduces a middleware framework to the a2a-python client SDK, which allows for various authentication mechanisms. The new AuthInterceptor automatically injects authentication credentials into outgoing requests based on the agent's defined security schemes. This change simplifies the process of handling different authentication methods like Bearer tokens and API keys for developers using the SDK.


holtskinner:
@dmandar Can you provide more info in the PR message as to what the purpose of this is?

.vscode/launch.json
```diff
@@ -25,7 +30,24 @@
...
+      "name": "Pytest All",
+      "type": "debugpy",
+      "request": "launch",
+      "module": "pytest",
+      "args": [
+        "-v",
+        "-s"
+      ],
+      "console": "integratedTerminal",
+      "justMyCode": true // Or false if you need to step into library code
```
holtskinner:
```suggestion
      "justMyCode": true
```
Comments don't work in JSON

src/a2a/client/__init__.py
```diff
@@ -1,5 +1,9 @@
+# a2a/client/__init__.py
+
```
holtskinner:
```suggestion
```

src/a2a/client/auth/credentials.py
```diff
@@ -0,0 +1,51 @@
+# a2a/client/auth/credentials.py
```
holtskinner: Remove these filename comments from all files

holtskinner (changes requested):

src/a2a/client/auth/credentials.py
```diff
@@ -0,0 +1,47 @@
...
+        self,
+        security_scheme_name: str,
+        context: ClientCallContext | None,
+    ) -> str | None:
+        if not context or 'sessionId' not in context.state:
+            return None
+        session_id = context.state['sessionId']
+        return self._store.get(session_id, {}).get(security_scheme_name)
+
+    async def set_credential(
```
holtskinner: Should this be `set_credentials()` (ending "s") to match the `get_credentials()` method name and the unit tests?
```suggestion
    async def set_credentials(
```

src/a2a/client/auth/credentials.py
```diff
@@ -0,0 +1,47 @@
...
+
+class InMemoryContextCredentialStore(CredentialService):
+    """A simple in-memory store for session-keyed credentials.
+
+    This class uses the 'sessionId' from the ClientCallContext state to
+    store and retrieve credentials...
+    """
+
+    def __init__(self) -> None:
+        # {session_id: {scheme_name: credential}}
```
holtskinner: Remove commented out code, use a docstring

src/a2a/client/auth/user.py
holtskinner, line 1: Is this file needed? Can you use the existing file `src/a2a/auth/user.py`

src/a2a/client/middleware.py
```diff
@@ -0,0 +1,50 @@
...
+    which is ideal for concerns like authentication, logging, or tracing.
+    """
+
+    @abstractmethod
+    async def intercept(
+        self,
+        method_name: str,
+        request_payload: dict[str, Any],
+        http_kwargs: dict[str, Any],
+        agent_card: 'AgentCard | None',
```
holtskinner: Is there a reason these are using strings instead of the actual type classes?

src/a2a/client/client.py
```diff
@@ -327,6 +365,16 @@ async def get_task(
...
             request.id = str(uuid4())
 
+        # Apply interceptors before sending
+        payload, modified_kwargs = await self._apply_interceptors(
+            'tasks/get',
+            request.model_dump(mode='json', exclude_none=True),
+            http_kwargs,
+            context,
+        )
+        response_data = await self._send_request(payload, modified_kwargs)
```
holtskinner: I'm seeing this same pattern for most of the calls to `self._apply_interceptors()` then `self._send_request()` would it make any sense to add the call to `self._apply_interceptors()` in the `self._send_request()` method so it doesn't need to be duplicated?

src/a2a/client/auth/interceptor.py
```diff
@@ -0,0 +1,83 @@
...
+        request_payload: dict[str, Any],
+        http_kwargs: dict[str, Any],
+        agent_card: AgentCard | None,
+        context: ClientCallContext | None,
+    ) -> tuple[dict[str, Any], dict[str, Any]]:
+        if (
+            not agent_card
+            or not agent_card.security
+            or not agent_card.securitySchemes
+        ):
```
holtskinner:
```suggestion
        if not all((agent_card, agent_card.security, agent_card.securitySchemes)):
```

src/a2a/client/auth/interceptor.py
```diff
@@ -0,0 +1,83 @@
...
+            return request_payload, http_kwargs
+
+        for requirement in agent_card.security:
+            for scheme_name in requirement:
+                credential = await self._credential_service.get_credentials(
+                    scheme_name, context
+                )
+                if credential and scheme_name in agent_card.securitySchemes:
+                    scheme_def_union = agent_card.securitySchemes[scheme_name]
```
holtskinner:
```suggestion
                    scheme_def_union = agent_card.securitySchemes.get(scheme_name)
```

src/a2a/client/auth/interceptor.py
```diff
@@ -0,0 +1,83 @@
...
+                        return request_payload, http_kwargs
+                    if isinstance(scheme_def, APIKeySecurityScheme):
+                        if scheme_def.in_ == In.header:
+                            headers[scheme_def.name] = credential
+                            logger.debug(
+                                f"Added API Key Header for scheme '{scheme_name}'."
+                            )
+                            http_kwargs['headers'] = headers
+                            return request_payload, http_kwargs
+                        # Note: API keys in query or cookie are not handled here.
```
holtskinner:
```suggestion
                match scheme_def:
                    # Case 1: Bearer token schemes (HTTP Bearer, OAuth2, OIDC)
                    case (
                        HTTPAuthSecurityScheme() if scheme_def.scheme.lower() == "bearer"
                    ) | OAuth2SecurityScheme() | OpenIdConnectSecurityScheme():
                        headers["Authorization"] = f"Bearer {credential}"
                        logger.debug(
                            f"Added Bearer token for scheme '{scheme_name}' (type: {scheme_def.type})."
                        )
                        return request_payload, http_kwargs

                    # Case 2: API Key in Header
                    case APIKeySecurityScheme(in_=In.header):
                        headers[scheme_def.name] = credential
                        logger.debug(
                            f"Added API Key Header for scheme '{scheme_name}'."
                        )
                        return request_payload, http_kwargs
                
                # Note: Other cases like API keys in query/cookie are not handled and will be skipped.
```

Could make sense to use `match` here to simplify logic.

holtskinner:
And be sure to resolve lint errors: https://github.com/google-a2a/a2a-python/actions/runs/15591061140/job/43910003200?pr=171

mikeas1 (approved):
LGTM once Holt's comments are addressed.

src/a2a/client/client.py
dmandar: Not quite. The send_message_streaming would break if we did that.

src/a2a/client/middleware.py
dmandar: This was essentially for a forward reference. I included a __future__ as an alternative.

src/a2a/client/auth/credentials.py
dmandar: good catch! Looks like i was running an incorrect test file. Fixed along with test case updates.

src/a2a/client/auth/user.py
dmandar: Removed.

PR #174, opened 2025-06-10
fix: Update Proto Version to 6.31.1 for compatibility

### LINKED ISSUE #173: Protobuf version skew
I've been getting this warning: 

`env\Lib\site-packages\google\protobuf\runtime_version.py:98: UserWarning: Protobuf gencode version 5.29.3 is exactly one major version older than the runtime version 6.31.1 at a2a.proto. Please update the gencode to avoid compatibility violations in the next runtime release.`

I'm trying to run the A2A server with: 
`uv run . --host 0.0.0.0 --port 8080`

If I downgrade to 5.29.3, i get: 

`
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
grpcio-reflection 1.73.0 requires protobuf<7.0.0,>=6.30.0, but you have protobuf 5.29.5 which is incompatible.
grpcio-tools 1.73.0 requires protobuf<7.0.0,>=6.30.0, but you have protobuf 5.29.5 which is incompatible.
`

There's a version skew.

holtskinner:
Fixes #173


buf.gen.yaml
```diff
@@ -19,12 +19,12 @@ managed:
 plugins:
   # Generate python protobuf related code
   # Generates *_pb2.py files, one for each .proto
-  - remote: buf.build/protocolbuffers/python:v29.3
+  - remote: buf.build/protocolbuffers/python:v31.1
```
rajeshvelicheti: Remove the version may be? always use the greatest?

buf.gen.yaml
```diff
@@ -19,12 +19,12 @@ managed:
...
     out: src/a2a/grpc
   # Generate python service code.
   # Generates *_pb2_grpc.py
   - remote: buf.build/grpc/python
     out: src/a2a/grpc
   # Generates *_pb2.pyi files.
-  - remote: buf.build/protocolbuffers/pyi:v29.3
+  - remote: buf.build/protocolbuffers/pyi:v31.1
```
rajeshvelicheti: Remove the version may be? always use the greatest?

rajeshvelicheti (approved):

PR #175, opened 2025-06-10
fix: Change constant names to UPPER_CASE for Pylint compliance

kevinlu310:


kevinlu310:
Just realized lower_case names are common and idiomatic for projects using Pydantic models.

PR #199, opened 2025-06-16
feat: Convert fields in `types.py` to use `snake_case`

holtskinner:
Makes types more pythonic

Pydantic aliases can be used for initialization and are used automatically for serialization.
Added a workaround to support `camelCase` when setting specific attributes.

NOTE: In 0.3.0 and later, only snake_case will be supported for direct attribute access.


holtskinner:
\gemini review

darrelmiller:
Is there some discussion of this change? It doesn't seem to add much value and it causes confusion with the use of JSON Schema which uses camelCased keywords.

pyproject.toml
```diff
@@ -43,7 +43,7 @@ grpc = ["grpcio>=1.60", "grpcio-tools>=1.60", "grpcio_reflection>=1.7.0", "proto
 homepage = "https://a2a-protocol.org/"
 repository = "https://github.com/a2aproject/a2a-python"
 changelog = "https://github.com/a2aproject/a2a-python/blob/main/CHANGELOG.md"
-documentation = "https://a2a-protocol.org/latest/sdk/python/"
+documentation = "https://a2a-protocol.org/sdk/python/"
```
holtskinner:
```suggestion
documentation = "https://a2a-protocol.org/latest/sdk/python/"
```

PR #209, opened 2025-06-17
refactor: Add missing parameters from `JSONRPCApplication.build()`

holtskinner:
- Resolve JSCPD errors
- initializers from subclass should automatically call the initializer of the superclass, explicitly listing them shouldn't be needed.

### DIFF
--- a/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
+++ b/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
@@ -45,6 +45,7 @@ from a2a.types import (
 from a2a.utils.constants import (
     AGENT_CARD_WELL_KNOWN_PATH,
     DEFAULT_RPC_URL,
+    EXTENDED_AGENT_CARD_PATH,
 )
 from a2a.utils.errors import MethodNotImplementedError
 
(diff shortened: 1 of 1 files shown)


src/a2a/server/apps/jsonrpc/starlette_app.py
```diff
@@ -107,12 +57,12 @@ def add_routes_to_app(
             rpc_url: The URL path for the A2A JSON-RPC endpoint (POST requests).
             extended_agent_card_url: The URL for the authenticated extended agent card endpoint.
         """
-        routes = self.routes(
+        app_routes = self.routes(
```
holtskinner:
```suggestion
        routes = self.routes(
```

src/a2a/server/apps/jsonrpc/starlette_app.py
```diff
@@ -107,12 +57,12 @@ def add_routes_to_app(
...
             agent_card_url=agent_card_url,
             rpc_url=rpc_url,
             extended_agent_card_url=extended_agent_card_url,
         )
-        app.routes.extend(routes)
+        app.routes.extend(app_routes)
```
holtskinner:
```suggestion
        app.routes.extend(routes)
```

PR #212, opened 2025-06-18
refactor!: Push notification changes

kthota-g:
BEGIN_COMMIT_OVERRIDE
refactor!: Removes `push_notifier` interface from the SDK and introduces `push_notification_config_store` and `push_notification_sender` for supporting push notifications.

feat: Adds support for more than one `push_notification_config` per task.

feat: Adds support for List and Delete push notification configurations.

---

Following is an example for the changes required with new version:

Previously:

```py
from a2a.server.tasks import InMemoryPushNotifier
...
...
httpx_client = httpx.AsyncClient()
request_handler = DefaultRequestHandler(
    agent_executor=CurrencyAgentExecutor(),
    task_store=InMemoryTaskStore(),
    push_notifier=InMemoryPushNotifier(httpx_client),
)
...
...
```

Now:

```py
from a2a.server.tasks import InMemoryPushNotificationConfigStore, BasePushNotificationSender
...
...
httpx_client = httpx.AsyncClient()
push_notification_config_store = InMemoryPushNotificationConfigStore()
push_notification_sender = BasePushNotificationSender(httpx_client,
                        config_store=push_notification_config_store)
        
request_handler = DefaultRequestHandler(
    agent_executor=CurrencyAgentExecutor(),
    task_store=InMemoryTaskStore(),
    push_config_store=push_notification_config_store,
    push_sender=push_notification_sender
)
...
...
```

Release-As: 0.2.11

END_COMMIT_OVERRIDE


holtskinner:
@kthota-g Can you add this to the PR description with the version this should be for?

```
Release-As: 0.2.9
```

kthota-g:
> @kthota-g Can you add this to the PR description with the version this should be for?
> 
> ```
> Release-As: 0.2.9
> ```

sg. There are additional interface changes required to merge along with this PR. I will tag the release once the PR for other changes are in.

kthota-g:
@swapydapy can you review this one more time? The changes to support multiple configs are complete and have updated the description with changes required for migration.

src/a2a/server/tasks/inmemory_push_notification_config_store.py
```diff
@@ -13,34 +13,49 @@
...
         self, task_id: str, notification_config: PushNotificationConfig
     ) -> None:
         """Sets or updates the push notification configuration for a task in memory."""
         async with self.lock:
-            self._push_notification_infos[task_id] = notification_config
+            if task_id not in self._push_notification_infos:
+                self._push_notification_infos[task_id] = []
+            
+            if notification_config.id is None:
+                notification_config.id = task_id
```
swapydapy, line 35: Re-using task_id might cause confusion while working with task Data. Can't we create new configId and use that.

swapydapy: I understand this logic has been added to preserve backward compatibility. Instead, can we generate new configID and keep a mapping of defaultConfigId for each TaskId. That way, we get our backward compatiblity and generate nice unique Id for config.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -27,11 +27,15 @@ def __init__(self, httpx_client: httpx.AsyncClient, config_store: PushNotificati
...
     async def send_notification(self, task: Task) -> None:
         """Sends a push notification for a task if configuration exists."""
-        push_info = await self._config_store.get_info(task.id)
-        if not push_info:
+        push_configs = await self._config_store.get_info(task.id)
+        if not push_configs:
             return
-        url = push_info.url
+        
+        for push_info in push_configs:
```
swapydapy: Currently this is going to **sequentially** send out notifs. Can we improve this, by creating awaitables for each disptach and use asyncio gather.

https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -393,11 +397,11 @@ async def on_get_task_push_notification_config(
             raise ServerError(error=TaskNotFoundError())
 
         push_notification_config = await self._push_config_store.get_info(params.id)
-        if not push_notification_config:
+        if not push_notification_config or not push_notification_config[0]:
```
swapydapy, line 413: If there are not push configs, this is going to raise "index out of range". Should check for length and then get the first config.

Also, we should return 404, not found, error, if there are no push configs.

tests/server/request_handlers/test_default_request_handler.py
```diff
@@ -1146,3 +1149,281 @@ async def consume_stream():
...
+            params, create_server_call_context()
+        )
+    assert result == None
+
+@pytest.mark.asyncio
+async def test_delete_task_push_notification_config_info_with_config():
+    """Test on_list_task_push_notification_config with push config+id"""
+    mock_task_store = AsyncMock(spec=TaskStore)
+    
+    sample_task = create_sample_task(task_id='task_info_not_found')
```
swapydapy: nit: please fix task name. Its misleading. Also similar issue in other test task names.

tests/server/request_handlers/test_default_request_handler.py
```diff
@@ -1146,3 +1149,281 @@ async def consume_stream():
...
+    mock_task_store.get.return_value = sample_task
+
+    push_config = PushNotificationConfig(url='http://example.com')
+
+    # insertion without id should replace the existing config
+    push_store = InMemoryPushNotificationConfigStore()
+    await push_store.set_info('task_1', push_config)
```
swapydapy: We should use the set_info API, instead of directly calling the inner push_store, right?

tests/server/request_handlers/test_default_request_handler.py
```diff
@@ -1146,3 +1149,281 @@ async def consume_stream():
...
+    )
+
+    assert len(result) == 2
+    assert result[0].taskId == 'task_1'
+    assert result[0].pushNotificationConfig == push_config1
+    assert result[1].taskId == 'task_1'
+    assert result[1].pushNotificationConfig == push_config2
+
+@pytest.mark.asyncio
+async def test_list_task_push_notification_config_info_with_config_and_no_id():
```
swapydapy, line 1353: We should also add test to verify if set and get work with configId.

tests/server/request_handlers/test_default_request_handler.py
```diff
@@ -1146,3 +1149,281 @@ async def consume_stream():
...
+    mock_task_store = AsyncMock(spec=TaskStore)
+    
+    sample_task = create_sample_task(task_id='task_info_not_found')
+    mock_task_store.get.return_value = sample_task
+
+    push_config = PushNotificationConfig(url='http://example.com')
+
+    # insertion without id should replace the existing config
+    push_store = InMemoryPushNotificationConfigStore()
+    await push_store.set_info('task_1', push_config)
```
swapydapy, line 1545: We should use set_info handler API, instead of directly calling inner push_store.

swapydapy (changes requested):

src/a2a/server/tasks/inmemory_push_notification_config_store.py
kthota-g: we need a predictable id for configuration when an id is not provided and currently the grpc implementation is using the taskId here - https://github.com/a2aproject/a2a-python/blob/deeb62de8333d229529e0d59512ffa689f5cb21c/src/a2a/utils/proto_utils.py#L262 as the config id.

Given that this is for backward compatibility and the task_id to config_id is 1-1 mapped. I don't think we require to define another map and support it in all persistent storage types. Probably we can add a prefix "PUSH_" or hash the task_id as the default config id.

src/a2a/server/request_handlers/default_request_handler.py
kthota-g: if not push_notification_config evaluates to true for empty arrays, so it won't throw index out of bounds.

I will take a look at the error type to throw

tests/server/request_handlers/test_default_request_handler.py
kthota-g: updated

tests/server/request_handlers/test_default_request_handler.py
kthota-g: added tests

tests/server/request_handlers/test_default_request_handler.py
kthota-g: done

tests/server/request_handlers/test_default_request_handler.py
kthota-g: changed

src/a2a/server/request_handlers/default_request_handler.py
kthota-g: We can't throw 404 status code for JSON-RPC methods, the status code should be 200 with inner error code.

We don't have a specific error type for PushNotificationConfigNotFound in the spec among A2AError types. InternalError is the closest, I have updated the message to reflect the nature of the error.

I will take a look to add new error type in another PR part of the spec updates.

src/a2a/server/tasks/base_push_notification_sender.py
kthota-g: done

src/a2a/server/tasks/inmemory_push_notification_config_store.py
swapydapy: I guess we can leave it like this for now. The clients should anyway transparently use these config IDs.

PR #215, opened 2025-06-18
feat: raise error for tasks in terminal states

swapydapy:
- make tasks to be not restartable after they reach a terminal state.
- Based on discussion in https://github.com/google-a2a/A2A/discussions/723


kthota-g (approved):
Let's update the spec with the details that tasks in terminal state cannot be restarted.

Issue #101, opened 2025-05-23
[Feat]: make dependency starlette optional


sll521:
### Is your feature request related to a problem? Please describe.

we have a project used a lower version starlette>=0.37.2,<0.38.0, but the a2a-sdk use starlette>=0.46.2, confict! My current solution is to directly modify the dependent version using the source code, but upgrading later will be troublesom. If there is a better way, please advise.

### Describe the solution you'd like

I hope the SDK can provide a way to solve my problem, below are my some idea:
way 1.  use a lower verson of starlette,may cause incompatibility issue
way 2. make starlette optional, then offer an abstract application do not dependent specific web framework，let user extend 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

klauswong:
Facing the same issue where the dependencies conflict with Chainlit requirements
As a tmp solution, we forked the a2a-python and lowered the version requirement
https://github.com/klauswong/a2a-python/commit/f5f726395da15998cc56e9d87973e1a2ec6d702c

darkhaniop:
@holtskinner, do you mind if I try implementing the second solution instead (or after #193 if it gets merged in the meantime)?
> way 2. make starlette optional, then offer an abstract application do not dependent specific web framework，let user extend

If we opt for "way 2," we can make both `starlette` and the newly added `FastAPI` frameworks optional, so that downstream projects only add dependencies they actually use, instead of carrying all possible frameworks.

This could be a better approach. What do you think?

holtskinner:
@darkhaniop Go ahead and make a PR for it

PR #217, opened 2025-06-19
refactor!: Make the FastAPI dependency optional

darkhaniop:
# Description

With this update, using A2AFastAPIApplication requires either adding the FastAPI package directly to the project or indicating the "fastapi" group when installing a2a-sdk, e.g., with uv "uv add a2a-sdk[fastapi]" or with pip "pip install a2a-sdk[fastapi]"

## Additional Checks

- Verified that the "helloworld" agent sample works with this change.
- Verified that a reasonable error message is printed when trying to use `A2AFastAPIApplication`.

---

- [x] Follow the [`CONTRIBUTING` Guide](https://github.com/google-a2a/a2a-python/blob/main/CONTRIBUTING.md).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,6 @@ authors = [{ name = "Google LLC", email = "googleapis-packages@google.com" }]
 requires-python = ">=3.10"
 keywords = ["A2A", "A2A SDK", "A2A Protocol", "Agent2Agent"]
 dependencies = [
-    "fastapi>=0.115.2",
     "httpx>=0.28.1",
     "httpx-sse>=0.4.0",
     "google-api-core>=1.26.0",
@@ -80,6 +79,10 @@ dev = [
     "uv-dynamic-versioning>=0.8.2",
     "types-protobuf",
     "types-requests",
+    "fastapi>=0.115.2",
+]
+fastapi = [
+    "fastapi>=0.115.2",
 ]
 
 [[tool.uv.index]]
--- a/src/a2a/server/apps/jsonrpc/fastapi_app.py
+++ b/src/a2a/server/apps/jsonrpc/fastapi_app.py
@@ -2,8 +2,11 @@ import logging
 
 from typing import Any
 
-from fastapi import FastAPI, Request, Response
-
+from a2a.server.apps.jsonrpc.fastapi_import_helpers import (
+    FastAPI,
+    Request,
+    Response,
+)
 from a2a.server.apps.jsonrpc.jsonrpc_app import (
     CallContextBuilder,
     JSONRPCApplication,
--- a/src/a2a/server/apps/jsonrpc/fastapi_import_helpers.py
+++ b/src/a2a/server/apps/jsonrpc/fastapi_import_helpers.py
@@ -1,0 +1,32 @@
+"""Helper functions for handling optional FastAPI package imports."""
+
+try:
+    from fastapi import FastAPI, Request, Response
+except ImportError:
+    _FASTAPI_DEPENDENCY_ERROR_MSG = """
+    The A2A Python SDK FastAPI app requires the FastAPI package, which
+    is an optional dependency. To fix this issue, please add the fastapi
+    package to your project, e.g. by executing
+
+        uv add fastapi
+
+    or install the A2A SDK with the optional FastAPI dependency:
+
+        uv add a2a-sdk[fastapi]
+    """
+
+    class _DummyFastAPIClasses:
+        """Parent class for dummy fastapi.* class declarations."""
+
+        def __init__(self) -> None:
+            """Raises ImportError when initiating a dummy fastapi.* instance."""
+            raise ImportError(_FASTAPI_DEPENDENCY_ERROR_MSG)
+
+    class FastAPI(_DummyFastAPIClasses):
+        """A dummy fastapi.FastAPI declaration."""
+
+    class Request(_DummyFastAPIClasses):
+        """A dummy fastapi.Request declaration."""
+
+    class Response(_DummyFastAPIClasses):
+        """A dummy fastapi.Response declaration."""
--- a/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
+++ b/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
@@ -7,7 +7,6 @@ from abc import ABC, abstractmethod
 from collections.abc import AsyncGenerator
 from typing import Any
 
-from fastapi import FastAPI
 from pydantic import ValidationError
 from sse_starlette.sse import EventSourceResponse
 from starlette.applications import Starlette
@@ -17,6 +16,7 @@ from starlette.responses import JSONResponse, Response
 
 from a2a.auth.user import UnauthenticatedUser
 from a2a.auth.user import User as A2AUser
+from a2a.server.apps.jsonrpc.fastapi_import_helpers import FastAPI
 from a2a.server.context import ServerCallContext
 from a2a.server.request_handlers.jsonrpc_handler import JSONRPCHandler
 from a2a.server.request_handlers.request_handler import RequestHandler
--- a/uv.lock
+++ b/uv.lock
@@ -10,7 +10,6 @@ resolution-markers = [
 name = "a2a-sdk"
 source = { editable = "." }
 dependencies = [
-    { name = "fastapi" },
     { name = "google-api-core" },
     { name = "grpcio" },
     { name = "grpcio-reflection" },
@@ -28,6 +27,7 @@ dependencies = [
 [package.dev-dependencies]
 dev = [
     { name = "datamodel-code-generator" },
+    { name = "fastapi" },
     { name = "mypy" },
     { name = "pytest" },
     { name = "pytest-asyncio" },
@@ -38,10 +38,12 @@ dev = [
     { name = "types-requests" },
     { name = "uv-dynamic-versioning" },
 ]
+fastapi = [
+    { name = "fastapi" },
+]
 
 [package.metadata]
 requires-dist = [
-    { name = "fastapi", specifier = ">=0.115.2" },
     { name = "google-api-core", specifier = ">=1.26.0" },
     { name = "grpcio", specifier = ">=1.60" },
     { name = "grpcio-reflection", specifier = ">=1.7.0" },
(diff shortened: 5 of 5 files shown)


holtskinner:
This won't solve the core issue on its own, because starlette is also a dependency and it depends on fastapi, so you'll need to remove that dependency as well.

darkhaniop:
Hi @holtskinner,

> This won't solve the core issue on its own, because starlette is also a dependency and it depends on fastapi, so you'll need to remove that dependency as well.

It seems the dependencies are the other way around, so if we do it one framework at a time, I think refactoring FastAPI first makes sense.

### Starlette dependency tree

A quick dependency check shows that `starlette (also `sse-starlette`) depends only on the `anyio` package:
```
dummy-project v0.1.0
└── starlette v0.47.0
    └── anyio v4.9.0
        ├── idna v3.10
        └── sniffio v1.3.1
```

<details>
<summary>for reference, see FastAPI dependency tree here</summary>

### FastAPI dependency tree

On the other hand, if a project requires `fastapi`, the dependency tree would also include `starlette`:
```
another-dummy-project v0.1.0
└── fastapi v0.115.13
    ├── pydantic v2.11.7
    │   ├── annotated-types v0.7.0
    │   ├── pydantic-core v2.33.2
    │   │   └── typing-extensions v4.14.0
    │   ├── typing-extensions v4.14.0
    │   └── typing-inspection v0.4.1
    │       └── typing-extensions v4.14.0
    ├── starlette v0.46.2
    │   └── anyio v4.9.0
    │       ├── idna v3.10
    │       └── sniffio v1.3.1
    └── typing-extensions v4.14.0
```
</details>

## SDK development vs downstream project dependencies

I kept FastAPI in dev dependencies so that when **working on the SDK itself**, i.e., running `uv sync` in the SDK dir, all frameworks (both Starlette and FastAPI) would be installed in the environment.

At the same time, with this refactoring **non-FastAPI downstream projects** (agents) that use `a2a-sdk` should become lighter without FastAPI and its dependencies. And those that use (a potentially different version of) FastAPI would less likely to encounter dependency conflicts.

## Considerations for next steps (refactoring `starlette`)

In this PR, I propose refactoring only the FastAPI dependency, as it requires fewer changes. After that, we can look into making the Starlette package optional.

As I stated earlier (see my [comment](https://github.com/google-a2a/a2a-python/issues/101#issuecomment-2978930221) in #101), it would be nice to eventually make all frameworks (both FastAPI and Starlette) optional.

However, since all of the samples and documentation are written under the assumption that installing `a2a-sdk` adds Starlette, making it optional would introduce a bigger breaking change and require more involved refactoring and testing:
* Docs and tutorials would also need to be updated (would require a follow-up PR in [google-a2a/a2a](https://github.com/google-a2a/A2A)).
* Samples would have to be updated to directly require Starlette (would require follow-up PRs in [google-a2a/a2a-samples](https://github.com/google-a2a/a2a-samples)).

holtskinner (approved):

darkhaniop:
It appears that the unit test task is failing because coverage has dropped below 90%. Running unit tests locally shows that all tests pass with a threshold of 89%.

```
Required test coverage of 89% reached. Total coverage: 89.47%
======================= 461 passed, 3 warnings in 2.75s =======================
```

Should we consider lowering option `--cov-fail-under`?

darkhaniop (commented):
Diff in 16059fb0a2ff7784a2f66ba5ca91cf8c510e7e23
```diff
--- a/src/a2a/server/apps/jsonrpc/fastapi_app.py
+++ b/src/a2a/server/apps/jsonrpc/fastapi_app.py
@@ -58,17 +58,6 @@ class A2AFastAPIApplication(JSONRPCApplication):
             context_builder=context_builder,
         )
 
-        self._check_fastapi_dependency()
-
-    def _check_fastapi_dependency(self) -> None:
-        """Checks if the FastAPI package is installed.
-
-        If instead of the actual FastAPI class, a dummy implementation from
-        ./fastapi_import_helpers.py is imported, initializing FastAPI() would
-        raise ImportError.
-        """
-        _app = FastAPI()
-
     def add_routes_to_app(
         self,
         app: FastAPI,
```

I think this removed code was a better way to check the installed FastAPI framework vs. how it was grouped with Starlette in a later update in this PR. Because, in the current version, we again lose the capability to use the `a2a-sdk` package with only Starlette (as both framework imports are within the same try-except block, which would fail if only one of the frameworks is present).

Also, maybe checking for the presence of a particular framework in the `JSONRPCApplication` constructor (which I believe is intended to be framework-agnostic) is not the best approach. For instance, a developer may want to bring another framework and pass it using the `context_builder` param.

---

*edit: clarity and add the commit ref with diff*

darkhaniop:
Since all three apps `A2AStarletteApplication`, `A2AFastAPIApplication`, and `JSONRPCApplication` are a part of `a2a-sdk` public API, we should assume that any of them can be used directly by downstream projects. Therefore, proper ImportError messages should be printed when initializing each type of app with missing optional dependencies. So, I put `raise ImportError()` with corresponding messages in the three modules.

Also, as I explained above, it is better to decouple framework dependencies to allow the initialization of `JSONRPCApplication` with only the necessary framework installed (instead of requiring all supported frameworks to be present).

In the latest refactoring, instead of telling the developer that all frameworks should be added with `a2a-sdk[http-server]`
```python
        if not _http_server_installed:
            raise ImportError(
                'The `a2a-sdk[http-server]` package is required to use the `JSONRPCApplication`.'
            )
```

I updated it to provide clearer details:
```python
        if not _package_starlette_installed:
            raise ImportError(
                'Packages `starlette` and `sse-starlette` are required to use the'
                ' `JSONRPCApplication`. They can be added as a part of `a2a-sdk`'
                ' optional dependencies, `a2a-sdk[http-server]`.'
            )

```

### Checks

I tested this version with a sample agent and verified that it prints the correct message if optional dependencies are missing.

Also, I checked if a project adds only `a2a-sdk` (without the `[http-server]` group), `starlette`, and `sse-starlette`, excluding `fastapi` (requirements that correspond to all or most Python sample agents in `a2a-samples`), it can initialize the app without issues.

---

P.S. The initial version of this PR proposed to introduce framework-specific dependency groups `a2a-sdk[fastapi]`, etc. But in the later update, it was changed to `a2a-sdk[http-server]`, which combines both Starlette and FastAPI. Therefore now, I think, it makes sense to give a clearer ImportError message.

darkhaniop:
> Thanks for the updates, can you add new tests for the updated sections?

I added some tests. BTW, I did not want to make them "too hacky" to force certain imports to truly fail, so I tested "import failures" by flipping `_package_fastapi_installed` and `_package_starlette_installed` during the tests.

darkhaniop:
The last force push of this branch from fc1ac9a27d28b4cad2947614fcdc13d9232b84e8 to e8233d15467ba1e2cc2e598d146dbb03e4a17a94 above is to avoid re-adding a notebook file `tests/eval/eval_sample_a2a_expense_reimburse_agent_server.ipynb` to the git tree.

According to [this explanation](https://github.com/a2aproject/a2a-python/pull/300#issuecomment-3070010571), it was removed from git history to keep the SDK tree lighter.
> Hi, I removed this commit from the repo and moved the notebook to a2a-samples since notebooks in commit history can significantly increase the size of cloned repos.

holtskinner:
> The last force push of this branch from [fc1ac9a](https://github.com/a2aproject/a2a-python/commit/fc1ac9a27d28b4cad2947614fcdc13d9232b84e8) to [e8233d1](https://github.com/a2aproject/a2a-python/commit/e8233d15467ba1e2cc2e598d146dbb03e4a17a94) above is to avoid re-adding a notebook file `tests/eval/eval_sample_a2a_expense_reimburse_agent_server.ipynb` to the git tree.
> 
> According to [this explanation](https://github.com/a2aproject/a2a-python/pull/300#issuecomment-3070010571), it was removed from git history to keep the SDK tree lighter.
> 
> > Hi, I removed this commit from the repo and moved the notebook to a2a-samples since notebooks in commit history can significantly increase the size of cloned repos.

Thanks for this.

tests/server/apps/jsonrpc/test_fastapi_app.py
```diff
@@ -0,0 +1,80 @@
...
+    def mock_app_params(self) -> dict:
+        # Mock http_handler
+        mock_handler = MagicMock(spec=RequestHandler)
+        # Mock agent_card with essential attributes accessed in __init__
+        mock_agent_card = MagicMock(spec=AgentCard)
+        # Ensure 'url' attribute exists on the mock_agent_card, as it's accessed
+        # in __init__
+        mock_agent_card.url = 'http://example.com'
+        # Ensure 'supportsAuthenticatedExtendedCard' attribute exists
+        mock_agent_card.supportsAuthenticatedExtendedCard = False
```
aneeshgarg: thi variable naem needs to be changed to snake case version

holtskinner:
@darkhaniop Sorry for the extra hassle, but could you make this work with the new refactoring in https://github.com/a2aproject/a2a-python/pull/348

tests/server/apps/jsonrpc/test_fastapi_app.py
darkhaniop: @aneeshgarg, thanks for reviewing the changes.
I plan to update the PR to reflect a recent major refactoring in the main branch. While doing that, I will check the mocked declarations you mentioned and apply the necessary adjustments.

darkhaniop: Addressed in 2f74bc825062ae0c47f52745f15726f9d7c64308.

darkhaniop:
> @darkhaniop Sorry for the extra hassle, but could you make this work with the new refactoring in #348

From what I understood, the newly added RESTful interface modules directly import Starlette and FastAPI (and since they are always initialized by the upper-level package in `a2a.server.apps.__init__.py`), we needed to refactor these new modules similar to the ones in `./apps/jsonrpc` to ensure abstract apps continue working (e.g., the `helloworld` sample).

So, I refactored the imports and made other minor adjustments to make it work with recent changes from the main branch.

Please let me know if you find anything missing.

src/a2a/server/apps/rest/rest_adapter.py
```diff
@@ -2,12 +2,35 @@
 import logging
 
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from sse_starlette.sse import EventSourceResponse
```
kthota-g: Duplicate imports to be removed

darkhaniop: Addressed in 109488cac8260690a0e6826d509fdef07d4f52ae.

darkhaniop:
@kthota-g, thanks for the review. I applied the changes.
No new unit tests were added. I don't think we need them for these cases.

In `rest_handler.py`, the import from starlette is currently used only for type annotations, so no additional tests are necessary.
And, it looks like methods in `error_handlers.py` are only used as decorators in `RESTAdapter`, which handles the import error message.

PR #236, opened 2025-06-24
fix: Fixes Short Circuit clause for context ID

### LINKED ISSUE #235: [Bug]: Short-circuit evaluation bug in TaskManager
### What happened?

Unless I have misunderstood the intent, there maybe an unintended short-circuit evaluation in the conditional assignment of `self.context_id`

https://github.com/a2aproject/a2a-python/blob/4050b8123c9a96dc942b74a58a2a76fee111906a/src/a2a/server/tasks/task_manager.py#L110-L111

1. When `self.context_id` is falsy (None, empty string, etc.), then the`not` clause for `self.context_id` evaluates to `True`
    - Due to short-circuit evaluation care of `and`, Python never evaluates the second part: `self.context_id != event.contextId`
    - The condition becomes   ~~ `True` and <not evaluated> = `True`. ~~
    - The code sets `self.context_id = event.contextId` (which is correct behavior), ok so far.
 
2. When `self.context_id` has a value:
    - the not in `self.context_id` evaluates to `False`
    - Due to short-circuit evaluation, the entire condition becomes `False`
    - So context ID validation `self.context_id != event.contextId` never runs

So the current code will never detect mismatched context IDs because the validation check part is unreachable when `self.context_id` already has a value. Like I say though I might be wrong and forgive wasting time if so. 

I think something like this should do the job:

```
         if not self.task_id:
             self.task_id = task_id_from_event
-        if not self.context_id and self.context_id != event.contextId:
+        if self.context_id and self.context_id != event.contextId:
+            raise ServerError(
+                error=InvalidParamsError(
+                    message=f"Context in event doesn't match TaskManager: {self.context_id} : {event.contextId}"
+                )
+            )
+        if not self.context_id:
             self.context_id = event.contextId
```

### Relevant log output

```shell

```

### Code of Conduct

lukehinds:
If #235 is an issue, this should fix. If its as intended, apologies for the noise and feel free to close

Fixes #235 🦕

### DIFF
--- a/src/a2a/server/events/event_consumer.py
+++ b/src/a2a/server/events/event_consumer.py
@@ -138,6 +138,10 @@ class EventConsumer:
                 # python 3.12 and get a queue empty error on an open queue
                 if self.queue.is_closed():
                     break
+            except Exception as e:
+                logger.error(f'Stopping event consumption due to exception: {e}')
+                self._exception = e
+                continue
 
     def agent_task_callback(self, agent_task: asyncio.Task[None]) -> None:
         """Callback to handle exceptions from the agent's execution task.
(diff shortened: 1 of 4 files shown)


holtskinner:
@pstephengoogle You're listed as the author of this statement, can you verify this?

mindpower (approved):

PR #259, opened 2025-06-30
feat: Support for database backend Task Store

kthota-g:
```
Release-As: 0.2.11
```

Add support for database backed TaskStore

### Installation

```bash
uv add a2a-sdk[postgresql] #postgres driver
or 
uv add a2a-sdk[mysql] #mysql driver
or
uv add a2a-sdk[sqlite] #sqlite driver
or
uv add a2a-sdk[sql]. #install all three sql drivers
```

### Usage

```py
from a2a.server.tasks import DatabaseTaskStore
from sqlalchemy.ext.asyncio import (
        create_async_engine,
    )
...
...
# Postgres - "postgresql+asyncpg://postgres:postgres@localhost:5432/a2a_test"
# Sqlite - "sqlite+aiosqlite:///file::memory:?cache=shared"
# Mysql - "mysql+aiomysql://admin:saysoyeah@localhost:3306/testDB"
engine = create_async_engine(
    "postgresql+asyncpg://postgres:postgres@localhost:5432/a2a_test", echo=False
)
request_handler = DefaultRequestHandler(
    agent_executor=CurrencyAgentExecutor(),
    task_store=DatabaseTaskStore(engine=engine),
    push_notifier=InMemoryPushNotifier(httpx_client),
)
```


.github/workflows/unit-tests.yml
```diff
@@ -20,13 +31,19 @@ jobs:
...
+          export POSTGRES_TEST_DSN="postgresql+asyncpg://postgres:postgres@localhost:5432/a2a_test"
+
       - name: Install uv
         uses: astral-sh/setup-uv@v6
       - name: Add uv to PATH
         run: |
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - name: Install dependencies
-        run: uv sync --dev
+        run: uv sync --dev  --extra sql
```
holtskinner:
```suggestion
        run: uv sync --dev --extra sql
```

.github/actions/spelling/expect.txt
holtskinner, line 1: These should all be in allow.txt

.github/workflows/unit-tests.yml
```diff
@@ -20,13 +31,19 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Set postgres for tests
```
swapydapy, line 34: Do we need to start the postgres docker instance as well?

swapydapy: I see there a services section here which starts a db. Then do we need that docker-compose?

kthota-g: There is no dependency on the docker images  for unit tests. I have removed this one.

PR #264, opened 2025-07-01
fix: client should not specify `taskId` if it doesn't exist

njbrake:
# Description

I'm submitting this PR as a proposal based off of https://github.com/a2aproject/a2a-python/issues/218#issuecomment-3005858571.

If the taskId doesn't exist, I was thinking that an exception should be thrown to prevent the client from specifying the id. The reason I think this change is required because if the client specifies a taskId but one doesnt exist, it won't actually get push notifications based off that first request, because the logic to `set_info()` gets skipped because there is no `task` when _setup_message_execution is called.

Found while working on https://github.com/mozilla-ai/any-agent/issues/414


kthota-g:
@njbrake this method has been refactored recently to address no_existing_task scenario. So, the push notifications get sent for new task scenarios as well with a new task_id.

But regardless, throwing an exception makes sense for the scenario where task object could not be found for the specified task_id. Would you mind adding a test for this scenario?

njbrake:
Thanks @kthota-g! Added two tests.

PR #269, opened 2025-07-02
feat(server): implement Redis-based event queue manager

RobinQu:
- Add RedisQueueManager class to handle event queues using Redis
- Implement core functionalities: add, get, tap, close, create_or_tap
- Add unit tests for RedisQueueManager
- Update .gitignore to exclude spec.json and .idea
- Refactor event type definition in event_queue.py
- Add fakeredis dependency in pyproject.toml

# Description

Related to https://github.com/a2aproject/A2A/issues/674#issuecomment-3027987885


RobinQu:
@pstephengoogle  I will fix failed checks later. Let's dicuss implmentation detail first.

pstephengoogle:
This aligns with how I thought this could be achieved. Do you feel this addresses all your requirements?

RobinQu:
> This aligns with how I thought this could be achieved. Do you feel this addresses all your requirements?

This is enough for a MVP version.  Some TTL housekeep for Redis keys should be added later. 

Please give me some help about your unit tests. It seems that your present code fails those tests.

RobinQu:
> > This aligns with how I thought this could be achieved. Do you feel this addresses all your requirements?
> 
> This is enough for a MVP version. Some TTL housekeep for Redis keys should be added later.
> 
> Please give me some help about your unit tests. It seems that your present code fails those tests.

Never mind. I figured the problem. Hope you guys give some comments and finish this PR.

pyproject.toml
```diff
@@ -21,6 +21,8 @@ dependencies = [
   "grpcio-tools>=1.60",
   "grpcio_reflection>=1.7.0",
   "protobuf==5.29.5",
+  "fakeredis>=2.30.1",
+  "redis>=6.2.0",
```
matino: IMO `redis` should be optional dependency. Also, you don't need to specify `fakeredis` here, since you already added it to `dev` group.

src/a2a/server/events/redis_queue_manager.py
```diff
@@ -0,0 +1,321 @@
...
+        node_id: A unique id to be associated with task id in global registry. If node id is not matched, events won't be populated to queues in other `RedisQueueManager`s.
+    """
+
+    def __init__(
+        self,
+        redis_client: Redis,
+        relay_channel_key_prefix: str = 'a2a.event.relay.',
+        task_registry_key: str = 'a2a.event.registry',
+        task_id_ttl_in_second: int = 60 * 60 * 24,
+        node_id: str = str(uuid.uuid4()),
```
matino: A much better practise is to use `node_id: str | None = None` and `self_node_id = node_id or str(uuid.uuid4())`, otherwise you end up with the same node ids, when creating multiple `RedisQueueManager` instances without providing `node_id`.

lukehinds:
Should this not be an abstract interface for any queue / event manager to implement with,  and we decouple this by defining a generic queue/event‐manager interface rather than embedding a Redis‐specific implementation directly into the library. 

Embedding a single vendor’s system (one that recently caused some drama around changing their license) risks locking users into that choice and makes it harder to support alternative backends down the line. Instead, let’s introduce an abstract interface (e.g. AbstractQueue or AbstractEventManager) so that anyone can plug in Redis, Valkey, RabbitMQ, AWS SQS, or any other system they prefer. This will keep our core library lean, flexible, and truly pluggable.

pyproject.toml
RobinQu: Done. 

BTW, should I update uv.lock file as well? After `uv sync`, uv.lock is changed a lot. Or you will do the file update in your release pipeline?

src/a2a/server/events/redis_queue_manager.py
RobinQu: Good catch. Fixed via e40519a1914baf3a3718f1ac3865408c07c9f282

RobinQu:
> Should this not be an abstract interface for any queue / event manager to implement with, and we decouple this by defining a generic queue/event‐manager interface rather than embedding a Redis‐specific implementation directly into the library.
> 
> Embedding a single vendor’s system (one that recently caused some drama around changing their license) risks locking users into that choice and makes it harder to support alternative backends down the line. Instead, let’s introduce an abstract interface (e.g. AbstractQueue or AbstractEventManager) so that anyone can plug in Redis, Valkey, RabbitMQ, AWS SQS, or any other system they prefer. This will keep our core library lean, flexible, and truly pluggable.

I raised the same opnion in  https://github.com/a2aproject/A2A/issues/674#issuecomment-3024203008

I think some major refactoring is needed in order to truely make the queue mechanism pluggable.  AFAIK, this libraray has already contains some vendor specific code like `DatabaseTaskStore` which supports mysql, pgsql and sqlite as `TaskStore` storage.

As a library user, I would like this library contains out-of-box implementations of necessary components, like `QueueManager` and `TaskStore`, as as well as proper abstractions to allow me to have my private toys. And these won't conflict with each other.

In terms of `QueueManager`, I am surprised to see there is no single implementation that supports distributed senario, and I think `RedisQeuueManager` would be a mininum component to demonstrate how you would handle multi processes or multi nodes in a horizontally scalable deplyoment. And to be honest, this implemntation is sufficient for small to medium bussiness, covering needs of most of library users.

lukehinds:
The more I think about this, the less inclined I am to have us set this as a precedence, if we have already done so with Databases, I don't think that's a good enough reason to double down and add more opinionated implementations. 

As this is a Linux Foundation project, we should attempt to provide interfaces available to all projects to use, otherwise we may (and very likely will) end up carrying  maintenance in tree of a many vendor solutions (some of which will fall out of maintenance). Trust me I have been there with Kubernetes, openstack and other such projects. 

For that reason I really don't think this is a good idea if we can avoid it, instead we should provide an agnostic / abstract base interfaces and then we can have /examples showing how to use with any given external project, just as we have for frameworks (langchain, crewAI etc).

ognis1205 (commented):
I agree with @lukehinds — and I think the same kind of discussion should apply to `DatabaseTaskStore` as well.

Given the protocol-oriented nature of this project, especially under the Linux Foundation, we should aim to keep server-side components as abstract and vendor-neutral as possible. In fact, it might be more appropriate to treat such implementations as reference implementations rather than embedding them in the core.

RobinQu:
@lukehinds How will the discussion be initiated and who should be invited? Is there anyone who is actually pushing forward this issue?

pstephengoogle:
Sorry was OOO for the long weekend. To clarify a few things:
- The DatabaseTaskStore is not tied to a specific vendor. It uses sqlalchemy library which allows for many DB implementations to be directly used.
- I agree we should not have a vendor specific implementation in the SDK. The pattern we are following is to define the right interfaces/logic in the SDK and show a proof of concept with a specific vendor solution in the samples repo. This way you can verify your abstractions are valid for at least one vendor, and hopefully others can implement different versions for different vendors until the SDK abstractions are correct.

To that end, I think it makes sense to commit an `ProducerConsumerQueueManager` or similar which provides the right logic and abstractions around using an distributed producer/consumer paradigm with the `QueueManager` interface. Then you can implement the redis based producer/consumer instance (in the samples repo) to prove the end-to-end behavior

PR #276, opened 2025-07-04
fix: add proper a2a request body documentation to Swagger UI

### LINKED ISSUE #274: [Bug]: No proper documentation for A2A request in Swagger UI for `A2AFastAPIApplication`
### What happened?

The way `A2AFastAPIApplication` sets up its A2A request route causes FastAPI’s Swagger UI to be less usable. Because the route is defined to receive a generic `Request` object, there's no model defined for the request body. As a result, FastAPI doesn’t generate any documentation for the request body, and there's no way to provide or test a request body in the Swagger UI at all.

This makes it impossible to use Swagger UI to test the endpoint directly, since you can’t input the necessary data.

**Current behavior:**

Image

**Expected behavior after fix:**

Image

### Relevant log output

```shell

```

### Code of Conduct

inesmcm26:
When using `A2AFastAPIApplication`, the A2A request endpoint showed no space to fill in a request body in Swagger UI.
This made it impossible to use Swagger UI to test the endpoint directly, since you couldn't input the necessary data.

**Changes**
- Added proper request body documentation using `A2ARequest` model
- Now shows a text box to enter JSON requests and an example of the expected request format

**Before**
image

**After**
image

Fixes #274

### DIFF
--- a/src/a2a/server/apps/jsonrpc/fastapi_app.py
+++ b/src/a2a/server/apps/jsonrpc/fastapi_app.py
@@ -1,5 +1,7 @@
 import logging
 
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from typing import Any
 
 from fastapi import FastAPI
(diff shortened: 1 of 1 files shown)


holtskinner:
Can you resolve the merge conflicts after #280 was added? It should improve some of what you were looking for.

PR #292, opened 2025-07-09
feat: add A2ABaseModel

### LINKED ISSUE #291: [Feat]: Support `populate_by_name = True` for Pydantic models

Currently, instantiating models like `APIKeySecurityScheme` using keyword arguments for fields with aliases (e.g., `in_`) does not work as expected:

```python
APIKeySecurityScheme(
    name='X-API-Key',
    in_=In.header,  # Raises validation error unless alias is respected
)
```

This happens because Pydantic requires `populate_by_name = True` in the model config to allow population via the field name instead of its alias (in in this case, which is a reserved keyword in Python).

Introduce a common base model (`A2ABaseModel`) that all models inherit from, and configure it with `populate_by_name = True` via `ConfigDict`.

```python
# src/a2a/_base.py
from pydantic import BaseModel, ConfigDict

class A2ABaseModel(BaseModel):
    model_config = ConfigDict(
        populate_by_name=True,  #Enables safer field population
    )
```

Then we pass the option `--base-class a2a._base.MyBaseModel` to the `datamodel-codegen` command.

 - Using **dict-style unpacking to avoid naming issues (e.g., **{'in': 'header'}), but this is less readable and more error-prone.

### Additional context

https://github.com/a2aproject/a2a-python/pull/290#discussion_r2195842201

### Code of Conduct

ognis1205:
# Description

Fixes #291 🦕


src/a2a/_base.py
```diff
@@ -0,0 +1,13 @@
...
+
+class A2ABaseModel(BaseModel):
+    """Base class for shared behavior across A2A data models.
+
+    Provides a common configuration (e.g., alias-based population) and
+    serves as the foundation for future extensions or shared utilities.
+    """
+
+    model_config = ConfigDict(
+        populate_by_name=True,
```
mikeas1: There's a warning in the Pydantic docs discouraging use of `populate_by_name` as of v2.11 and recommending use of `validate_by_name` instead. It appears to be functionally equivalent, so prefer using that.

https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.populate_by_name

src/a2a/_base.py
```diff
@@ -0,0 +1,6 @@
+from pydantic import BaseModel, ConfigDict
+
+class A2ABaseModel(BaseModel):
+    model_config = ConfigDict(
+        populate_by_name=True,
+    )
```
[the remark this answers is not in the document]
mikeas1: We may do this in the future, but not now.

src/a2a/_base.py
ognis1205: Good catch — appreciate the info!

Switched over to `validate_by_name` and `validate_by_alias` per the latest Pydantic guidance.

mikeas1 (approved):

PR #324, opened 2025-07-18
fix: openapi working in sub-app

racinmat:
The `A2AFastAPIApplication` is very useful, but currently can not be used as a [Sub Application](https://fastapi.tiangolo.com/advanced/sub-applications/) without breaking the openapi specification.
[Lifespan does not work for Sub Applications](https://fastapi.tiangolo.com/advanced/events/#sub-applications), which makes this otherwise useful integration work in existing larger FastAPI apps.

This fix gets rid of the lifespan and instead enriches the openapi on the first call of the [openapi method](https://github.com/fastapi/fastapi/blob/0.116.1/fastapi/applications.py#L966).

I tested it locally and everything works with sub-application.

The following works after the fix with openapi components populated:
```python
from fastapi import FastAPI
from a2a.server.apps import A2AFastAPIApplication
app = FastAPI()
agent_app = A2AFastAPIApplication(...).build()
app.mount("/a2a", agent_app)
```

I am not adding `@override` although this overrides the `openapi`, because the pyright fails on it, see [the failed lint run](https://github.com/a2aproject/a2a-python/actions/runs/16370448989/job/46257297467?pr=324).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
   "pydantic>=2.11.3",
   "sse-starlette",
   "starlette",
-  "protobuf==5.29.5",
+  "protobuf>=5.29.5",
   "google-api-core>=1.26.0",
 ]
 
--- a/src/a2a/client/transports/grpc.py
+++ b/src/a2a/client/transports/grpc.py
@@ -87,7 +87,7 @@ class GrpcTransport(ClientTransport):
                 metadata=proto_utils.ToProto.metadata(request.metadata),
             )
         )
-        if response.task:
+        if response.HasField('task'):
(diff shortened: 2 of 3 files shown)


holtskinner:
@racinmat Thanks for the contribution, can you add tests for your change to show that it works with sub-applications?

racinmat:
@holtskinner added

holtskinner (approved):

holtskinner:
@kthota-g @pstephengoogle Can you please review?

racinmat:
@kthota-g @pstephengoogle please, this would really simplify my life, and I think it would help others too.
Is there anything blocking it?

racinmat:
@holtskinner is there anyone else who could review it instead?

PR #333, opened 2025-07-21
feat: Add server-side support for plumbing requested and activated extensions

mikeas1:
# Description

This adds PR adds support for reading requested extensions from a client, via the `X-A2A-Extensions` header, and plumbing that through the ServerCallContext.

Here's my rough thinking on the design choices:

1. Requests to activate extensions are specified in a transport-specific manner. For HTTP, it's using headers. For gRPC, it's using side-channel metadata (which are also HTTP headers).
2. However, indicating that an extension was requested is a transport-independent concept. So, it should be passed to the transport-independent layers of the SDK.
3. We already have a means of extracting relevant information from the transport layer (and others) and passing down to the independent layers: the `ServerCallContext`.
4. So, we can put the list of requested extensions here. The transport layers are responsible for pulling out the list of requested extensions in whatever transport specific means they are communicated, and passing these down via the ServerCallContext.
5. Returning the list of extensions that were activated is slightly more annoying. I didn't want to have to refactor every layer to plumb both the request-specific response type (such as SendMessageResponse) AND the list of extensions activated. That's just too intrusive.
6. So, I need some way to communicate the list of activated extension from the independent layers back "up" to the transport layers via something other than the return value. I'm not thrilled with the decision here, but the answer is to use interior mutability: I put a mutable set in the ServerCallContext, and allow lower levels to modify it to add the activated extensions.
7. The transport layers then need to read the list of activated extensions from the ServerCallContext and pass those back in the response in the transport-defined method.

The result is that we have full stack plumbing of requested and activated extensions by way of the ServerCallContext.


src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -99,7 +100,15 @@ def build(self, request: Request) -> ServerCallContext:
...
         state['headers'] = dict(request.headers)
-        return ServerCallContext(user=user, state=state)
+        return ServerCallContext(
+            user=user,
+            state=state,
+            requested_extensions={
+                ext
+                for h in request.headers.getlist(HTTP_EXTENSION_HEADER)
+                for ext in h.split(', ')
+            },
```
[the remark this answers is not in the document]
mikeas1: Done.

tests/server/apps/jsonrpc/test_jsonrpc_app.py
```diff
@@ -86,5 +100,149 @@ def some_other_method(self):
...
 
+class TestJSONRPCExtensions:
+    @pytest.fixture
+    def mock_handler(self):
+        handler = AsyncMock(spec=RequestHandler)
+        handler.on_message_send.return_value = SendMessageResponse(
+            root=SendMessageSuccessResponse(
+                id='1',
+                result=Message(
+                    messageId='test',
```
holtskinner: camelCase for these attributes is now deprecated and will be removed in 0.3.0

Please switch these parameters to use snake_case

src/a2a/server/agent_execution/context.py
```diff
@@ -143,6 +143,24 @@ def metadata(self) -> dict[str, Any]:
...
+        if self._call_context:
+            self._call_context.activated_extensions.add(uri)
+
+    @property
+    def requested_extensions(self) -> set[str]:
+        """Extensions that the client requested to activate."""
+        return (
+            self._call_context.requested_extensions
+            if self._call_context
+            else set()
```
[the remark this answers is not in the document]
mikeas1: Line length issue.

src/a2a/extensions/common.py
```diff
@@ -0,0 +1,27 @@
...
+        for ext in v.split(',')
+        if (stripped := ext.strip())
+    }
+
+
+def find_extension_by_uri(card: AgentCard, uri: str) -> AgentExtension | None:
+    """Find an AgentExtension in an AgentCard given a uri."""
+    for ext in card.capabilities.extensions or []:
+        if ext.uri == uri:
+            return ext
```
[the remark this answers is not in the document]
mikeas1: Line length limits makes this less concise.

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -42,6 +48,19 @@ def build(self, context: grpc.aio.ServicerContext) -> ServerCallContext:
...
+def _get_metadata_value(
+    context: grpc.aio.ServicerContext, key: str
+) -> list[str]:
+    md = context.invocation_metadata
+    raw_values: list[str | bytes] = []
+    if isinstance(md, Metadata):
+        raw_values = md.get_all(key)
+    elif isinstance(md, Sequence):
+        lower_key = key.lower()
+        raw_values = [e for (k, e) in md if k == lower_key]
```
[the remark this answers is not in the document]
mikeas1: Fixed.

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -16,10 +16,16 @@
         "'pip install a2a-sdk[grpc]'"
     ) from e
 
+from grpc.aio import Metadata
```
holtskinner: Move this import inside the try-catch since grpc is now optional

tests/extensions/test_common.py
```diff
@@ -0,0 +1,58 @@
...
+    ext1 = AgentExtension(uri='foo', description='The Foo extension')
+    ext2 = AgentExtension(uri='bar', description='The Bar extension')
+    card = AgentCard(
+        name='Test Agent',
+        description='Test Agent Description',
+        version='1.0',
+        url='http://test.com',
+        skills=[],
+        defaultInputModes=['text/plain'],
+        defaultOutputModes=['text/plain'],
```
holtskinner: convert all of the camelCase fields to snake_case

tests/server/apps/jsonrpc/test_jsonrpc_app.py
mikeas1: Done.

src/a2a/server/request_handlers/grpc_handler.py
mikeas1: Done.

tests/extensions/test_common.py
mikeas1: Done.

src/a2a/extensions/common.py
```diff
@@ -0,0 +1,27 @@
...
+    return {
+        stripped
+        for v in values
+        for ext in v.split(',')
+        if (stripped := ext.strip())
+    }
+
+
+def find_extension_by_uri(card: AgentCard, uri: str) -> AgentExtension | None:
+    """Find an AgentExtension in an AgentCard given a uri."""
```
kthota-g, line 22: Does the extension uri have any other significance other than acting as an identifier? e.g. if I have an extension that exposes custom functionality at an endpoint, could this uri be used?

mikeas1: The URI field here is for identifying the extension. What you're suggesting is possible, but it would be part of the AgentExtension.params fields. For example:

```json
{
  "uri": "https://github.com/mikeas1/a2a-catalog-ext/v1",
  "params": {
    "catalog_endpoint": "https://example.com/agents/catalog"
  }
}
```

You'd need two URIs: one for the consumer to understand what the extension is (so the consumer understands that this is an extension that describes a custom protocol), one for the endpoint that serves the extension-defined protocol.

PR #342, opened 2025-07-24
refactor: Refactor client code to support multi-transport interactions and simplify client interfaces

pstephengoogle:
# Description

Refactor the client codebase to support automated, configurable, client factory to generate a client based on the agent card details. Adds support for generation of clients based on JSONRPC, gRPC and RESTful interfaces.


src/a2a/utils/transports.py
holtskinner, line 1: This should probably be handled at the protocol level itself. See https://github.com/a2aproject/A2A/pull/896

holtskinner (changes requested):

src/a2a/utils/transports.py
pstephengoogle: We can clean that up when the two prs are merged. For now we can start here

PR #356, opened 2025-07-29
chore: Add support for authenticated extended card method

kthota-g:
# Description

Types are manually generated for implementing this PR, but this feature has types dependency on https://github.com/a2aproject/A2A/pull/929 and can only be merged once the type changes are merged into this repository.

This PR adds support for agent/authenticatedExtendedCard JSON-RPC method for serving authenticated extended card.

The previous HTTP GET /agent/authenticatedExtendedCard endpoint has been deprecated and will be removed in a future release.

Release-As: 0.3.0


src/a2a/server/apps/jsonrpc/fastapi_app.py
```diff
@@ -89,6 +89,7 @@ def add_routes_to_app(
         )(self._handle_requests)
         app.get(agent_card_url)(self._handle_get_agent_card)
 
+        # TODO: deprecated endpoint to be removed in a future release
```
dmandar, line 99: nit: log an error if the endpoint gets called?

dmandar, line 99: nit: log a warning if the endpoint gets called?

holtskinner: `DeprecationWarning` probably makes more sense?

kthota-g: it's added part of the method here - https://github.com/a2aproject/a2a-python/blob/d20477e33f7d88d430be545370b0845119d830ea/src/a2a/server/apps/jsonrpc/jsonrpc_app.py#L463

holtskinner (approved):

PR #363, opened 2025-07-29
refactor: Refactor client into BaseClient + ClientTransport

mikeas1:
A refactor of the refactor.

This pulls out the common code from across transport-specific clients into one BaseClient, then shoves all the transport-specific code into implementations of `ClientTransport`.

### DIFF
--- a/src/a2a/client/legacy.py
+++ b/src/a2a/client/legacy.py
@@ -15,6 +15,7 @@ from a2a.types import (
     CancelTaskRequest,
     CancelTaskResponse,
     CancelTaskSuccessResponse,
+    GetTaskPushNotificationConfigParams,
     GetTaskPushNotificationConfigRequest,
     GetTaskPushNotificationConfigResponse,
     GetTaskPushNotificationConfigSuccessResponse,
@@ -31,6 +32,7 @@ from a2a.types import (
     SetTaskPushNotificationConfigRequest,
     SetTaskPushNotificationConfigResponse,
     SetTaskPushNotificationConfigSuccessResponse,
+    TaskIdParams,
     TaskResubscriptionRequest,
 )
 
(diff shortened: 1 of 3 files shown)


src/a2a/client/transports/base.py
```diff
@@ -0,0 +1,103 @@
...
+        self,
+        request: MessageSendParams,
+        *,
+        context: ClientCallContext | None = None,
+    ) -> AsyncGenerator[
+        Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent
+    ]:
+        """Sends a streaming message request to the agent and yields responses as they arrive."""
+        return
+        yield
```
holtskinner, line 42: Not sure these are needed?

src/a2a/client/transports/base.py
```diff
@@ -0,0 +1,103 @@
...
+        self,
+        request: TaskIdParams,
+        *,
+        context: ClientCallContext | None = None,
+    ) -> AsyncGenerator[
+        Task | Message | TaskStatusUpdateEvent | TaskArtifactUpdateEvent
+    ]:
+        """Reconnects to get task updates."""
+        return
+        yield
```
holtskinner, line 91: Not sure these are needed?

src/a2a/client/transports/base.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @abstractmethod
+    async def get_card(
+        self,
+        *,
+        context: ClientCallContext | None = None,
+    ) -> AgentCard:
+        """Retrieves the agent's card."""
```
holtskinner:
```suggestion
        """Retrieves the Agent Card."""
```

src/a2a/client/transports/base.py
mikeas1: It's ugly, but without them you get type checking errors if you implement this ABC with a generator. It claims you are changing the signature in the overridden method. I did a quick check with gemini and it seemed to agree this was a tooling deficiency and this workaround was correct, for whatever that's worth.

src/a2a/client/legacy.py
```diff
@@ -0,0 +1,214 @@
...
+            httpx_client, agent_card, url, interceptors
+        )
+
+    async def send_message(
+        self,
+        request: SendMessageRequest,
+        *,
+        http_kwargs: dict[str, Any] | None = None,
+        context: ClientCallContext | None = None,
+    ) -> SendMessageResponse:
```
holtskinner, line 66: Missing docstrings in this module, can you copy them from the version on main?

mikeas1: Done.

PR #379, opened 2025-08-04
fix: Client hangs when implementing `AgentExecutor` and `await`ing twice in execute method

### LINKED ISSUE #367: [Bug]: Client hangs when implementing AgentExecutor and awaiting twice in execute method
### What happened?

A bug happened!
When implementing the AgentExecutor and calling await twice on the complete method within the execute method, the client consumption gets stuck. The queue does not return the second complete result to the client.
Specifically, the issue occurs in the demo_agent_executor.py file where the execute() method is called twice using await:

```    
async def execute(
        self,
        context: RequestContext,
        event_queue: EventQueue,
    ) -> None:
        await updater.add_artifact(
            [Part(root=TextPart("hello! "))],
            name='conversion_result',
        )

        await updater.complete()
        await updater.complete()
```

### Relevant log output

```shell

```

### Code of Conduct

meoow113:
Fixes #367

 - Add clear_events method for immediate queue cleanup
 - Integrate with event_consumer to prevent hanging
 - Support child queue clearing for complete cleanup


src/a2a/server/events/event_queue.py
```diff
@@ -154,3 +154,39 @@
...
+                logger.debug(f'Cleared {cleared_count} unprocessed events from EventQueue.')
+
+            # Clear all child queues
+            if clear_child_queues:
+                child_tasks = []
+                for child in self._children:
+                    child_tasks.append(asyncio.create_task(child.clear_events()))
+
+                if child_tasks:
+                    await asyncio.wait(child_tasks, return_when=asyncio.ALL_COMPLETED)
```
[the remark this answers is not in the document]
holtskinner: @meoow113 Can you please address if relevant?

src/a2a/server/events/event_consumer.py
```diff
@@ -125,6 +125,7 @@ async def consume_all(self) -> AsyncGenerator[Event]:
                 # other part is waiting for an event or a closed queue.
                 if is_final_event:
                     logger.debug('Stopping event consumption in consume_all.')
+                    await self.queue.clear_events()
```
holtskinner: Is there a reason for why this is a separate method instead of incorporating into `close()`?

holtskinner (changes requested):
Can you please add more in your PR description for what this does?

src/a2a/server/events/event_consumer.py
meoow113: For this bug, I have two approaches:
Add an immediate parameter to the close method: Whether to close immediately. If true, then regardless of whether there are messages in the queue, all messages will be discarded and the queue will be closed immediately;
Target only the current bug scenario: In this scenario, I clear the messages before calling the close method, so I can successfully close the message queue.

meoow113:
> Can you please add more in your PR description for what this does?

ok, i have updated my PR description

meoow113:
I have explained this in other replay

meoow113:
> Can you please add more in your PR description for what this does?

I have explained this in other replay

pstephengoogle:
Overall, this looks good. Can we add unit tests to exercise the new patterns/behaviors?

pstephengoogle (approved):

PR #386, opened 2025-08-06
Add agent card as a route in rest adapter

rajeshvelicheti:
# Description

### DIFF
--- a/src/a2a/server/apps/rest/rest_adapter.py
+++ b/src/a2a/server/apps/rest/rest_adapter.py
@@ -93,6 +93,9 @@ class RESTAdapter:
         Returns:
             A JSONResponse containing the agent card data.
         """
+        if self.agent_card.supports_authenticated_extended_card:
+            return await self.handle_authenticated_agent_card(request)
+
         # The public agent card is a direct serialization of the agent_card
         # provided at initialization.
         return JSONResponse(
@@ -174,8 +177,7 @@ class RESTAdapter:
             ('/v1/tasks', 'GET'): functools.partial(
                 self._handle_request, self.handler.list_tasks
             ),
+            ('/v1/card', 'GET') : self.handle_get_agent_card,
         }
-        if self.agent_card.supports_authenticated_extended_card:
-            routes[('/v1/card', 'GET')] = self.handle_authenticated_agent_card
 
         return routes


src/a2a/server/apps/rest/rest_adapter.py
```diff
@@ -174,8 +177,7 @@ def routes(self) -> dict[tuple[str, str], Callable[[Request], Any]]:
             ('/v1/tasks', 'GET'): functools.partial(
                 self._handle_request, self.handler.list_tasks
             ),
+            ('/v1/card', 'GET'): self.handle_get_agent_card,
```
kthota-g: The /v1/card is the endpoint used specifically for authenticated card for rest transport - https://a2a-protocol.org/latest/specification/#356-method-mapping-reference-table

The standard agent card should be exposed at well-known url.

kthota-g (changes requested):

src/a2a/server/apps/rest/rest_adapter.py
rajeshvelicheti: Thanks @kthota-g This was required for Agent Engine. @pstephengoogle this is based on our conversation yesterday, I can change if required.

kthota-g: I imagine that the issue is agent card is not available for ReST based A2A Server and the default url where it should be available is well-known url, similar to the JSON-RPC transports.

The /v1/card should be available for serving the extended card, if extended card is flagged true.

If we simply serve the agent card at /v1/card, what would be the url for extended cards?

pstephengoogle: So yes, the standard REST server should support the standard well known path serving, since it is already providing an http1 surface. But the the card (either public if no provided extended card or mutation callback is provided) should also serve from the /v1/card path. For agent space, they should be able to use trawler to get the well known card, and/or the /v1/card path for the extended card

pstephengoogle: Please look at the callback modifiers that mandar added to the jsonrpc application to add here. We should have the following behavior:

User provides a static card
Optionally, user provides a runtime mutation callback to the static card - the output of the Mutation(StaticCard) should be what serves from wellknown path
Optionally, user provides a runtime, credentialed callback - the output of the CredentialedMutation(StaticCard, RequestContext) should be what serves from the /v1/card path.

If no mutation is provided - the static card is served from both places
If runtime mutation is provided, but no credentialed mutation, the Mutation(StaticCard) should serve both well known and /v1/card

holtskinner: @rajeshvelicheti Does this need to be addressed further?

rajeshvelicheti: @kthota-g could you please review again. Attached is the well-known endpoint for public url. The API routes now include /v1/card for the authenticated url and well_known/agent_json for the public url. We will document this in the colab and other places, so users could query the public agent before getting to authenticated agent.

Screenshot 2025-08-11 at 4 48 10 PM

@pstephengoogle I had to modify the signatures in rest_adapter to follow the same pattern as in the other routes, making it easy for the Agent Engine updates in the future. Please let me know if you see any issue with this.

i am fixing the unit tests now.

PR #410, opened 2025-08-15
Update _base.py

coyoteXujie:
Pydantic v2 中适配字段名和字段别名的方式，原始方式
    model_config = ConfigDict(
        # SEE: https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.populate_by_name
        validate_by_name=True,
        validate_by_alias=True,
        serialize_by_alias=True,
        alias_generator=to_camel_custom,
    )
会导致  实例化 AgentCard 方法的时候 报错
pydantic_core._pydantic_core.ValidationError: 2 validation errors for AgentCard defaultInputModes
  Field required [type=missing, input_value={'name': 'Hello World Age...ed_extended_card': True}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
defaultOutputModes

# Description

### DIFF
--- a/src/a2a/_base.py
+++ b/src/a2a/_base.py
@@ -31,8 +31,6 @@ class A2ABaseModel(BaseModel):
 
     model_config = ConfigDict(
         # SEE: https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.populate_by_name
-        validate_by_name=True,
-        validate_by_alias=True,
-        serialize_by_alias=True,
+        populate_by_name=True,
         alias_generator=to_camel_custom,
     )


mikeas1:
Unfortunately, our current use of Pydantic is only compatible with v2.11. If you use v2.10, you'll run into errors that you discovered.

There are 2 problems:
1. We use validate_by_alias and validate_by_name. These were only introduced in v2.11. We could switch to using `populate_by_name=True`, which is compatible.
2. We use `serialize_by_alias`. This is the key problem that prevents v2.10 backwards compatibility. The protocol defines camelCase property names, so we MUST serialize our Pydantic models using camelCase property names. However, our property names are in snake_case to be idiomatic. We get around this by defining camelCase aliases and always serializing the models using the alias.

To downgrade to Pydantic version v2.10, we'd need to update every call to `model_dump`/`mode_dump_json` to specify `by_alias=True`.

coyoteXujie:
Thank you for your reply. I will upgrade my version of Pydantic. The model_dump method is indeed a difficult problem to handle. In some places, the object of BaseModel is returned




草原狼
***@***.***

PR #418, opened 2025-08-20
fix: Re-constructing Agent RPC URL in the AgentCard using the card's request attributes

### LINKED ISSUE #398: [Bug]: Agent URL in the AgentCard returned by A2AStarletteApplication may be inaccessible
### What happened?

When running behind load balancers or any kind of gateways, the original request may come with a different hostname, schema, port, and even path.

For example, an agent in Cloud Run service may be exposed via a load balancer with a hostname, port and schema different from the Google Cloud Run service URL. 

**Proposed behavior**

In `A2AStarletteApplication`, `_handle_get_agent_card` should create a copy of the saved AgentCard with agent's URL constructed with regards to the AgentCard HTTP request URL (host, schema, port) and, **if present**, `X-Forwarded-Proto`, `X-Forwarded-Host` and  `X-Forwarded-Path` headers.

Even though `X-Forwarded-Path` is a non-standard header, load balancers with URL maps use it to pass original URL path used by the client. 

When `X-Forwarded-Path` is present, part of the path before the AgentCard path, should be added to the beginning of the agent RPC url path. 
For example, if the AgentCard path is set to `/.well-known/agent-card.json` (default value), but `X-Forwarded-Path` header is `/agents/myagent/.well-known/agent-card.json` then the result path for the agent's RPC URL should be `/agents/myagent`. 

So, It maybe that an agent deployed to `https://my-agent-1234567.us-central1.run.app` service is exposed via a load balancer as `https://api.company.com/agents/my-agent` via a URL map. The agent card would be exposed `https://api.company.com/agents/my-agent/.well-known/agent-card.json`.

Here is what I use as a workaround: 

```python
class A2AStarletteApplicationWithHost(A2AStarletteApplication):
    """This class makes sure the agent's url in the AgentCard has the same
    host, port and schema as in the request.
    """

    @override
    def build(
            self,
            agent_card_url: str = AGENT_CARD_WELL_KNOWN_PATH,
            rpc_url: str = DEFAULT_RPC_URL,
            **kwargs: Any,
        ) -> Starlette:
        self.rpc_url = rpc_url
        self.card_url = agent_card_url
        return super().build(
            agent_card_url=agent_card_url,
            rpc_url=rpc_url,
            **kwargs
        )


    @override
    async def _handle_get_agent_card(self, request: Request) -> JSONResponse:
        """Handles requests for the agent card endpoint
        by dynamically building the agent's RPC url (`url` property in AgentCard).

        Args:
            request: The incoming Starlette Request object.

        Returns:
            A JSONResponse containing the agent card data.
        """
        port = None
        if "X-Forwarded-Host" in request.headers:
            host = request.headers["X-Forwarded-Host"]
        else:
            host = request.url.hostname
            port = request.url.port
        if "X-Forwarded-Proto" in request.headers:
            scheme = request.headers["X-Forwarded-Proto"]
        else:
            scheme = request.url.scheme
        if not scheme:
            scheme = "http"
        if ":" in host: # type: ignore
            comps = host.rsplit(":", 1) # type: ignore
            host = comps[0]
            port = comps[1]

        path = ""
        if "X-Forwarded-Path" in request.headers:
            # Handle URL maps, e.g. https://myagents.com/agents/myagent to https://myagent-12345678.us-central1.run.app
            path = request.headers["X-Forwarded-Path"].strip()
            if (path
                and path.lower().endswith(self.card_url.lower())
                and len(path) - len(self.card_url) > 1
            ):
                path = path[:-len(self.card_url)].rstrip("/") + self.rpc_url
        else:
            path = self.rpc_url

        card = self.agent_card.model_copy()
        source_parsed = URL(card.url)
        if port:
            card.url = str(
                source_parsed.replace(
                    hostname=host,
                    port=port,
                    scheme=scheme,
                    path=path
                )
            )
        else:
            card.url = str(
                source_parsed.replace(
                    hostname=host,
                    scheme=scheme,
                    path=path
                )
            )

        return JSONResponse(
            card.model_dump(mode='json', exclude_none=True)
        )
```

### Relevant log output

```shell

```

### Code of Conduct

vladkol:
# Re-constructing Agent RPC URL in the AgentCard using the card's request attributes

Current AgentCard request handler implementation in `JSONRPCApplication` and `RESTAdapter` return AgentCard with the agent's RPC url is it's originally created in code, often as local endpoints (e.g. `http://localhost:8000`).

Modern cloud runtime usually allows multiple ways to access deployed services. Some of them, such as load balancers may expose these services under vanity URLs. Others, like [Google Cloud Run](https://cloud.google.com/), have endpoint URLs dynamically assigned at a time of the deployment.

This PR adds "rewriting" of the agent's RPC url based on how the AgentCard was requested: with host, port, protocol and path adjusted according to the way the card's request was made. It accounts for deployments with load balancers, URL mapping, etc. 

Fixes #398 
Will also properly fix https://github.com/google/adk-python/issues/2405

Release-As: 0.3.3


kthota-g (changes requested):
It's not required for `agent` and `agent_card` to be hosted from the same url. For example, agent card may be hosted on `http://example.com` and the `agent_card.url` could be specified as `http://agent.example.com`. So, we can't default to using the forwarded headers for the url.

[Card modifier](https://github.com/a2aproject/a2a-python/blob/a3e7e1ef2684f979a3b8cbde1f9fd24ce9154e40/src/a2a/server/apps/jsonrpc/starlette_app.py#L57) should be used for this scenario so users can mutate the card as needed.

PR #420, opened 2025-08-21
feat: metadata serialization for custom objects and large integers

youngchannelforyou:
- Fix JSON serialization error for custom objects in metadata by adding _make_json_serializable helper method that converts non-serializable objects to strings recursively
- Fix precision loss for large integers (>2^53-1) by storing them as strings and converting back to int when deserializing
- Update type hints to use union syntax instead of tuple for isinstance
- All tests now pass including test_metadata_with_custom_objects and test_metadata_edge_cases

# Description

### DIFF
--- a/src/a2a/utils/proto_utils.py
+++ b/src/a2a/utils/proto_utils.py
@@ -47,13 +47,52 @@ class ToProto:
         if metadata is None:
             return None
         return struct_pb2.Struct(
-            # TODO: Add support for other types.
             fields={
-                key: struct_pb2.Value(string_value=value)
+                key: cls._convert_value_to_proto(value)
                 for key, value in metadata.items()
-                if isinstance(value, str)
             }
         )
+
+    @classmethod
+    def _convert_value_to_proto(cls, value: Any) -> struct_pb2.Value:
+        """Convert Python value to protobuf Value."""
+        if value is None:
+            proto_value = struct_pb2.Value()
+            proto_value.null_value = 0
+            return proto_value
+        if isinstance(value, bool):
+            return struct_pb2.Value(bool_value=value)
+        if isinstance(value, int):
+            if abs(value) > (2**53 - 1):
+                return struct_pb2.Value(string_value=str(value))
+            return struct_pb2.Value(number_value=float(value))
+        if isinstance(value, float):
+            return struct_pb2.Value(number_value=value)
+        if isinstance(value, str):
+            return struct_pb2.Value(string_value=value)
+        if isinstance(value, dict):
+            serializable_dict = cls._make_json_serializable(value)
+            json_data = json.dumps(serializable_dict, ensure_ascii=False)
+            struct_value = struct_pb2.Struct()
+            json_format.Parse(json_data, struct_value)
+            return struct_pb2.Value(struct_value=struct_value)
+        if isinstance(value, list | tuple):
+            list_value = struct_pb2.ListValue()
+            for item in value:
+                converted_item = cls._convert_value_to_proto(item)
+                list_value.values.append(converted_item)
+            return struct_pb2.Value(list_value=list_value)
+        return struct_pb2.Value(string_value=str(value))
+
+    @classmethod
+    def _make_json_serializable(cls, value: Any) -> Any:
+        if value is None or isinstance(value, str | int | float | bool):
+            return value
+        if isinstance(value, dict):
+            return {k: cls._make_json_serializable(v) for k, v in value.items()}
+        if isinstance(value, list | tuple):
+            return [cls._make_json_serializable(item) for item in value]
+        return str(value)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/a2a/utils/proto_utils.py
```diff
@@ -47,14 +47,53 @@
...
+
+    @classmethod
+    def _make_json_serializable(cls, value: Any) -> Any:
+        if value is None or isinstance(value, str | int | float | bool):
+            return value
+        if isinstance(value, dict):
+            return {k: cls._make_json_serializable(v) for k, v in value.items()}
+        if isinstance(value, list | tuple):
+            return [cls._make_json_serializable(item) for item in value]
+        return str(value)
```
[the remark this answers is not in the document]
holtskinner: @youngchannelforyou Take a look at this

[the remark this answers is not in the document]
youngchannelforyou: I’m done now. Could you check one more for me?

pstephengoogle:
I am not sure we should be trying to infer integers from strings during this translation. It potentially causes incompatibility between the two representations when translating back and forth. If someone is using a string to represent an integer in their struct they should be responsible for parsing this out where needed and it shouldn't be in the translation layer, imho

It seems that you really want a preprocessor on the dict provided in (dict->Struct) and a post processor on the dict produced in (Struct->dict) to apply some normalizations. I think that should be the responsibility of the creators/consumers and not in the proto translation layer itself. Can you provide these as stand alone utility methods?

youngchannelforyou:
@pstephengoogle Thank you for the feedback! I've refactored the proto_utils to remove automatic type inference from the translation layer and provided standalone utility functions for preprocessing/post-processing instead.

src/a2a/utils/proto_utils.py
```diff
@@ -46,6 +46,88 @@ def dict_to_struct(dictionary: dict[str, Any]) -> struct_pb2.Struct:
...
+    Returns:
+        A serializable value.
+    """
+    if isinstance(value, dict):
+        return {k: make_dict_serializable(v) for k, v in value.items()}
+    if isinstance(value, list | tuple):
+        return [make_dict_serializable(item) for item in value]
+    if isinstance(value, str | int | float | bool) or value is None:
+        return value
+    return str(value)
```
holtskinner:
```suggestion
    if isinstance(value, (str, int, float, bool)) or value is None:
        return value
    if isinstance(value, dict):
        return {k: make_dict_serializable(v) for k, v in value.items()}
    if isinstance(value, list | tuple):
        return [make_dict_serializable(item) for item in value]
    return str(value)
```
Minor efficiency improvement, start with base case and use [PEP 695](https://peps.python.org/pep-0695/)

src/a2a/utils/proto_utils.py
```diff
@@ -46,6 +46,88 @@ def dict_to_struct(dictionary: dict[str, Any]) -> struct_pb2.Struct:
...
+            for k, v in value.items()
+        }
+    if isinstance(value, list | tuple):
+        return [
+            normalize_large_integers_to_strings(item, max_safe_digits)
+            for item in value
+        ]
+    if isinstance(value, int) and abs(value) > (10**max_safe_digits - 1):
+        return str(value)
+    return value
```
holtskinner:
```suggestion
    max_safe_int = 10**max_safe_digits - 1

    def _normalize(item: Any) -> Any:
        if isinstance(item, int) and abs(item) > max_safe_int:
            return str(item)
        if isinstance(item, dict):
            return {k: _normalize(v) for k, v in item.items()}
        if isinstance(item, list | tuple):
            return [_normalize(i) for i in item]
        return item

    return _normalize(value)
```

Pre-calculate `max_safe_int` to avoid repeated exponent operations

src/a2a/utils/proto_utils.py
```diff
@@ -46,6 +46,88 @@ def dict_to_struct(dictionary: dict[str, Any]) -> struct_pb2.Struct:
...
+        return [
+            parse_string_integers_in_dict(item, max_safe_digits)
+            for item in value
+        ]
+    if (
+        isinstance(value, str)
+        and value.lstrip('-').isdigit()
+        and len(value.lstrip('-')) > max_safe_digits
+    ):
+        return int(value)
```
holtskinner:
```suggestion
    if isinstance(value, str):
        # Handle potential negative numbers.
        stripped_value = value.lstrip('-')
        if stripped_value.isdigit() and len(stripped_value) > max_safe_digits:
            return int(value)
```

Remove extra call to `.lstrip()`

holtskinner (changes requested):
Minor efficiency suggestions

holtskinner (approved):

holtskinner:
@youngchannelforyou Just update the PR message with the new structure and we're good to merge

youngchannelforyou:
Thank you so much for the code review that inspires novice engineers. I'll come back with better code.

PR #424, opened 2025-08-22
feat: Update proto conversion utilities

pstephengoogle:
# Proto Util improvements

- Improves metadata data support in proto conversions (supports more than just map<string, string> now and supports full json object map<string, any>
- Adds support to convert extension data on agent cards

### DIFF
--- a/src/a2a/utils/proto_utils.py
+++ b/src/a2a/utils/proto_utils.py
@@ -46,14 +46,7 @@ class ToProto:
     ) -> struct_pb2.Struct | None:
         if metadata is None:
             return None
-        return struct_pb2.Struct(
-            # TODO: Add support for other types.
-            fields={
-                key: struct_pb2.Value(string_value=value)
-                for key, value in metadata.items()
-                if isinstance(value, str)
-            }
-        )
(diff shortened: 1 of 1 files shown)


pwwpche (approved):

src/a2a/utils/proto_utils.py
```diff
@@ -916,3 +939,25 @@ def role(cls, role: a2a_pb2.Role) -> types.Role:
...
             case _:
                 return types.Role.agent
+
+
+def dict_to_struct(dictionary: dict[str, Any]) -> struct_pb2.Struct:
+    """Converts a Python dict to a Struct proto.
+
+    Unforunately, using the json_format.ParseDict does not work because this
+    wants the dictionary to be an exact match of the Struct proto with fields
+    and keys and values, not the traditional python dict struture.
```
holtskinner:
```suggestion
    Unfortunately, using `json_format.ParseDict` does not work because this
    wants the dictionary to be an exact match of the Struct proto with fields
    and keys and values, not the traditional Python dict structure.
```

holtskinner (approved):

PR #440, opened 2025-08-28
fix: Prevent client disconnect from stopping task execution

### LINKED ISSUE #296: [Bug]: Agent Execution Cancelled by Client-Side Cancellation During message/stream
### What happened?

Currently, with the default `DefaultRequestHandler` implementation, if a client that sent a `message/stream` request cancels the request or unexpectedly loses connection, the agent's execution is cancelled.

I've confirmed this behavior using the following steps:

1.  Modified `a2a-samples/samples/python/agents/helloworld` to include `await asyncio.sleep(1)` to simulate a slightly longer agent execution time.
2.  Started the A2A server.
3.  Sent a `message/stream` request using `curl` and immediately cancelled it.

    ```
    curl -X POST \
        -i -s \
        --header "Content-Type: application/json" \
        --data "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"message/stream\",\"params\":{\"message\":{\"role\":\"user\",\"parts\":[{\"kind\":\"text\",\"text\":\"テストメッセージ\"}],\"messageId\":\"some-id\"}}}" \
        http://localhost:9999/
    ```

The server displayed the following error:

```
asyncio.exceptions.CancelledError: Cancelled by cancel scope 1074c75d0
```

Considering the existence of the `tasks/resubscribe` RPC method, it seems unintended for agent execution to be cancelled due to a disconnection mid-request. I've confirmed that this behavior does not occur with `message/send` requests.

## Environment

```
❯ uv run python --version
Python 3.11.13
```

a2a-sample commit hash: [3ac8e0b44446edede8f7ea398d48c758285f9eeb](https://github.com/a2aproject/a2a-samples/tree/3ac8e0b44446edede8f7ea398d48c758285f9eeb)

### Relevant log output

```shell

```

### Code of Conduct

ovidiutaralesca:
# Issue

* Client disconnect triggered synchronous cleanup.
* That awaited the producer task, effectively tying producer lifetime to the client connection.
* Reconnecting with `tasks/resubscribe` would not receive further events because the producer had already been forced to finish.

This behaviour no longer raises a `asyncio.exceptions.CancelledError` like claimed in #296 due to this fix: #383, but `tasks/resubscribe` still didn't behave as expected.

# How it's reproduced

In any streaming agent: Simply sending a (longer-running) `message/stream`, disconnecting, and then reconnecting to the task using `tasks/resubscribe` will no longer yield events, even though the task should have been still running.

# Fix

## Code

The fix is an one-liner. Now: 
* Client disconnect schedules cleanup in the background and returns immediately.
* Producer continues; resubscribe taps the existing queue and receives subsequent events.
* Cleanup still runs once the producer completes.

## Tests

**Existing tests:**
* Changed existing tests that asserted on `AgentExecutor.execute` by adding an `asyncio.Event` latch to wait until the background producer hits `execute`.

**New tests:**
* `test_stream_disconnect_then_resubscribe_receives_future_events` -- start streaming, disconnect, resubscribe, and confirm future events are received.
* `test_on_message_send_stream_client_disconnect_triggers_background_cleanup_and_producer_continues` -- to validate that disconnecting is non-blocking, producer continues, and cleanup completes afterward.

Fixes #296

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler.py
+++ b/src/a2a/server/request_handlers/default_request_handler.py
@@ -67,6 +67,7 @@ class DefaultRequestHandler(RequestHandler):
     """
 
     _running_agents: dict[str, asyncio.Task]
+    _background_tasks: set[asyncio.Task]
 
     def __init__(  # noqa: PLR0913
         self,
@@ -102,6 +103,9 @@ class DefaultRequestHandler(RequestHandler):
         # TODO: Likely want an interface for managing this, like AgentExecutionManager.
         self._running_agents = {}
         self._running_agents_lock = asyncio.Lock()
+        # Tracks background tasks (e.g., deferred cleanups) to avoid orphaning
+        # asyncio tasks and to surface unexpected exceptions.
+        self._background_tasks = set()
 
     async def on_get_task(
         self,
@@ -355,10 +359,11 @@ class DefaultRequestHandler(RequestHandler):
             raise
         finally:
             if interrupted_or_non_blocking:
-                # TODO: Track this disconnected cleanup task.
-                asyncio.create_task(  # noqa: RUF006
+                cleanup_task = asyncio.create_task(
                     self._cleanup_producer(producer_task, task_id)
                 )
+                cleanup_task.set_name(f'cleanup_producer:{task_id}')
+                self._track_background_task(cleanup_task)
             else:
                 await self._cleanup_producer(producer_task, task_id)
 
@@ -394,10 +399,11 @@ class DefaultRequestHandler(RequestHandler):
                 )
                 yield event
         finally:
-            # TODO: Track this disconnected cleanup task.
-            asyncio.create_task(  # noqa: RUF006
+            cleanup_task = asyncio.create_task(
                 self._cleanup_producer(producer_task, task_id)
             )
+            cleanup_task.set_name(f'cleanup_producer:{task_id}')
+            self._track_background_task(cleanup_task)
 
     async def _register_producer(
         self, task_id: str, producer_task: asyncio.Task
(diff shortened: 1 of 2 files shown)


tests/server/request_handlers/test_default_request_handler.py
```diff
@@ -1129,6 +1139,202 @@ def sync_get_event_stream_gen_for_prop_test(*args, **kwargs):
...
+    orig_create_task = asyncio.create_task
+
+    def create_task_spy(coro):
+        nonlocal produced_task, cleanup_task
+        task = orig_create_task(coro)
+        if produced_task is None:
+            produced_task = task
+        else:
+            cleanup_task = task
+        return task
```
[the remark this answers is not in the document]
holtskinner: @ovidiutaralesca Please look into this

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -273,9 +244,7 @@ def _validate_task_id_match(self, task_id: str, event_task_id: str) -> None:
         """Validates that agent-generated task ID matches the expected task ID."""
         if task_id != event_task_id:
             logger.error(
-                'Agent generated task_id=%s does not match the RequestContext task_id=%s.',
-                event_task_id,
-                task_id,
+                f'Agent generated task_id={event_task_id} does not match the RequestContext task_id={task_id}.'
```
holtskinner: Please fix this lint error
src/a2a/server/request_handlers/default_request_handler.py:247:17: G004 Logging statement uses f-string

Should just be reverted to previous version

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -112,26 +111,6 @@ async def on_get_task(
...
-            task = Task(
-                id=task.id,
-                context_id=task.context_id,
-                status=task.status,
-                artifacts=task.artifacts,
-                history=limited_history,
-                metadata=task.metadata,
-                kind=task.kind,
-            )
-
```
holtskinner: Why is this being removed?

tests/server/request_handlers/test_default_request_handler.py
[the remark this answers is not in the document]
ovidiutaralesca: Added it!

src/a2a/server/request_handlers/default_request_handler.py
ovidiutaralesca: @holtskinner unstashed something and pushed by mistake. Reverted to the actual fix. You guys are quick 😅

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -385,7 +385,10 @@ async def on_message_send_stream(
                 )
                 yield event
         finally:
-            await self._cleanup_producer(producer_task, task_id)
+            # TODO: Track this disconnected cleanup task.
+            asyncio.create_task(  # noqa: RUF006
+                self._cleanup_producer(producer_task, task_id)
+            )
```
holtskinner: Rather than ignoring the lint error and adding a TODO, can you add the tracking that is needed?

ovidiutaralesca: @holtskinner Yes, I noticed it's used throughout the default request handler. The only reason I could think of is that fire-and-forget method may be **marginally** more memory efficient.
Tracked the tasks. Now we log if they end ungracefully. Wdyt?

holtskinner: Thank you for updating that. (I agree it was throughout the default request handler, but it needed to be addressed)

PR #486, opened 2025-10-01
test: push notification e2e tests

lkawka:
# Description

This PR adds two end-to-end tests for push notifications, designed to verify that the default push notification implementation functions as expected. These tests involve setting up two distinct servers: an agent server that hosts a basic test agent, and a notifications server responsible for receiving notifications from the agent server.

BEGIN_COMMIT_OVERRIDE
test: push notification e2e tests
END_COMMIT_OVERRIDE

### DIFF
--- a/src/a2a/utils/errors.py
+++ b/src/a2a/utils/errors.py
@@ -66,6 +66,17 @@ class ServerError(Exception):
 
         Args:
             error: The specific A2A or JSON-RPC error model instance.
-                   If None, an `InternalError` will be used when formatting the response.
         """
         self.error = error
+
+    def __str__(self) -> str:
+        """Returns a readable representation of the internal Pydantic error."""
+        if self.error is None:
+            return 'None'
+        if self.error.message is None:
+            return self.error.__class__.__name__
+        return self.error.message
+
+    def __repr__(self) -> str:
+        """Returns an unambiguous representation for developers showing how the ServerError was constructed with the internal Pydantic error."""
+        return f'{self.__class__.__name__}({self.error!r})'


tests/e2e/push_notifications/notifications_app.py
```diff
@@ -0,0 +1,54 @@
...
+
+def create_notifications_app() -> FastAPI:
+    """Creates a simple push notification injesting HTTP+REST application."""
+    app = FastAPI()
+    store_lock = asyncio.Lock()
+    store: dict[str, list] = {}
+
+    @app.post('/notifications')
+    async def add_notification(request: Request):
+        """Endpoint for injesting notifications from agents. It receives a JSON
```
mikeas1, line 26: typo: injest -> ingest

tests/e2e/push_notifications/test_default_push_notification_support.py
```diff
@@ -0,0 +1,221 @@
...
+from notifications_app import create_notifications_app
+from utils import (
+    create_app_process,
+    find_free_port,
+    wait_for_server_ready,
+)
+
+
+@pytest.fixture(scope='session')
+def port_lock():
```
mikeas1: I'm not sure I understand why this is necessary. Can you expand the comment a bit to explain what problem this addresses?

tests/e2e/push_notifications/test_default_push_notification_support.py
```diff
@@ -0,0 +1,221 @@
...
+
+@pytest.mark.asyncio
+async def test_notification_triggering_with_in_message_config_e2e(
+    notifications_server: str, agent_server: str, http_client: httpx.AsyncClient
+):
+    """
+    Tests push notification triggering for in-message push notification config.
+    """
+    # Send a message with a push notification config.
+    response = await http_client.post(
```
mikeas1: Any reason to directly send HTTP requests instead of using an A2AClient? Same with constructing the objects as dicts instead of using the data types.

tests/e2e/push_notifications/test_default_push_notification_support.py
```diff
@@ -0,0 +1,221 @@
...
+            },
+            'request': {
+                'messageId': 'r-1',
+                'role': 'ROLE_USER',
+                'content': [{'text': 'Hello Agent!'}],
+            },
+        },
+    )
+    assert response.status_code == 200
+    task_id = response.json()['task']['id']
```
mikeas1: Consider parsing this into a response message rather than working with it as a dict.

tests/e2e/push_notifications/test_default_push_notification_support.py
```diff
@@ -0,0 +1,221 @@
...
+    # Get the task to make sure that further input is required.
+    response = await http_client.get(f'{agent_server}/v1/tasks/{task_id}')
+    assert response.status_code == 200
+    assert response.json()['status']['state'] == 'TASK_STATE_INPUT_REQUIRED'
+
+    # Set a push notification config.
+    response = await http_client.post(
+        f'{agent_server}/v1/tasks/{task_id}/pushNotificationConfigs',
+        json={
+            'parent': f'tasks/{task_id}',
```
mikeas1: Technically you don't specify this parameter in the body, it's taken from the path.

tests/e2e/push_notifications/utils.py
```diff
@@ -0,0 +1,46 @@
...
+def run_server(app, host, port) -> None:
+    """Runs a uvicorn server."""
+    uvicorn.run(app, host=host, port=port, log_level='warning')
+
+
+def wait_for_server_ready(url: str, timeout: int = 10) -> None:
+    """Polls the provided URL endpoint until the server is up."""
+    start_time = time.time()
+    while True:
+        try:
```
mikeas1: nit: instead of the try/catch, you can do:
```py
with contextlib.suppress(httpx.ConnectError):
  with httpx.Client() as client:
     # ...
```

tests/e2e/push_notifications/test_default_push_notification_support.py
lkawka: So, I didn't know which ports were gonna be occupied and just decided to get available ephemeral ports. This lock ensures the server is fully operational before another server attempts to claim a port. I realize that the probability of collision is low and can remove if you think it adds unnecessary complexity.

tests/e2e/push_notifications/test_default_push_notification_support.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    # Create an A2A client with a push notification config.
+    a2a_client = ClientFactory(
+        ClientConfig(
+            supported_transports=[TransportProtocol.http_json],
+            push_notification_configs=[
+                PushNotificationConfig(
+                    id='in-message-config',
+                    url=f'{notifications_server}/notifications',
+                    token=uuid.uuid4().hex,
```
mikeas1: Can you add a TODO that we should validate this token is passed back?

tests/e2e/push_notifications/test_default_push_notification_support.py
mikeas1: So, if I understand correctly:

1. The test starts both a notifications server and an A2A server as separate processes.
2. Each of these servers needs a port to serve on.
3. The test needs to know what the port is so that it can connect to it.
4. So we find a port on the testing process side by binding a socket, identifying the port that was selected, then tearing down that socket.
5. The lock is to ensure that we do this port finding process serially

Since the port finding happens on the test harness side, I think it's guaranteed that it will happen serially given Python is single-threaded? The method isn't async, so I don't believe it can be interrupted between finding the port and using it. Therefore we won't need the lock. If each sub-process chose a port I believe it would be needed.

lkawka: You're right, thanks for the detailed explanation!

tests/e2e/push_notifications/test_default_push_notification_support.py
lkawka: I added the verification of these tokens to the tests

PR #489, opened 2025-10-02
fix: Add `__str__` and `__repr__` methods to `ServerError`

aneeshgarg:
Add `__str__` and `__repr__` methods to `ServerError` and fix the misleading comment in the constructor

### DIFF
--- a/src/a2a/utils/errors.py
+++ b/src/a2a/utils/errors.py
@@ -74,7 +74,7 @@ class ServerError(Exception):
         if self.error is None:
             return 'None'
         if self.error.message is None:
-            return f'{type(self.error)}'
+            return self.error.__class__.__name__
         return self.error.message
 
     def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


src/a2a/utils/errors.py
```diff
@@ -66,6 +66,17 @@ def __init__(
...
             error: The specific A2A or JSON-RPC error model instance.
-                   If None, an `InternalError` will be used when formatting the response.
         """
         self.error = error
+
+    def __str__(self) -> str:
+        """Returns a readable representation of the internal Pydantic error."""
+        if self.error is None:
+            return 'None'
+        if self.error.message is None:
```
mikeas1, line 76: How about something like `f{self.__class__.__name__}({self.error})`, which looks like `ServerError(JSONParseError(...))`.

aneeshgarg: The main reason I did not do it is because I wanted to keep that structured message for __repr__. This should be user friendly (less developer'y if there is such a word).

PR #509, opened 2025-10-15
feat: Add a `ClientFactory.connect()` method for easy client creation

mikeas1:
# Description

This PR adds a convenience method for constructing a Client from either an AgentCard URL or an AgentCard directly. The goal is to reduce the number of lines of code required for simple client creation, but still enabling more advanced handled of client construction.

Usage example:

```python
my_agent_url =  'https://travel-agent.example.com'
client = await ClientFactory.connect(my_agent_url)
await client.send_message(...)
```

BEGIN_COMMIT_OVERRIDE
feat: Add a `ClientFactory.connect()` method for easy client creation
Release-As: 0.3.10
END_COMMIT_OVERRIDE

### DIFF
--- a/src/a2a/client/client_factory.py
+++ b/src/a2a/client/client_factory.py
@@ -120,6 +120,13 @@ class ClientFactory:
         creating multiple clients via this method is less efficient than
         constructing an instance of ClientFactory and reusing that.
 
+        .. code-block:: python
+
+            # This will search for an AgentCard at /.well-known/agent-card.json
+            my_agent_url = 'https://travel.agents.example.com'
+            client = await ClientFactory.connect(my_agent_url)
+
+
         Args:
           agent: The base URL of the agent, or the AgentCard to connect to.
           client_config: The ClientConfig to use when connecting to the agent.


src/a2a/client/client_factory.py
```diff
@@ -101,6 +103,64 @@ def _register_defaults(
...
+        interceptors: list[ClientCallInterceptor] | None = None,
+        relative_card_path: str | None = None,
+        resolver_http_kwargs: dict[str, Any] | None = None,
+        extra_transports: dict[str, TransportProducer] | None = None,
+    ) -> Client:
+        """Convenience method for constructing a client.
+
+        Constructs a client that connects to the specified agent. Note that
+        creating multiple clients via this method is less efficient than
+        constructing an instance of ClientFactory and reusing that.
```
holtskinner, line 121: Can you add a code sample in this docstring like in the PR Message? Either on the `ClientFactory` itself or for this specific method

Follow format of https://github.com/a2aproject/a2a-python/blob/317df0a2a4b4e405da6af57eaf5edd9f0df7eec2/src/a2a/client/client_factory.py#L44-L51

PR #522, opened 2025-10-24
refactor: Separate protobuf dependencies into sensible extras

vinoo999:
# Description

This introduces a `proto` extra into the package. The core library itself has no direct dependency on protobuf and only rest server and clients require a2a. google-api-core is only a dependency for grpc. Refactor accordingly to reduce dependency mangling introduced by proto for non grpc/rest servers. Assuming a CI regenerates the uv.lock file but can regenerate it otherwise. 

This does not follow the runtime validation pattern in https://github.com/a2aproject/a2a-python/pull/217 as proto dependencies are localized in `proto_utils.py`.


pyproject.toml
```diff
@@ -29,9 +27,11 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-http-server = ["fastapi>=0.115.2", "sse-starlette", "starlette"]
+proto = ["protobuf>=5.29.5"]
```
holtskinner: I'm not sure protobuf really needs to be a separate dependency group from grpc

README.md
```diff
@@ -48,7 +48,9 @@ Install the core SDK and any desired extras using your preferred package manager
 | ------------------------ | ------------------------------------------ | -------------------------------------------- |
 | **Core SDK**             | `uv add a2a-sdk`                           | `pip install a2a-sdk`                        |
 | **All Extras**           | `uv add "a2a-sdk[all]"`                    | `pip install "a2a-sdk[all]"`                 |
-| **HTTP Server**          | `uv add "a2a-sdk[http-server]"`            | `pip install "a2a-sdk[http-server]"`         |
+| **protobuf**             | `uv add "a2a-sdk[proto]"`                  | `pip install "a2a-sdk[proto]"`               |
+| **JSON-RPC Server**      | `uv add "a2a-sdk[jsonrpc-server]"`         | `pip install "a2a-sdk[jsonrpc-server]"`         |
+| **HTTP (Rest) Server**   | `uv add "a2a-sdk[http-server]"`            | `pip install "a2a-sdk[http-server]"`         |
```
holtskinner:
```suggestion
| **HTTP (REST) Server**   | `uv add "a2a-sdk[http-server]"`            | `pip install "a2a-sdk[http-server]"`         |
```

holtskinner (changes requested):

pyproject.toml
vinoo999: can remove the extra altogether if you want to be explicit for the REST and gRPC servers. My goal was removing protobuf dependencies especially for JSON-RPC servers + clients. 

If you prefer can do:
```
jsonrpc-server = ["fastapi>=0.115.2", "sse-starlette", "starlette"]
http-server = ["a2a-sdk[jsonrpc-server]", "protobuf>=5.29.5"]
grpc = ["grpcio>=1.60", "grpcio-tools>=1.60", "grpcio_reflection>=1.7.0", "google-api-core>=1.26.0", "protobuf>=5.29.5"]

```

vinoo999: http-server is a bit confusing naming of the extra now that JSON-RPC is split out but needed for backward compatibility

holtskinner:
@pstephengoogle Can you review and make sure this follows with current plans for the protocol?

sambhav:
@pstephengoogle / @holtskinner any updates on this?

.github/actions/spelling/allow.txt
lkawka, line 1: Why make changes to this file? It appears that entires were in an alphabetical order here.

README.md
```diff
@@ -48,7 +48,8 @@ Install the core SDK and any desired extras using your preferred package manager
 | ------------------------ | ------------------------------------------ | -------------------------------------------- |
 | **Core SDK**             | `uv add a2a-sdk`                           | `pip install a2a-sdk`                        |
 | **All Extras**           | `uv add "a2a-sdk[all]"`                    | `pip install "a2a-sdk[all]"`                 |
-| **HTTP Server**          | `uv add "a2a-sdk[http-server]"`            | `pip install "a2a-sdk[http-server]"`         |
+| **JSON-RPC Server**      | `uv add "a2a-sdk[jsonrpc-server]"`         | `pip install "a2a-sdk[jsonrpc-server]"`         |
+| **HTTP (REST) Server**   | `uv add "a2a-sdk[http-server]"`            | `pip install "a2a-sdk[http-server]"`         |
```
lkawka: Let's copy the terminology from the spec documentation: **HTTP+JSON/REST Server**

pyproject.toml
```diff
@@ -29,9 +27,10 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-http-server = ["fastapi>=0.115.2", "sse-starlette", "starlette"]
+jsonrpc-server = ["fastapi>=0.115.2", "sse-starlette", "starlette"]
+http-server = ["a2a-sdk[jsonrpc-server]", "protobuf>=5.29.5"]
```
lkawka, line 31: HTTP+JSON/REST also requires these dependencies. 

rest_handler.py depends on a2a_pb2: https://github.com/a2aproject/a2a-python/blob/89e9b7cb0ab24c3c22b7da04e04d77ec5f15dc1c/src/a2a/server/request_handlers/rest_handler.py#L18

.github/actions/spelling/allow.txt
vinoo999: unclear. i think this came directly from the github action when updating readme. i can revert

README.md
vinoo999:
```suggestion
| **HTTP+JSON/REST Server**   | `uv add "a2a-sdk[http-server]"`            | `pip install "a2a-sdk[http-server]"`         |
```

pyproject.toml
vinoo999: This is already encompassed here. a2a.grpc.a2a_pb2 has no grpcio dependencies. Just protobuf. This is encapsulated in the PR already.

Any concerns here?

.github/actions/spelling/allow.txt
vinoo999: in fact I think this may have come from the Format script which must have a bug for capitalization

vinoo999: https://github.com/a2aproject/a2a-python/pull/550### to address

vinoo999: reran with ^ and applied. removed postres dup

pyproject.toml
lkawka: `a2a_pb2` seems to depend on `google-api-core`:
https://github.com/a2aproject/a2a-python/blob/89e9b7cb0ab24c3c22b7da04e04d77ec5f15dc1c/src/a2a/grpc/a2a_pb2.py#L26

I think we need to include it here as well.

.github/actions/spelling/allow.txt
lkawka: Can we leave it as it was? I'm not entirely familiar with how this file is used and I worry that dependents may be case-sensitive.

holtskinner: This was intentionally left case-sensitive, because the case-insensitive sort wasn't always consistent.

pyproject.toml
vinoo999: You're right. I'll add some tests that do the individual extras plus tests accordingly. I'm going to move this PR to draft because it looks like 1.0 is going to require some changes that depend directly on the proto rather than the a2a spec json even in jsonrpc. In doing so this split may not be directly possible in the way I'd like.

vinoo999:
Moving to draft. pending 1.0 will see if this is still possible. based on discussion the proto dependency might be required in all protocol bindings.

ishymko:
As 1.0 version uses proto generated types as a domain model, it's no longer feasible.

Thanks @vinoo999!

PR #525, opened 2025-10-28
feat: Add client-side extension support

### LINKED ISSUE #504: [Feat]: Extension support for Client

Right now if I want to support an A2A extension in the `Client` I have to create a custom interceptor. Here's what I've resolved to doing:
```python
from typing import Any

from a2a.client import ClientCallContext, ClientCallInterceptor
from a2a.extensions.common import HTTP_EXTENSION_HEADER
from a2a.types import AgentCard


class ExtensionsInterceptor(ClientCallInterceptor):
    def __init__(self, extensions: list[str]):
        """
        Args:
            extensions: List of extension URIs to signal support for.
        """
        self.extensions = extensions

    async def intercept(
        self,
        method_name: str,
        request_payload: dict[str, Any],
        http_kwargs: dict[str, Any],
        agent_card: AgentCard | None,
        context: ClientCallContext | None,
    ) -> tuple[dict[str, Any], dict[str, Any]]:
        """Add extension header to HTTP request."""
        if self.extensions:
            headers = http_kwargs.get("headers", {})
            existing_extensions = headers.get(HTTP_EXTENSION_HEADER, "")
            split = existing_extensions.split(", ") if existing_extensions else []
            headers[HTTP_EXTENSION_HEADER] = ", ".join(self.extensions + split)
            http_kwargs["headers"] = headers

        return request_payload, http_kwargs
```
and then set that interceptor when creating a `Client`

Ideally extension support would be first class in the `Client`. Maybe add it to `ClientConfig` like:
```python
class ClientConfig:
    ...
    extensions: list[str]
```

Using the interceptor described above.

### Additional context

This is required for the A2A <> x402 integration.

### Code of Conduct

sokoliva:
feat(jsonrpc, rest, grpc): Add client-side extension support

**This commit introduces support for clients to declare the extensions they support.**

- Adds an `extensions` list to ClientConfig.
- Updates ClientFactory to pass extensions to JsonRpcTransport, RestTransport and GrpcTransport.
- Adds `update_extension_header` function to a2a-python/src/a2a/extensions/common.py.
- Modifies  JsonRpcTransport and  RestTransport methods to include the extensions header.
- Adds a private method `_get_grpc_metadata` to GrpcTransport.
- Modifies  GrpcTransport methods to include the extension metadata.
- Adds tests for the new client side extensions logic in JSON-RPC, REST and gRPC transports, including a new test file `test_rest_client.py`.
- Adds extensions list as input variable to BaseClient methods.

Fixes #504 🦕

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.3.14](https://github.com/a2aproject/a2a-python/compare/v0.3.13...v0.3.14) (2025-11-17)
+
+
+### Features
+
+* **jsonrpc:** add option to disable oversized payload check in JSONRPC applications ([ba142df](https://github.com/a2aproject/a2a-python/commit/ba142df821d1c06be0b96e576fd43015120fcb0b))
+
 ## [0.3.13](https://github.com/a2aproject/a2a-python/compare/v0.3.12...v0.3.13) (2025-11-13)
 
 
--- a/src/a2a/client/__init__.py
+++ b/src/a2a/client/__init__.py
@@ -7,6 +7,7 @@ from a2a.client.auth import (
     CredentialService,
     InMemoryContextCredentialStore,
 )
+from a2a.client.base_client import BaseClient
 from a2a.client.card_resolver import A2ACardResolver
 from a2a.client.client import Client, ClientConfig, ClientEvent, Consumer
 from a2a.client.client_factory import ClientFactory, minimal_agent_card
@@ -51,6 +52,7 @@ __all__ = [
     'A2AClientTimeoutError',
     'A2AGrpcClient',
     'AuthInterceptor',
+    'BaseClient',
     'Client',
     'ClientCallContext',
     'ClientCallInterceptor',
--- a/src/a2a/server/apps/jsonrpc/fastapi_app.py
+++ b/src/a2a/server/apps/jsonrpc/fastapi_app.py
@@ -77,6 +77,7 @@ class A2AFastAPIApplication(JSONRPCApplication):
             [AgentCard, ServerCallContext], AgentCard
         ]
         | None = None,
+        max_content_length: int | None = 10 * 1024 * 1024,  # 10MB
     ) -> None:
         """Initializes the A2AFastAPIApplication.
 
@@ -94,6 +95,8 @@ class A2AFastAPIApplication(JSONRPCApplication):
             extended_card_modifier: An optional callback to dynamically modify
               the extended agent card before it is served. It receives the
               call context.
+            max_content_length: The maximum allowed content length for incoming
+              requests. Defaults to 10MB. Set to None for unbounded maximum.
         """
         if not _package_fastapi_installed:
             raise ImportError(
@@ -108,6 +111,7 @@ class A2AFastAPIApplication(JSONRPCApplication):
             context_builder=context_builder,
             card_modifier=card_modifier,
             extended_card_modifier=extended_card_modifier,
+            max_content_length=max_content_length,
         )
 
     def add_routes_to_app(
--- a/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
+++ b/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
@@ -91,8 +91,6 @@ else:
         Response = Any
         HTTP_413_REQUEST_ENTITY_TOO_LARGE = Any
 
-MAX_CONTENT_LENGTH = 10_000_000
-
 
 class StarletteUserProxy(A2AUser):
     """Adapts the Starlette User class to the A2A user representation."""
@@ -185,6 +183,7 @@ class JSONRPCApplication(ABC):
             [AgentCard, ServerCallContext], AgentCard
         ]
         | None = None,
+        max_content_length: int | None = 10 * 1024 * 1024,  # 10MB
     ) -> None:
         """Initializes the JSONRPCApplication.
 
@@ -202,6 +201,8 @@ class JSONRPCApplication(ABC):
             extended_card_modifier: An optional callback to dynamically modify
               the extended agent card before it is served. It receives the
               call context.
+            max_content_length: The maximum allowed content length for incoming
+              requests. Defaults to 10MB. Set to None for unbounded maximum.
         """
         if not _package_starlette_installed:
             raise ImportError(
@@ -220,6 +221,7 @@ class JSONRPCApplication(ABC):
             extended_card_modifier=extended_card_modifier,
         )
         self._context_builder = context_builder or DefaultCallContextBuilder()
+        self._max_content_length = max_content_length
 
     def _generate_error_response(
         self, request_id: str | int | None, error: JSONRPCError | A2AError
@@ -261,6 +263,22 @@ class JSONRPCApplication(ABC):
             status_code=200,
         )
 
+    def _allowed_content_length(self, request: Request) -> bool:
+        """Checks if the request content length is within the allowed maximum.
+
+        Args:
+            request: The incoming Starlette Request object.
+
+        Returns:
+            False if the content length is larger than the allowed maximum, True otherwise.
+        """
+        if self._max_content_length is not None:
+            with contextlib.suppress(ValueError):
+                content_length = int(request.headers.get('content-length', '0'))
+                if content_length and content_length > self._max_content_length:
+                    return False
+        return True
+
     async def _handle_requests(self, request: Request) -> Response:  # noqa: PLR0911
         """Handles incoming POST requests to the main A2A endpoint.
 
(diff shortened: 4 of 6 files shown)


src/a2a/client/transports/jsonrpc.py
```diff
@@ -71,26 +73,41 @@
...
+        if self.client_extensions:
+            headers = http_kwargs.get('headers', {})
+            existing_extensions = headers.get(HTTP_EXTENSION_HEADER, '')
+            split = (
+                existing_extensions.split(', ') if existing_extensions else []
+            )
+            updated_extensions = list(set(self.client_extensions + split))
+            headers[HTTP_EXTENSION_HEADER] = ', '.join(updated_extensions)
+            http_kwargs['headers'] = headers
+        return http_kwargs
```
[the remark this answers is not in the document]
sokoliva: Parsing Bug is fixed. Code Duplication will be fixed in future if needed.

tests/client/test_jsonrpc_client.py
```diff
@@ -785,3 +786,181 @@ async def test_close(self, mock_httpx_client: AsyncMock):
...
+            HTTP_EXTENSION_HEADER
+        ].split(', ')
+        actual_extensions = set(actual_extensions_list)
+        expected_extensions = {
+            'test_extension_1',
+            'test_extension_2',
+            'test_extension_3',
+        }
+        assert len(actual_extensions_list) == 3
+        assert actual_extensions == expected_extensions
```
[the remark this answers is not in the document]
sokoliva: Done

src/a2a/client/transports/jsonrpc.py
```diff
@@ -79,6 +81,26 @@ def __init__(
             if agent_card
             else True
         )
+        self.client_extensions = client_extensions
```
lkawka: Extensions touch on both client and server, and given that we are in the context of Client, I believe that the `client_` prefixes are unnecessary.

src/a2a/client/transports/rest.py
```diff
@@ -61,6 +63,26 @@ def __init__(
...
+        existing_extensions_str = headers.get(HTTP_EXTENSION_HEADER, '')
+
+        existing_extensions = [
+            e.strip() for e in existing_extensions_str.split(',') if e.strip()
+        ]
+
+        all_extensions = set(self.client_extensions)
+        all_extensions.update(existing_extensions)
+
+        headers[HTTP_EXTENSION_HEADER] = ', '.join(list(all_extensions))
```
lkawka: ', ' -> ',' will reduce the payload size a tiny bit

lkawka (commented):
Two things:
- gRPC implementation seems to be missing.
- We would like to have per-method overrides on the `Client` interface.

src/a2a/client/transports/grpc.py
```diff
@@ -106,8 +109,10 @@ async def send_message_streaming(
                 configuration=proto_utils.ToProto.message_send_configuration(
                     request.configuration
                 ),
-                metadata=proto_utils.ToProto.metadata(request.metadata),
-            )
+                metadata=update_extension_metadata(
+                    request.metadata, self.extensions
+                ),
```
lkawka: This is incorrect.

We don't want to put the list of activated extensions inside the main payload. This needs to be put in the side-channel (i.e. headers for HTTP and [metadata](https://grpc.io/docs/guides/metadata/) for gRPC). 

This also applies to other places in this file.

src/a2a/client/transports/grpc.py
```diff
@@ -54,6 +57,18 @@ def __init__(
...
+    def _get_grpc_metadata(
+        self,
+        extensions: list[str] | None = None,
+    ) -> list[tuple[str, str]] | None:
+        """Creates gRPC metadata for extensions."""
+        if extensions:
+            self.extensions = extensions
+        if not self.extensions:
+            return None
+        return [(HTTP_EXTENSION_HEADER, ', '.join(self.extensions))]
```
lkawka: nit: ', ' -> ',' it will take less space

src/a2a/client/transports/grpc.py
```diff
@@ -54,6 +57,18 @@ def __init__(
...
             else True
         )
+        self.extensions = extensions
+
+    def _get_grpc_metadata(
+        self,
+        extensions: list[str] | None = None,
+    ) -> list[tuple[str, str]] | None:
+        """Creates gRPC metadata for extensions."""
+        if extensions:
```
lkawka: Two things:
1. I would do `if extensions is not None` that way we can override the default list from config to an empty list by passing `[]`.
1. Why are you assigning this value to `self.extensions`? This will make the override stick to subsequent call which is not what we want.

src/a2a/client/transports/grpc.py
```diff
@@ -180,19 +207,22 @@ async def get_task_callback(
...
+            ),
+            metadata=self._get_grpc_metadata(extensions),
         )
         return proto_utils.FromProto.task_push_notification_config(config)
 
     async def get_card(
         self,
         *,
         context: ClientCallContext | None = None,
+        extensions: list[str] | None = None,
```
lkawka, line 225: This appears to be unused

src/a2a/client/transports/jsonrpc.py
```diff
@@ -122,6 +126,9 @@ async def send_message(
             self._get_http_args(context),
             context,
         )
+        modified_kwargs, self.extensions = update_extension_header(
```
lkawka: Similar issue to above. I think we shouldn't override the internal state (`self.extensions`).

src/a2a/extensions/common.py
```diff
@@ -25,3 +27,23 @@ def find_extension_by_uri(card: AgentCard, uri: str) -> AgentExtension | None:
...
+    if new_extensions:
+        active_extensions = new_extensions
+    if active_extensions:
+        headers = http_kwargs.setdefault('headers', {})
+        existing_extensions_str = headers.get(HTTP_EXTENSION_HEADER, '')
+
+        existing_extensions = get_requested_extensions(
+            [existing_extensions_str]
+        )
+        all_extensions = existing_extensions.union(active_extensions)
```
lkawka: Why do you assume that we have to update existing extensions and not just add new ones? I think it's safe to assume that we can just always set the value to what's passed in the `extensions` list.

sokoliva: Ack

src/a2a/client/transports/jsonrpc.py
sokoliva: Ack

src/a2a/client/transports/grpc.py
sokoliva: Added extension handling within get_card method.

src/a2a/client/transports/grpc.py
sokoliva: Agreed, thank you.

src/a2a/client/transports/grpc.py
sokoliva: Ack

src/a2a/extensions/common.py
```diff
@@ -25,3 +27,14 @@ def find_extension_by_uri(card: AgentCard, uri: str) -> AgentExtension | None:
...
     return None
+
+
+def update_extension_header(
+    http_kwargs: dict[str, Any],
+    extensions: list[str] | None,
+) -> dict[str, Any]:
+    """Update the X-A2A-Extensions header and update active extensions."""
+    if extensions is not None:
+        headers = http_kwargs.setdefault('headers', {})
```
lkawka, line 39: I think this line isn't needed.

src/a2a/extensions/common.py
```diff
@@ -25,3 +27,14 @@ def find_extension_by_uri(card: AgentCard, uri: str) -> AgentExtension | None:
             return ext
 
     return None
+
+
+def update_extension_header(
+    http_kwargs: dict[str, Any],
+    extensions: list[str] | None,
+) -> dict[str, Any]:
+    """Update the X-A2A-Extensions header and update active extensions."""
```
lkawka: I guess it's only the first part of the sentence now.

sokoliva: FIxed

src/a2a/extensions/common.py
sokoliva: Resolved in person. Conclusion: it is needed. It returns the headers in http_kwargs (http_kwargs.get('headers', {})) and if http_kwargs does not contain headers it sets it. I added a test for the latter case in test_common.py.

src/a2a/client/transports/jsonrpc.py
```diff
@@ -216,6 +230,10 @@ async def get_task(
             self._get_http_args(context),
             context,
         )
+        modified_kwargs = update_extension_header(
+            modified_kwargs,
+            extensions if extensions is not None else self.extensions,
+        )
```
lkawka, line 235: I should have pointed it out earlier. I think it might make sense to put this before the `self._apply_interceptors`, giving interceptors access to the list of extensions.

tests/integration/test_client_server_integration.py
```diff
@@ -767,3 +769,58 @@ def channel_factory(address: str) -> Channel:
...
+            request=message_to_send, extensions=extensions
+        ):
+            pass
+
+        mock_send_request.assert_called_once()
+        call_args, _ = mock_send_request.call_args
+        kwargs = call_args[1]
+        headers = kwargs.get('headers', {})
+        assert 'X-A2A-Extensions' in headers
+        assert headers['X-A2A-Extensions'] == ','.join(extensions)
```
lkawka: Nit: It's better to use literals in assertions, eliminating the ambiguity of which side is incorrect in case of an error.

```python
assert headers['X-A2A-Extensions'] == 'https://example.com/test-ext/v1,https://example.com/test-ext/v2'
```

src/a2a/client/transports/jsonrpc.py
sokoliva: Done

tests/integration/test_client_server_integration.py
sokoliva: Done

PR #527, opened 2025-10-31
feat: introduce TransportProtocol validation

martimfasantos:
This pull request improves the way transport protocols are used by some of the classes in A2A agent types by updating relevant fields to use it instead of plain strings.

**Transport protocol type safety improvements:**

* Updated the `transport` field in the `AgentInterface` class to use the new `TransportProtocol` enum instead of a string, including updating example values.
* Updated the `preferred_transport` field in the `AgentCard` class to use the new `TransportProtocol` enum, set its default to `TransportProtocol.jsonrpc`, and updated example values.

### DIFF
--- a/tests/client/test_auth_middleware.py
+++ b/tests/client/test_auth_middleware.py
@@ -106,10 +106,10 @@ def store():
 
 
 @pytest.mark.asyncio
-async def test_auth_interceptor_skips_when_no_agent_card(store):
-    """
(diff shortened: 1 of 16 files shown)


lkawka:
Hi [martimfasantos](https://github.com/martimfasantos),

Unfortunately, we have an automation that generates these types from the specification and we don't want to make any manual changes here.

PR #544, opened 2025-11-14
feat(jsonrpc): add option to disable oversized payload check in JSONRPC applications

sokoliva:
This commit adds a constructor boolean variable disable_content_length_check to the base class JSONRPCApplication and its derived classes A2AFastAPIApplication and A2AStarletteApplication. In JSONRPCApplication's method _handle_requests it adds a if disable_content_length_check check before checking if the payload exceeds the MAX_CONTENT_LENGTH limit.

This enables agent creators to disable the 10MB payload size limitation.

BEGIN_COMMIT_OVERRIDE
feat(jsonrpc): add option to disable oversized payload check in JSONRPC applications

Release-As: 0.3.14
END_COMMIT_OVERRIDE

### DIFF
--- a/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
+++ b/src/a2a/server/apps/jsonrpc/jsonrpc_app.py
@@ -132,7 +132,7 @@ class DefaultCallContextBuilder(CallContextBuilder):
         """
         user: A2AUser = UnauthenticatedUser()
         state = {}
-        with contextlib.suppress(AttributeError):
+        with contextlib.suppress(Exception):
             user = StarletteUserProxy(request.user)
             state['auth'] = request.auth
         state['headers'] = dict(request.headers)


src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -291,18 +295,22 @@
...
-                if content_length and content_length > MAX_CONTENT_LENGTH:
-                    return self._generate_error_response(
-                        request_id,
-                        A2AError(
-                            root=InvalidRequestError(
-                                message='Payload too large'
-                            )
-                        ),
+            # If content lenght check is not diasbled,
+            # treat very large payloads as invalid request (-32600) before routing
```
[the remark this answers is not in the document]
sokoliva: fixed

src/a2a/server/apps/jsonrpc/starlette_app.py
```diff
@@ -76,6 +77,8 @@
             extended_card_modifier: An optional callback to dynamically modify
               the extended agent card before it is served. It receives the
               call context.
+            disable_content_length_check: An optional, if True disables the check
+            for oversized payloads.
```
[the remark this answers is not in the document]
sokoliva: fixed

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -261,6 +265,22 @@ def _generate_error_response(
...
+        """Checks if the request content length exceeds the maximum allowed size.
+
+        Args:
+            request: The incoming Starlette Request object.
+
+        Returns:
+            True if the content length is within the allowed limit, False otherwise.
+        """
+        if not self._disable_content_length_check:
+            with contextlib.suppress(Exception):
```
lkawka: This catch clause it very broad. Can we change it to a more specific exception?

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -261,6 +265,22 @@ def _generate_error_response(
             status_code=200,
         )
 
+    def _check_content_length(self, request: Request) -> bool:
```
lkawka: Nit: I think `_allowed_content_length` would read better.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -202,6 +203,8 @@ def __init__(  # noqa: PLR0913
             extended_card_modifier: An optional callback to dynamically modify
               the extended agent card before it is served. It receives the
               call context.
+            disable_content_length_check: An optional, if True disables the check
```
yarolegovich: instead of allowing to disable the check let's make the `max_content_length` configurable. this way it's more flexible. developers can provide a big value if they want to effectively disable it

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
sokoliva: Yes. Changed to "ValueError".

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
sokoliva: Noted.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
sokoliva: Done. "None" is used for disabling.

PR #549, opened 2025-11-18
fix(json-rpc): add stream_send_timeout for JSON-RPC streaming endpoints

### LINKED ISSUE #545: [Bug]: JSON-RPC streaming endpoints hard‑close when send() blocks longer than Starlette’s default timeout
### Current behavior

`JSONRPCApplication` always builds `EventSourceResponse` with Starlette’s defaults (`src/a2a/server/apps/jsonrpc/jsonrpc_app.py:323`). When an agent pauses longer than the default `send_timeout`, the next write raises `SendTimeoutError`, the response task exits, and the client never receives the remaining events. Long-lived agents tasks hit this path while waiting on external work.

### Expected/desired behavior

Developers using the SDK should be able to configure a larger timeout or disable it entirely without subclassing the server. Semi-idle agents should keep their SSE channel open, even if send() occasionally blocks longer than five seconds.

### Proposed fix

1. Add a `stream_send_timeout: float | None` option on `JSONRPCApplication` (plumbed through `A2AStarletteApplication`/`A2AFastAPIApplication`) and pass it to `EventSourceResponse(...)`. Preserve the current default when the option is omitted.
1. Allow per-request overrides via `ServerCallContext.state["stream_send_timeout"]` so specific handlers can raise/lower the ceiling.
1. Document the new control (and the failure mode) in the SDK guide, recommending that long-running agents either disable or adjust the timeout, or emit frequent heartbeat events.

### Relevant log output

```shell

```

### Code of Conduct

devesh1011:
# Description
This PR fixes issue #545 where JSON-RPC streaming endpoints were closing prematurely due to Starlette's default 5-second send timeout when agents take longer to respond.

## Problem
JSON-RPC streaming connections using `EventSourceResponse` were being closed after 5 seconds by Starlette's default `send_timeout`, causing issues when agents need more time to process and respond to requests.

## Solution
Added a configurable `stream_send_timeout` parameter to the JSON-RPC application classes:

Fixes #545  🦕


sokoliva:
Obsolete: affected files no longer exist

PR #552, opened 2025-11-19
fix: return updated `agent_card` in `JsonRpcTransport.get_card()`

### LINKED ISSUE #551: [Bug]: JsonRpcTransport.get_card() returns stale card instead of authenticated extended card
### What happened?

The get_card() method in JsonRpcTransport returns the wrong card when fetching an authenticated extended card. It returns the old card instead of the newly fetched one.

The Problem is that when an authenticated extended card is fetched, the method:
Correctly fetches the extended card from server
Correctly updates self.agent_card = response.root.result
❌ Incorrectly returns the OLD card variable

this is the section where it has issue:

https://github.com/ShishirRmc/a2a-python/blob/051ab20c395daa2807b0233cf1c53493e41b60c2/src/a2a/client/transports/jsonrpc.py#L338-L377

evidence: https://github.com/ShishirRmc/a2a-python/blob/051ab20c395daa2807b0233cf1c53493e41b60c2/src/a2a/client/transports/rest.py#L358-L361

The issue Defeats the entire purpose of fetching authenticated extended cards.
I have a one line fix ready and can submit a PR if you'd like.

### Relevant log output

```shell
No runtime logs available, this is a code logic bug found through static analysis.

The method updates self.agent_card with the new extended card but returns 
the stale 'card' variable instead. This is inconsistent with RestTransport's 
correct implementation.
```

### Code of Conduct

ShishirRmc:
### Changes
- Fixed `JsonRpcTransport.get_card()` to return the newly fetched authenticated extended card instead of the stale card

### Details
Changed line from `return card` to `return self.agent_card` to ensure the method returns the updated card after fetching the authenticated extended version.

This aligns the JsonRpcTransport behavior with RestTransport's correct implementation.

### Testing
- Verified the fix matches RestTransport's pattern
- Confirmed internal state and return value are now consistent

Release-As: 0.3.18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## [0.3.17](https://github.com/a2aproject/a2a-python/compare/v0.3.16...v0.3.17) (2025-11-24)
+
+
+### Features
+
+* **client:** allow specifying `history_length` via call-site `MessageSendConfiguration` in `BaseClient.send_message` ([53bbf7a](https://github.com/a2aproject/a2a-python/commit/53bbf7ae3ad58fb0c10b14da05cf07c0a7bd9651))
+
+## [0.3.16](https://github.com/a2aproject/a2a-python/compare/v0.3.15...v0.3.16) (2025-11-21)
+
+
+### Bug Fixes
+
+* Ensure metadata propagation for `Task` `ToProto` and `FromProto` conversion ([#557](https://github.com/a2aproject/a2a-python/issues/557)) ([fc31d03](https://github.com/a2aproject/a2a-python/commit/fc31d03e8c6acb68660f6d1924262e16933c5d50))
+
+## [0.3.15](https://github.com/a2aproject/a2a-python/compare/v0.3.14...v0.3.15) (2025-11-19)
+
+
+### Features
+
+* Add client-side extension support ([#525](https://github.com/a2aproject/a2a-python/issues/525)) ([9a92bd2](https://github.com/a2aproject/a2a-python/commit/9a92bd238e7560b195165ac5f78742981760525e))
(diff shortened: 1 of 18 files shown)


holtskinner:
@ShishirRmc Can you adjust the unit tests to check for this new behavior. This is also technically a breaking change, even if you have now corrected the intended behavior.

@lkawka @mikeas1 Not sure if this behavior is intended or not, can you verify?

sokoliva:
> @lkawka @mikeas1 Not sure if this behavior is intended or not, can you verify?

@holtskinner I didn't find any evidence that this is intended, but I did find a couple of reasons to believe it is **not**, suggesting a mistake was made.

1. There is nothing to suggest that this is the intended behavior in the [spec](https://a2a-protocol.org/latest/)
2.  In both `RestTransport` and `GrpcTransport` get_card retrieves newly fetched authenticated extended card.`JsonRpcTransport` is the only one in which `get_card` returns the `stale card`.
3. `get_card` methods of all three transports all have the same description which is **"Retrieves the agent's card."** and method of base class `ClientTransport` has **"Retrieves the AgentCard."**. I would assume that their functionality is the same - to retrieve the **current** `agent_card`, not the former one. 
4.  When `_needs_extended_card=True`, in `get_card` methods of `RestTransport` and `GrpcTransport`, the values of variables `card` and `self.aget_card` are the same at point of return. This means it makes no difference whether `card` or `self.agent_card` is returned. JsonRpcTransport is the only transport where those two values are different. In all 3 methods they return `card`, so there is a chance this is a bug created by copy/paste between transports.

Since all 3 methods were added on the same [commit](https://github.com/a2aproject/a2a-python/commit/82a6b7cc9b83484a4ceabc2323e14e2ff0270f87), this feels like a likely mistake to make.

sokoliva:
Hi @ShishirRmc, thank you for pointing out this bug and fixing it! I gave approval to you PR. :)
Could you resolve Merge conflicts so we can merge it to main?

ShishirRmc:
@sokoliva I think we're all set now, everything should be up to date.

PR #556, opened 2025-11-20
[DRAFT] feat!: Upgrade A2A to v1.0

Tehsmash:
# Description

This updates the SDK to be A2A v1.0 compliant, all types are generated from the v1.0 a2a.proto. JSONRPC/HTTP+JSON transports are converted to use the a2a types encoded using ProtoJSON directly from the generated types.

Thank you for opening a Pull Request!
Before submitting your PR, there are a few things you can do to make sure it goes smoothly:

- [x] Follow the [`CONTRIBUTING` Guide](https://github.com/a2aproject/a2a-python/blob/main/CONTRIBUTING.md).
- [x] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
  - Important Prefixes for [release-please](https://github.com/googleapis/release-please):
    - `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
    - `feat:` represents a new feature, and correlates to a SemVer minor.
    - `feat!:`, or `fix!:`, `refactor!:`, etc., which represent a breaking change (indicated by the `!`) and will result in a SemVer major.
- [ ] Ensure the tests and linter pass (Run `bash scripts/format.sh` from the repository root to format)
- [ ] Appropriate docs were updated (if necessary)

Fixes #559

PR #572, opened 2025-11-28
refactor!:  upgrade SDK to A2A 1.0 specs

### LINKED ISSUE #559: [Feat]: Upgrade A2A to v1.0

A2A v1.0 is coming, we need to upgrade the SDK to support that specification

Refactor the SDK to use the new v1.0 version of the specification

N/A

### Additional context

_No response_

### Code of Conduct

muscariello:
## Summary

This PR migrates the a2a-python SDK from Pydantic-based types to protobuf-generated types, completing the upgrade to A2A v1.0. Fixes #559 

## Breaking Changes

- Replace Pydantic-based type system with protobuf types from `a2a_pb2`
- Update `Part` usage from `Part(root=TextPart(text=...))` to `Part(text=...)`
- Update `Role` enum from `Role.user`/`Role.agent` to `Role.ROLE_USER`/`Role.ROLE_AGENT`
- Update `TaskState` enum to use `TASK_STATE_*` prefix

## Changes

- Update all source files to use proto types directly from `a2a_pb2`
- Replace `model_dump()` with `MessageToDict()` for JSON serialization
- Replace `model_copy(deep=True)` with `CopyFrom()` for proto cloning
- Add new types module with proto imports and SDK-specific extras
- Add `proto_utils` module with identity conversion utilities
- Fix REST handler resource name formats for push notification configs
- Fix gRPC handler to use `SubscribeToTask` instead of `TaskSubscription`
- Fix database task store to handle proto objects from ORM
- Update all test files for proto patterns and assertions
- Fix spelling check failures by updating allow list
- Fix inclusive language check failures:
    - Replace `master` with `main` in documentation
    - Rename `Dummy` classes to `Mock` in tests

## Testing

- **601 tests passing**
- 23 tests skipped (expected - require DB connections or cryptography deps)

## Related

Builds on top of PR #556

Release-As: 1.0.0


buf.gen.yaml
```diff
@@ -2,7 +2,7 @@
 version: v2
 inputs:
   - git_repo: https://github.com/a2aproject/A2A.git
-    ref: main
+    ref: transports
```
Tehsmash: We need to change this back to "main" to pick up the latest fixes from the a2a.proto and before we merge we need to switch to the tag for v1.0 (when its cut)

src/a2a/client/auth/interceptor.py
```diff
@@ -51,7 +67,9 @@ async def intercept(
                     )
                     if not scheme_def_union:
                         continue
-                    scheme_def = scheme_def_union.root
+                    scheme_def = _get_security_scheme_value(scheme_def_union)
+                    if not scheme_def:
+                        continue
```
Tehsmash, line 47: Personal preference would be to use the functions available from proto to do the matching below instead of ending up with a non-statically typed function here. E.g. 

```python
match scheme.WhichOneof('scheme'):
    case 'http_auth_security_scheme' if scheme.http_auth_security_scheme.lower() == 'bearer':
    	...
```

tbh I think it would read better as a set of `if` blocks:

```python
scheme = agent_card.security_schemes.get(scheme_name)

if scheme.HasField('http_auth_security_scheme') and scheme.http_auth_security_scheme.lower() == 'bearer':
   ... 

if scheme.HasField('oauth2_security_scheme') or scheme.HasField('open_id_connect_security_scheme'):
   ...
```

src/a2a/types/extras.py
```diff
@@ -0,0 +1,360 @@
...
+
+from typing import Any, Literal, Union
+
+from google.protobuf.json_format import MessageToDict
+from google.protobuf.message import Message as ProtoMessage
+from pydantic import BaseModel, Field, RootModel, field_serializer
+
+# Alias for backward compatibility - the proto uses SendMessageRequest
+# where old code might use MessageSendParams
+from a2a.types.a2a_pb2 import SendMessageRequest as MessageSendParams
```
Tehsmash: Lets not do this, any code that is already using a2a.types.MessageSendParams would need change the imports anyway and I think this will be confusing.

src/a2a/types/extras.py
```diff
@@ -0,0 +1,360 @@
...
+from a2a.types.a2a_pb2 import SubscribeToTaskRequest as TaskResubscriptionRequest
+from a2a.types.a2a_pb2 import (
+    Message,
+    Task,
+    TaskPushNotificationConfig,
+)
+
+# Alias for streaming - same as SendMessageRequest in the proto
+SendStreamingMessageRequest = SendMessageRequest = MessageSendParams
```
Tehsmash: -1, lets just update the code to use SendMessageRequest.

src/a2a/types/extras.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+# Transport protocol constants for backward compatibility
+# These were an enum in the old Pydantic types, now they're just strings
+class TransportProtocol:
+    """Transport protocol string constants for backward compatibility."""
+    jsonrpc = "JSONRPC"
+    http_json = "HTTP+JSON"
+    grpc = "GRPC"
```
Tehsmash: In the client code I add these as constants, if we want to put them somewhere common I would add a "constants.py" / constants module outside of "types".

src/a2a/types/extras.py
```diff
@@ -0,0 +1,360 @@
...
+    jsonrpc: Literal["2.0"] = "2.0"
+    id: str | int | None = None
+    result: None = None
+
+
+class GetAuthenticatedExtendedCardSuccessResponse(A2ABaseModel):
+    """Success response for GetAuthenticatedExtendedCard RPC."""
+    jsonrpc: Literal["2.0"] = "2.0"
+    id: str | int | None = None
+    result: Any  # AgentCard
```
Tehsmash: Let not add our own implementation of this. In the client code I pull in the "jsonrpc" package from pypi, which has the JSONRPC20Response type already defined, then we only need to concern ourselves with populating the result or error fields.

src/a2a/server/apps/jsonrpc/fastapi_app.py
```diff
@@ -45,15 +46,10 @@ def openapi(self) -> dict[str, Any]:
...
-            defs = a2a_request_schema.pop('$defs', {})
-            component_schemas = openapi_schema.setdefault(
-                'components', {}
-            ).setdefault('schemas', {})
-            component_schemas.update(defs)
-            component_schemas['A2ARequest'] = a2a_request_schema
+            # A2ARequest is now a Union type of proto messages, so we can't use
+            # model_json_schema. Instead, we just mark it as added without
+            # adding the schema since proto types don't have Pydantic schemas.
+            # The OpenAPI schema will still be functional for the endpoints.
```
Tehsmash, line 72: LOL ... "the code doesn't support this any more so we'll just mark this test as passed anyway" 

In all seriousness this should be replaced by the generated a2a.json schema I think.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -154,22 +157,18 @@ class JSONRPCApplication(ABC):
...
+        'message/stream': SendStreamingMessageRequest,
+        'tasks/get': GetTaskRequest,
+        'tasks/cancel': CancelTaskRequest,
+        'tasks/pushNotificationConfig/set': SetTaskPushNotificationConfigRequest,
+        'tasks/pushNotificationConfig/get': GetTaskPushNotificationConfigRequest,
+        'tasks/pushNotificationConfig/list': ListTaskPushNotificationConfigRequest,
+        'tasks/pushNotificationConfig/delete': DeleteTaskPushNotificationConfigRequest,
+        'tasks/resubscribe': TaskResubscriptionRequest,
+        'agent/authenticatedExtendedCard': GetExtendedAgentCardRequest,
     }
```
Tehsmash, line 173: The method names have changed to align with the gRPC ones, i.e message/send is now "SendMessage". We need to fix that in the client and here.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -35,12 +41,9 @@
     JSONRPCErrorResponse,
     JSONRPCRequest,
     JSONRPCResponse,
```
Tehsmash: Lets pull in the jsonrpc dependency so that we can remove these from the types.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
Tehsmash, line 504: I think we should revisit this code and match on the JSON RPC method first then perform the correct decoding as there are body types like "SendMessageRequest" which are reused for different methods.

For example:

```python
def _process_request(self, request: JSONRPC20Request):
	match request.method:
		case "SendMessage":
			input: SendMessageRequest = ParseDict(request.params, SendMessageRequest())
			result = await self.handler.on_send_message(input)
		case "CancelTask":
			input: CancelTaskRequest = ParseDict(request.params, CancelTaskRequest())
			...
...
```

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -207,11 +231,14 @@ async def _setup_message_execution(
             A tuple of (task_manager, task_id, queue, result_aggregator, producer_task)
         """
         # Create task manager and validate existing task
+        # Proto empty strings should be treated as None
+        task_id = params.request.task_id or None
```
Tehsmash: Need to confirm if "task_id" should be just the UUID, or should be in the form "tasks/<uuid>".
I think AIP for this said that it should be "tasks/<uuid>" because this is the FQN for the task.

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -452,20 +479,25 @@ async def on_set_task_push_notification_config(
...
+        # Build the response config with the proper name
+        return TaskPushNotificationConfig(
+            name=f'{params.parent}/pushNotificationConfigs/{params.config_id}',
+            push_notification_config=params.config.push_notification_config,
+        )
 
     async def on_get_task_push_notification_config(
         self,
-        params: TaskIdParams | GetTaskPushNotificationConfigParams,
+        params: CancelTaskRequest | GetTaskPushNotificationConfigRequest,
```
Tehsmash:
```suggestion
        params: GetTaskPushNotificationConfigRequest,
```

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -490,28 +523,29 @@ async def on_get_task_push_notification_config(
             )
 
         return TaskPushNotificationConfig(
-            task_id=params.id,
+            name=params.name,
             push_notification_config=push_notification_config[0],
         )
```
Tehsmash, line 536: This function needs re-visiting, you should be able to have more than one pushNotificationConfig per task.

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -490,28 +523,29 @@ async def on_get_task_push_notification_config(
...
 
     async def on_resubscribe_to_task(
```
Tehsmash: this function needs renaming `on_subscribe_to_task`

src/a2a/client/auth/interceptor.py
muscariello: Not sure how's HasField improves things here.

src/a2a/types/extras.py
muscariello: this is a quick removal that in any case is supposed to be removed before the PR is complete.

src/a2a/types/extras.py
muscariello: See above.

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -545,24 +579,26 @@ async def on_list_task_push_notification_config(
...
 
         push_notification_config_list = await self._push_config_store.get_info(
-            params.id
+            task_id
         )
 
         return [
             TaskPushNotificationConfig(
-                task_id=params.id, push_notification_config=config
+                name=f'tasks/{task_id}/pushNotificationConfigs/{config.id}',
```
Tehsmash:
```suggestion
                name=f'{params.parent}/pushNotificationConfigs/{config.id}',
```

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -535,7 +569,7 @@ async def on_resubscribe_to_task(
 
     async def on_list_task_push_notification_config(
         self,
-        params: ListTaskPushNotificationConfigParams,
+        params: ListTaskPushNotificationConfigRequest,
         context: ServerCallContext | None = None,
     ) -> list[TaskPushNotificationConfig]:
```
Tehsmash: This should return a `ListTaskPushNotificationConfigResponse`

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -490,28 +523,29 @@ async def on_get_task_push_notification_config(
...
+            name=params.name,
             push_notification_config=push_notification_config[0],
         )
 
     async def on_resubscribe_to_task(
         self,
-        params: TaskIdParams,
+        params: SubscribeToTaskRequest,
         context: ServerCallContext | None = None,
     ) -> AsyncGenerator[Event]:
```
Tehsmash:
```suggestion
    ) -> AsyncGenerator[StreamResponse]:
```

src/a2a/types/extras.py
muscariello: To move this elsewhere I would have to solve a circular dependency first. There needs to be additional refactoring work before I can do that.

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -134,7 +135,12 @@ async def SendMessage(
                 a2a_request, server_context
             )
             self._set_extension_metadata(context, server_context)
-            return proto_utils.ToProto.task_or_message(task_or_message)
+            result = proto_utils.ToProto.task_or_message(task_or_message)
+            # Wrap in SendMessageResponse based on type
+            if isinstance(result, a2a_pb2.Task):
+                return a2a_pb2.SendMessageResponse(task=result)
+            else:
+                return a2a_pb2.SendMessageResponse(msg=result)
```
Tehsmash: `on_message_send` should return a `SendMessageResponse` otherwise we have to handle this in every handler.

src/a2a/server/request_handlers/jsonrpc_handler.py
```diff
@@ -94,29 +107,35 @@ async def on_message_send(
...
         try:
             task_or_message = await self.request_handler.on_message_send(
-                request.params, context
+                request, context
             )
+            # Wrap the result in SendMessageResponseProto for consistent client parsing
+            if isinstance(task_or_message, Task):
+                response_proto = SendMessageResponseProto(task=task_or_message)
+            else:
+                response_proto = SendMessageResponseProto(msg=task_or_message)
```
Tehsmash: update on_message_send to return a SendMessageResponse

src/a2a/server/request_handlers/jsonrpc_handler.py
```diff
@@ -144,24 +163,21 @@ async def on_message_send_stream(
         """
         try:
             async for event in self.request_handler.on_message_send_stream(
-                request.params, context
+                request, context
             ):
+                # Wrap the event in StreamResponse for consistent client parsing
+                stream_response = proto_utils.ToProto.stream_response(event)
```
Tehsmash: The generator here should be updated to return StreamResponse objects so we don't need to handle this here.

src/a2a/server/request_handlers/jsonrpc_handler.py
```diff
@@ -222,24 +238,21 @@ async def on_resubscribe_to_task(
         """
         try:
             async for event in self.request_handler.on_resubscribe_to_task(
-                request.params, context
+                request, context
             ):
+                # Wrap the event in StreamResponse for consistent client parsing
+                stream_response = proto_utils.ToProto.stream_response(event)
```
Tehsmash: ditto, "on_subscribe_to_task" should return an AsyncGenerator(StreamResponse)

src/a2a/server/request_handlers/request_handler.py
```diff
@@ -125,7 +126,7 @@ async def on_set_task_push_notification_config(
     @abstractmethod
     async def on_get_task_push_notification_config(
         self,
-        params: TaskIdParams | GetTaskPushNotificationConfigParams,
+        params: CancelTaskRequest | GetTaskPushNotificationConfigRequest,
```
Tehsmash:
```suggestion
        params: GetTaskPushNotificationConfigRequest,
```

src/a2a/server/request_handlers/response_helpers.py
```diff
@@ -124,19 +134,21 @@ def prepare_response_object(
...
-    response = A2AError(
-        root=InvalidAgentResponseError(
-            message='Agent returned invalid type response for this method'
-        )
+    error = InvalidAgentResponseError(
+        message='Agent returned invalid type response for this method'
     )
 
-    return build_error_response(request_id, response, response_type)
+    return build_error_response(request_id, error, response_type)
```
Tehsmash: We should update these methods to use the types from the jsonrpc library, for example

```
if isinstance(body, success_response_types):
	response = JSONRPC20Response()
	response.params = MessageToDict(response)
	return response
```

src/a2a/server/request_handlers/rest_handler.py
```diff
@@ -137,7 +141,7 @@ async def on_cancel_task(
         """
         task_id = request.path_params['id']
         task = await self.request_handler.on_cancel_task(
-            TaskIdParams(id=task_id), context
+            CancelTaskRequest(name=f'tasks/{task_id}'), context
         )
         if task:
             return MessageToDict(proto_utils.ToProto.task(task))
```
Tehsmash: Not sure why we need `proto_utils.ToProto.task(task)` here

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -57,7 +58,7 @@ async def _dispatch_notification(
                 headers = {'X-A2A-Notification-Token': push_info.token}
             response = await self._client.post(
                 url,
-                json=task.model_dump(mode='json', exclude_none=True),
+                json=MessageToDict(task),
```
Tehsmash: push notifications should send "StreamResponse" payloads according to the v1.0 spec

src/a2a/server/request_handlers/request_handler.py
```diff
@@ -143,7 +144,7 @@ async def on_get_task_push_notification_config(
     @abstractmethod
     async def on_resubscribe_to_task(
```
Tehsmash: on_subscribe_to_task

src/a2a/server/request_handlers/request_handler.py
```diff
@@ -143,7 +144,7 @@ async def on_get_task_push_notification_config(
...
         self,
-        params: TaskIdParams,
+        params: CancelTaskRequest,
```
Tehsmash:
```suggestion
        params: SubscribeToTaskRequest,
```

src/a2a/server/models.py
```diff
@@ -57,26 +59,30 @@ def __init__(self, pydantic_type: type[T], **kwargs: dict[str, Any]):
...
-        """Convert a JSON-like dictionary from the database back to a Pydantic model."""
+        """Convert a JSON-like dictionary from the database back to a Pydantic model or Protobuf message."""
         if value is None:
             return None
-        return self.pydantic_type.model_validate(value)
+        # Check if it's a protobuf message class
+        if isinstance(self.pydantic_type, type) and issubclass(self.pydantic_type, ProtoMessage):
+            return ParseDict(value, self.pydantic_type())  # type: ignore[return-value]
+        # Assume it's a Pydantic model
+        return self.pydantic_type.model_validate(value)  # type: ignore[union-attr]
```
Tehsmash: No need to support Pydantic here IMO, just need to support the protobuf types. 

One thought: We need to understand if we need to support old messages/tasks stored in the DB and translate them to the protobuf types. Otherwise we may require people to drop the DBs for their Agents

src/a2a/utils/error_handlers.py
Tehsmash, line 38: We should move all these errors to errors.py instead of extras.

src/a2a/utils/parts.py
```diff
@@ -33,16 +30,17 @@ def get_data_parts(parts: list[Part]) -> list[dict[str, Any]]:
     Returns:
         A list of dictionaries containing the data from any `DataPart` objects found.
     """
-    return [part.root.data for part in parts if isinstance(part.root, DataPart)]
+    from google.protobuf.json_format import MessageToDict
```
Tehsmash: No need for an inline import, lets move this to the top of the file.

src/a2a/utils/proto_utils.py
Tehsmash, line 1: We can remove this file and its classes/functions, as we're using the proto types everywhere now.

src/a2a/utils/task.py
```diff
@@ -25,11 +31,11 @@ def new_task(request: Message) -> Task:
     if not request.parts:
         raise ValueError('Message parts cannot be empty')
     for part in request.parts:
-        if isinstance(part.root, TextPart) and not part.root.text:
-            raise ValueError('TextPart content cannot be empty')
+        if part.text is not None and not part.text:
```
Tehsmash:
```suggestion
        if part.HasField('text') and not part.text:
```

src/a2a/utils/task.py
```diff
@@ -85,8 +91,12 @@ def apply_history_length(task: Task, history_length: int | None) -> Task:
...
-        limited_history = task.history[-history_length:]
+        limited_history = list(task.history[-history_length:])
         # Create a new task instance with limited history
-        return task.model_copy(update={'history': limited_history})
-
+        task_copy = Task()
+        task_copy.CopyFrom(task)
+        # Clear and re-add history items
+        del task_copy.history[:]
+        task_copy.history.extend(limited_history)
```
Tehsmash, line 100:
```suggestion
        task_copy.history[:] = limited_history
```

src/a2a/client/auth/interceptor.py
Tehsmash: As explained our call, its mostly about readability, but also prevent us using reflection to check the python types, and instead we can just rely on the boolean check of "does the request have this field set" then it can use it.

src/a2a/client/auth/interceptor.py
```diff
@@ -7,26 +7,36 @@
...
+    | HTTPAuthSecurityScheme
+    | OAuth2SecurityScheme
+    | OpenIdConnectSecurityScheme
+    | MutualTlsSecurityScheme
+    | None
+)
+
 
-def _get_security_scheme_value(scheme: SecurityScheme):
+def _get_security_scheme_value(scheme: SecurityScheme) -> _SecuritySchemeValue:
```
Tehsmash: I think this change has just proven my point about moving this inline in the function below and changing how the "match" is done.

src/a2a/server/apps/jsonrpc/fastapi_app.py
muscariello: I am not using a2a.json at all. So need to rethink about this test.

src/a2a/server/tasks/task_manager.py
```diff
@@ -143,7 +143,7 @@ async def save_task_event(
             if task.status.HasField('message'):
                 task.history.append(task.status.message)
             if event.metadata:
-                task.metadata.update(event.metadata)
+                task.metadata.update(dict(event.metadata))  # type: ignore[arg-type]
```
Tehsmash: We should use task.metadata.MergeFrom/CopyFrom here depending on the desired behaviour

src/a2a/client/auth/interceptor.py
muscariello: Check now @Tehsmash

src/a2a/server/request_handlers/default_request_handler.py
muscariello: next commit

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
muscariello: I think I have fixed this everywhere now. I will keep checking...

src/a2a/server/request_handlers/default_request_handler.py
muscariello: I am not sure you're right. Let's discuss tomorrow...

src/a2a/types/extras.py
muscariello: I am moving things from the refactor out of extras.py until it is empty. Then I will remove the file. A few more commits I think.

src/a2a/client/auth/interceptor.py
```diff
@@ -35,63 +28,64 @@ async def intercept(
         """Applies authentication headers to the request if credentials are available."""
         if (
             agent_card is None
-            or agent_card.security is None
-            or agent_card.security_schemes is None
+            or not agent_card.security
+            or not agent_card.security_schemes
```
Tehsmash, line 35: Double checking the protobuf docs https://protobuf.dev/reference/python/python-generated/#embedded_message it sounds like we may need to use `agent_card.HasField("...")` this separates the difference between agent_card.security set, not set and set but empty from what I can tell.

src/a2a/client/auth/interceptor.py
Tehsmash: I don't seem to be able to resolve my own threads on the a2a repos, but I'll leave a LGTM if I think it should be resolved.

src/a2a/utils/errors.py
```diff
@@ -1,22 +1,172 @@
...
+    message: str = 'Invalid agent response'
+    data: Any | None = None
+
+
+class AuthenticatedExtendedCardNotConfiguredError(A2ABaseModel):
+    """A2A-specific error for authenticated extended card not configured (-32007)."""
+
+    code: Literal[-32007] = -32007
+    message: str = 'Authenticated Extended Card is not configured'
+    data: Any | None = None
```
Tehsmash: The A2A exception's here shouldn't have JSONRPC specific data, the JSONRPC server/client components should have a map from the exceptions to the JSONRPC code and vice-versa for raising an exception on receipt of a JSONRPCError 

JSONRPC specific exceptions like MethodNotFound should live in the JSONRPC implementation code or a separate JSONRPC common package IMO.

src/a2a/utils/error_handlers.py
Tehsmash: LGTM

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
Tehsmash: LGTM

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
```diff
@@ -313,113 +306,117 @@ async def _handle_requests(self, request: Request) -> Response:  # noqa: PLR0911
             if not self._allowed_content_length(request):
                 return self._generate_error_response(
                     request_id,
```
Tehsmash, line 313: I think at least the "request_id" check stuff can move below the JSONRPC20Request.from_data. E.g.

```
base_request = JSONRPC20Request.from_data(body)
...

if base_request.id is not None and not isinstance(
                    base_request.id, str | int
                ):
	base_request.id=None
```

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
Tehsmash: I still think this would be cleaner than doing METHOD_TO_MODEL first and matching on the classes after the fact. And makes it more obvious IMO that the final / default case of the "match" would be "MethodNotFound".

src/a2a/server/request_handlers/default_request_handler.py
Tehsmash: "StreamResponse" is the type which wraps the fact that the stream can include different types of message, and allows the consumer to differentiate safely. Regardless of the transport this body should be used so I don't know why we wouldn't do it here.

src/a2a/server/request_handlers/default_request_handler.py
Tehsmash: After checking the _push_config_store, I think we need to update this like

```
self._push_config_store.get_info
```

returns a list, so we need to do something like:

```
config_id = _extract_config_id(params.name)
for config in push_notification_configs:
	if config.id == config_id:
		return TaskPushNotificationConfig...
raise ServerError(....)
```

src/a2a/server/request_handlers/default_request_handler.py
Tehsmash: LGTM

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -452,20 +484,25 @@ async def on_set_task_push_notification_config(
...
+            task_id,
+            params.config.push_notification_config,
         )
 
-        return params
+        # Build the response config with the proper name
+        return TaskPushNotificationConfig(
+            name=f'{params.parent}/pushNotificationConfigs/{params.config_id}',
+            push_notification_config=params.config.push_notification_config,
+        )
```
Tehsmash, line 503: Do you think there is a reason not to change the "PushNotificationConfigStore" to take "TaskPushNotificationConfig" instead of a "PushNotificationConfig"?

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -288,7 +320,7 @@ async def _send_push_notification_if_needed(
 
     async def on_message_send(
         self,
-        params: MessageSendParams,
+        params: SendMessageRequest,
         context: ServerCallContext | None = None,
     ) -> Message | Task:
```
Tehsmash, line 327: SendMessageResponse
```suggestion
    ) -> SendMessageResponse
```

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -357,7 +389,7 @@ async def push_notification_callback() -> None:
 
     async def on_message_send_stream(
         self,
-        params: MessageSendParams,
+        params: SendMessageRequest,
         context: ServerCallContext | None = None,
     ) -> AsyncGenerator[Event]:
```
Tehsmash, line 396:
```suggestion
    ) -> AsyncGenerator[StreamResponse]:
```

src/a2a/types/extras.py
Tehsmash: LGTM

src/a2a/types/extras.py
Tehsmash: LGTM

src/a2a/types/extras.py
Tehsmash: LGTM

src/a2a/types/extras.py
Tehsmash: LGTM

src/a2a/server/apps/jsonrpc/fastapi_app.py
Tehsmash: What I meant is that if fastapi wants to serve an OpenAPI schema, then we need to use the a2a.json instead of generating one from the python types.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
Tehsmash: LGTM

src/a2a/server/request_handlers/default_request_handler.py
Tehsmash: LGTM

src/a2a/server/request_handlers/request_handler.py
Tehsmash: LGTM

src/a2a/server/request_handlers/response_helpers.py
Tehsmash: LGTM, although I would like to see this handled a bit better overall, e.g. I'm not sure we need a helper function for creating the good responses when its just one line essentially:

```
return JSONRPC20Response(result=MessageToDict(response, preserving_proto_field_name=False), _id=request_id).data
 ```
 
 Then we just need to have a helper for errors like:
 
 ```
 func build_jsonrpc_error_response(e: Exception):
	data = None
	code = <code for unknown error>
	
 	if isinstance(TaskNotFound):
 		code = 12313
 		data = e.data / {"taskID": e.taskID}
 		
 	return JSONRPC20Response(error={
		"code": code,
		"msg": str(e),
		"data": data,
 	})
 ```

File: src/a2a/server/request_handlers/rest_handler.py (hunk 23 above)
Comment (Tehsmash) [reply src/a2a/server/request_handlers/rest_handler.py]: LGTM

File: src/a2a/server/request_handlers/request_handler.py (hunk 25 above)
Comment (Tehsmash) [reply src/a2a/server/request_handlers/request_handler.py]: LGTM

File: src/a2a/server/request_handlers/request_handler.py (hunk 26 above)
Comment (Tehsmash) [reply src/a2a/server/request_handlers/request_handler.py]: LGTM

File: src/a2a/server/models.py (hunk 27 above)
Comment (Tehsmash) [reply src/a2a/server/models.py]: I think this is still relevant to models.py even though its been disconnected. You can resolve this and I'll open a new comment.

File: src/a2a/server/models.py
Comment (Tehsmash) [inline src/a2a/server/models.py:1]: No need to support Pydantic here IMO, just need to support the protobuf types.

One thought: We need to understand if we need to support old messages/tasks stored in the DB and translate them to the protobuf types. Otherwise we may require people to drop the DBs for their Agents

File: src/a2a/utils/parts.py (hunk 28 above)
Comment (Tehsmash) [reply src/a2a/utils/parts.py]: LGTM

File: src/a2a/utils/proto_utils.py
Comment (Tehsmash) [reply src/a2a/utils/proto_utils.py:1]: Most of the stuff from here is gone, i think we can nuke the helper function here if we fix the response from the default_request_handler.

**Comment by muscariello** [comment]
@Tehsmash I am going to wait a few more days for the A2A 1.0 specs to get the most recent updates.

File: src/a2a/client/auth/interceptor.py [hunk 39]
```diff
@@ -3,18 +3,34 @@
...
+        return scheme.api_key_security_scheme
+    elif which == 'http_auth_security_scheme':
+        return scheme.http_auth_security_scheme
+    elif which == 'oauth2_security_scheme':
+        return scheme.oauth2_security_scheme
+    elif which == 'open_id_connect_security_scheme':
+        return scheme.open_id_connect_security_scheme
+    elif which == 'mtls_security_scheme':
+        return scheme.mtls_security_scheme
+    return None
```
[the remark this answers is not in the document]
lkawka: +1

src/a2a/utils/task.py
lkawka: +1

src/a2a/utils/constants.py
```diff
@@ -4,3 +4,18 @@
...
+
+
+# Transport protocol constants
+# These match the protocol binding values used in AgentCard
+TRANSPORT_JSONRPC = 'JSONRPC'
+TRANSPORT_HTTP_JSON = 'HTTP+JSON'
+TRANSPORT_GRPC = 'GRPC'
+
+
+class TransportProtocol:
```
lkawka, line 16: This looks like a good candidate for StrEnum

src/a2a/server/request_handlers/default_request_handler.py
lkawka: +1

lkawka (commented):
Hi Luca, thanks for the PR; apologies for the delay.

The team agrees with moving to proto-based types. Please address Sam's open comments, but otherwise, it looks good.

Proposed next steps:
1. Resolve issues and submit to the [a2aproject:1.0-a2a_proto_refactor](https://github.com/a2aproject/a2a-python/tree/1.0-a2a_proto_refactor) branch.
2. Merge into 1.0-dev via a separate PR to avoid conflicts with existing implementations.

muscariello:
@lkawka I will go through all the remaining comments on Monday. Thanks for the review.

muscariello:
@Tehsmash  could you please check again? I am going to recheck all the previous comments on Monday but please recheck as well. I included the latest updates based on the merge queue in the spec based on the latest TSC votes. We'll see on Tue if the spec can be considered complete for v1. 
We could also merge this one and propose another PR after to  avoid conflicts.

scripts/grpc_gen_post_processor.py
Tehsmash, line 1: This should be removed as its a left over.

src/a2a/client/auth/interceptor.py
Tehsmash: @muscariello  Did you have a chance to double check this?

src/a2a/client/transports/jsonrpc.py
```diff
@@ -69,19 +57,22 @@ def __init__(
         if url:
             self.url = url
         elif agent_card:
-            self.url = agent_card.url
+            if agent_card.supported_interfaces:
+                self.url = agent_card.supported_interfaces[0].url
```
Tehsmash, line 61: I don't think we can assume that supported_interfaces[0] is JSONRPC, we need logic like we have in "rest.py" which checks the interface type and pulls out the first JSONRPC type.

It would probably make sense to have a common function (if there isn't one already) which finds the first supported_interface of a given type and returns it.

src/a2a/client/base_client.py
```diff
@@ -77,76 +77,79 @@ async def send_message(
...
-        else:
-            config = base_config
 
-        params = MessageSendParams(
+        if configuration:
+            config.MergeFrom(configuration)
+            # Proto3 doesn't support HasField for scalars, so MergeFrom won't
+            # override with default values (e.g. False). We explicitly set it here
+            # assuming configuration is authoritative.
+            config.blocking = configuration.blocking
```
Tehsmash, line 83: This seems weird to me. I think it might really be a bug in the a2a.proto. 

https://protobuf.dev/programming-guides/field_presence/#presence-proto3

> This default behavior of not tracking presence without the optional label is different from the proto2 behavior. We recommend using the optional label with proto3 unless you have a specific reason not to.

This enables explict presence, in the next versions of Proto, "editions", they go back to having explict presense with out the "optional" flag.

With the "optional" flag set in the proto, "MergeFrom" works as expected:

```
$ uv run src/protobooleanmerge/test.py
Thing1 HasField(blocking):  True  Value:  False
Thing2 HasField(blocking):  True  Value:  True
After MergeFrom:
Thing2 HasField(blocking):  True  Value:  False
```

src/a2a/client/client_factory.py
Tehsmash, line 233: I think we need to adjust this logic, at the moment the "server_set" will build a map keeping the last (therefore least preferential) supported_interface for each protocol_binding type the server specifies.

In the spec you can have multiple interface of the same protocol_binding type (i.e. 2 JSONRPC interfaces) so that you can have say a preferred higher bandwidth option and a lower bandwidth option as fail over.

For simplicity we should probably just take the first item in the card's supported interface which is in the supported_protocol_bindings list something like:

```
for supported_interface in card.supported_interfaces:
	if supported_interface.protocol_binding in self._config.supported_protocol_bindings:
		transport_protocol = x
        transport_url = supported_interface.url
		break
```

For the client preference mode we can do:

```
for protocol_binding in self._config.supported_protocol_bindings:
	supported_interface = next([si for si in card.supported_interfaces if si.protocol_binding == protocol_binding], None)
	if supported_interface != None:
		transport_protocol = x
        transport_url = supported_interface.url
		break
```

src/a2a/server/agent_execution/simple_request_context_builder.py
```diff
@@ -13,6 +14,8 @@ def __init__(
         self,
         should_populate_referred_tasks: bool = False,
         task_store: TaskStore | None = None,
+        task_id_generator: IDGenerator | None = None,
+        context_id_generator: IDGenerator | None = None,
```
Tehsmash, line 18: Looks like this is getting included in this PR because of a merge to main :/ perhaps we need to make sure that the base branch is up to date?

src/a2a/server/apps/jsonrpc/fastapi_app.py
Tehsmash: I think we still need to solve this right?

Maybe we need to add ProtoJSON generator openAPI generate to the buf.gen.yaml then it can generate a static a2a.json that can be loaded and served from here?

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
Tehsmash: We can do a follow up to clean this up post v1

src/a2a/server/apps/rest/rest_adapter.py
```diff
@@ -152,7 +158,7 @@ async def handle_get_agent_card(
         if self.card_modifier:
             card_to_serve = self.card_modifier(card_to_serve)
 
-        return card_to_serve.model_dump(mode='json', exclude_none=True)
+        return MessageToDict(card_to_serve, preserving_proto_field_name=True)
```
Tehsmash: why "preserving_proto_field_name=True"? I thought all json encoding should be using the standard ProtoJSON encoding to lowerCamel case.

src/a2a/server/request_handlers/default_request_handler.py
Tehsmash: This function still needs an update

src/a2a/server/request_handlers/default_request_handler.py
Tehsmash: Still relevant

src/a2a/server/request_handlers/jsonrpc_handler.py
```diff
@@ -86,38 +86,54 @@ def __init__(
...
+                    'task': MessageToDict(
+                        task_or_message, preserving_proto_field_name=False
+                    )
+                }
+            elif isinstance(task_or_message, Message):
+                result = {
+                    'message': MessageToDict(
+                        task_or_message, preserving_proto_field_name=False
+                    )
+                }
```
Tehsmash: i think this should be building a SendMessageResponse with task or message set, then using MessageToDict (in fact this is how its being handled in the other transports already).

Personally I think the default_request_handler "on_send_message" should return a SendMessageResponse then we don't need to duplicate that code in all the other transport handlers..

src/a2a/server/tasks/base_push_notification_sender.py
Tehsmash: still relevant

src/a2a/client/client_factory.py
```diff
@@ -14,14 +14,18 @@
...
     AgentCapabilities,
     AgentCard,
     AgentInterface,
-    TransportProtocol,
 )
 
 
+TRANSPORT_PROTOCOLS_JSONRPC = 'JSONRPC'
+TRANSPORT_PROTOCOLS_GRPC = 'GRPC'
+TRANSPORT_PROTOCOLS_HTTP_JSON = 'HTTP+JSON'
```
Tehsmash: We can use the constants that are now defined in utils/constants.py

src/a2a/utils/errors.py
Tehsmash: still relevant

src/a2a/utils/parts.py
Tehsmash, line 1: Be aware this will change if Darrel's "Part" proposal merges.

src/a2a/utils/task.py
Tehsmash: still relevant

src/a2a/server/tasks/task_manager.py
muscariello: I will avoid casting

src/a2a/server/tasks/base_push_notification_sender.py
muscariello: Will do this.

src/a2a/server/request_handlers/default_request_handler.py
muscariello: I will do this.

src/a2a/server/request_handlers/default_request_handler.py
muscariello: I will do this.

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
muscariello: Post v1 ok for me.

src/a2a/client/auth/interceptor.py
muscariello: The HasField method is not available for repeated fields (security) or map fields (security_schemes) in Protocol Buffers v3 (Python).

BTW I tested.

For these types, checking truthiness (e.g., if not agent_card.security) is the correct way to check if they are empty/unset. I've added a comment to the code to clarify this.

src/a2a/server/apps/jsonrpc/fastapi_app.py
muscariello: Done. a2a.json generated by a bash script called from the project toml. 
This is now loaded here as a2a.json.

muscariello:
@lkawka please go ahead and merge. We will still have minor updates coming from the A2A spec v1. But I would send other smaller PRs later.

muscariello:
It looks like I can merge (?!) but I will wait a LGTM from you @lkawka...

PR #581, opened 2025-12-05
feat: Implement Agent Card Signing and Verification per Spec

sokoliva:
This PR introduces digital signatures for Agent Cards to ensure authenticity and integrity, adhering to the A2A specification for [Agent Card Signing (Section 8.4).](https://a2a-protocol.org/latest/specification/#84-agent-card-signing)

## Changes:
- Implement `Canonicalization` Logic (`src/a2a/utils/signing.py`)
- Add `Signing` and `Verification` Utilities (`src/a2a/utils/signing.py`):
   - `create_agent_card_signer` which generates an `agent_card_signer` for signing `AgentCards`
   - `create_signature_verifier` which generates a `signature_verifier` for verification of `AgentCard` signatures
- Enable signature verification support for `json-rpc`, `rest` and `gRPC` transports
- Add Protobuf Conversion for Signatures (`src/a2a/utils/proto_utils.py`) ensuring `AgentCardSignature` can be serialized and deserialized for gRPC transport
- Add related tests:
   - integration tests for fetching signed cards from the Server
   - unit tests for signing util
   - unit tests for protobuf conversions 

release-please:force-run

BEGIN_COMMIT_OVERRIDE
feat: Implement Agent Card Signing and Verification per Spec

Release-As: 0.3.22
END_COMMIT_OVERRIDE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,7 +101,7 @@
 ### Bug Fixes
 
 * apply `history_length` for `message/send` requests ([#498](https://github.com/a2aproject/a2a-python/issues/498)) ([a49f94e](https://github.com/a2aproject/a2a-python/commit/a49f94ef23d81b8375e409b1c1e51afaf1da1956))
-* **client:** `A2ACardResolver.get_agent_card` will auto-populate with `agent_card_path` when `relative_card_path` is empty ([#508](https://github.com/a2aproject/a2a-python/issues/508)) ([ba24ead](https://github.com/a2aproject/a2a-python/commit/ba24eadb5b6fcd056a008e4cbcef03b3f72a37c3))
+* **client:** `A2ACardResolver.get_agent_card` will autopopulate with `agent_card_path` when `relative_card_path` is empty ([#508](https://github.com/a2aproject/a2a-python/issues/508)) ([ba24ead](https://github.com/a2aproject/a2a-python/commit/ba24eadb5b6fcd056a008e4cbcef03b3f72a37c3))
 
 
 ### Documentation
@@ -438,8 +438,8 @@
 * Event consumer should stop on input_required ([#167](https://github.com/a2aproject/a2a-python/issues/167)) ([51c2d8a](https://github.com/a2aproject/a2a-python/commit/51c2d8addf9e89a86a6834e16deb9f4ac0e05cc3))
 * Fix Release Version ([#161](https://github.com/a2aproject/a2a-python/issues/161)) ([011d632](https://github.com/a2aproject/a2a-python/commit/011d632b27b201193813ce24cf25e28d1335d18e))
 * generate StrEnum types for enums ([#134](https://github.com/a2aproject/a2a-python/issues/134)) ([0c49dab](https://github.com/a2aproject/a2a-python/commit/0c49dabcdb9d62de49fda53d7ce5c691b8c1591c))
-* library should released as 0.2.6 ([d8187e8](https://github.com/a2aproject/a2a-python/commit/d8187e812d6ac01caedf61d4edaca522e583d7da))
-* remove error types from enqueable events ([#138](https://github.com/a2aproject/a2a-python/issues/138)) ([511992f](https://github.com/a2aproject/a2a-python/commit/511992fe585bd15e956921daeab4046dc4a50a0a))
+* library should be released as 0.2.6 ([d8187e8](https://github.com/a2aproject/a2a-python/commit/d8187e812d6ac01caedf61d4edaca522e583d7da))
+* remove error types from enqueueable events ([#138](https://github.com/a2aproject/a2a-python/issues/138)) ([511992f](https://github.com/a2aproject/a2a-python/commit/511992fe585bd15e956921daeab4046dc4a50a0a))
 * **stream:** don't block event loop in EventQueue ([#151](https://github.com/a2aproject/a2a-python/issues/151)) ([efd9080](https://github.com/a2aproject/a2a-python/commit/efd9080b917c51d6e945572fd123b07f20974a64))
 * **task_updater:** fix potential duplicate artifact_id from default v… ([#156](https://github.com/a2aproject/a2a-python/issues/156)) ([1f0a769](https://github.com/a2aproject/a2a-python/commit/1f0a769c1027797b2f252e4c894352f9f78257ca))
 
--- a/Gemini.md
+++ b/Gemini.md
@@ -4,7 +4,7 @@
 - uv as package manager
 
 ## How to run all tests
-1. If dependencies are not installed install them using following command
+1. If dependencies are not installed, install them using the following command
    ```
    uv sync --all-extras 
    ```
--- a/src/a2a/utils/error_handlers.py
+++ b/src/a2a/utils/error_handlers.py
@@ -117,12 +117,12 @@ def rest_stream_error_handler(
                 ', Data=' + str(error.data) if error.data else '',
             )
             # Since the stream has started, we can't return a JSONResponse.
-            # Instead, we runt the error handling logic (provides logging)
+            # Instead, we run the error handling logic (provides logging)
             # and reraise the error and let server framework manage
             raise e
         except Exception as e:
             # Since the stream has started, we can't return a JSONResponse.
-            # Instead, we runt the error handling logic (provides logging)
+            # Instead, we run the error handling logic (provides logging)
             # and reraise the error and let server framework manage
             raise e
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -5,7 +5,7 @@
     uv run pytest -v -s client/test_client_factory.py
     ```
 
-In case of failures, you can cleanup the cache:
+In case of failures, you can clean up the cache:
 
 1. `uv clean`
 2. `rm -fR .pytest_cache .venv __pycache__`
--- a/tests/e2e/push_notifications/test_default_push_notification_support.py
+++ b/tests/e2e/push_notifications/test_default_push_notification_support.py
@@ -35,7 +35,7 @@ from a2a.types import (
 @pytest.fixture(scope='module')
 def notifications_server():
     """
-    Starts a simple push notifications injesting server and yields its URL.
+    Starts a simple push notifications ingesting server and yields its URL.
     """
     host = '127.0.0.1'
     port = find_free_port()
@@ -148,7 +148,7 @@ async def test_notification_triggering_after_config_change_e2e(
     notifications_server: str, agent_server: str, http_client: httpx.AsyncClient
 ):
     """
-    Tests notification triggering after setting the push notificaiton config in a seperate call.
+    Tests notification triggering after setting the push notification config in a separate call.
     """
     # Configure an A2A client without a push notification config.
     a2a_client = ClientFactory(


src/a2a/utils/signing.py
```diff
@@ -0,0 +1,161 @@
...
+                canonical_payload = canonicalize_agent_card(agent_card)
+                encoded_payload = base64url_encode(
+                    canonical_payload.encode('utf-8')
+                ).decode('utf-8')
+                token = f'{agent_card_signature.protected}.{encoded_payload}.{agent_card_signature.signature}'
+
+                jws.verify(
+                    token=token,
+                    key=verification_key,
+                    algorithms=None,
```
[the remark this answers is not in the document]
sokoliva: Done

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,161 @@
...
+        return agent_card
+
+    return agent_card_signer
+
+
+def create_signature_verifier(
+    key_provider: Callable[
+        [str | None, str | None], str | bytes | dict[str, Any] | Key
+    ],
+) -> Callable[[AgentCard], None]:
```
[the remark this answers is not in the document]
sokoliva: Makes sense. Done.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,161 @@
...
+    """Creates a function that verifies AgentCard signatures.
+
+    Args:
+        key_provider: A callable that takes key-id (kid) and JSON web key url (jku) and returns the verification key.
```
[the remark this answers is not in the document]
sokoliva: Done

tests/utils/test_signing.py
```diff
@@ -0,0 +1,145 @@
...
+):
+    """Test canonicalize_agent_card with defaults, optionals, and exceptions.
+
+    - extensions is omitted as it's not set and optional.
+    - protocolVersion is included because it's always added by canonicalize_agent_card.
+    - signatures should be omitted.
+    """
+    sample_agent_card.signatures = (
+        [{'protected': 'protected_header', 'signature': 'test_signature'}],
+    )
```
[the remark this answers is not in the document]
sokoliva: Makes sense. Done

src/a2a/client/base_client.py
```diff
@@ -270,12 +271,16 @@ async def get_card(
         Args:
             context: The client call context.
             extensions: List of extensions to be activated.
+            key_provider: A callable that takes key-id (kid) and JSON web key url (jku)
+            and returns the verification key for signature verification.
```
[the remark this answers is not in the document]
sokoliva: Fixed.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+    raise ImportError(
+        'A2A Signing requires python-jose to be installed. '
+        'Install with: '
+        "'pip install a2a-sdk[signing]'"
+    ) from e
+
+from a2a.types import AgentCard, AgentCardSignature
+
+
+def clean_empty(d: Any) -> Any:
```
lkawka: Ideally, we would do without such function and have the serializer do it for us. Have you tried the json mode on the `model_dump` function [1] ?

[1]: https://docs.pydantic.dev/latest/concepts/serialization/#json-mode

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
+import json
+
+from collections.abc import Callable
+from typing import Any
+
+
+try:
+    from jose import jws
```
lkawka: Can you explain the rational behind using this library?

IMHO `PyJWT` seems like a better option. It has more stars, is sponsored by a big company and is used by FastAPI.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+        exclude={'signatures'},
+        exclude_defaults=True,
+        by_alias=True,
+    )
+    # Ensure 'protocol_version' is always included
+    protocol_version_alias = (
+        AgentCard.model_fields['protocol_version'].alias or 'protocol_version'
+    )
+    if protocol_version_alias not in card_dict:
+        card_dict[protocol_version_alias] = agent_card.protocol_version
```
lkawka: Why wouldn't the protocol version be included? card_dict is constructed from agent_card, so if it wasn't included initially, this won't change much either?

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Returns:
+        A callable that takes an AgentCard and returns the modified AgentCard with a signature.
+    """
+
+    def agent_card_signer(agent_card: AgentCard) -> AgentCard:
+        """The actual card_modifier function."""
+        canonical_payload = canonicalize_agent_card(agent_card)
+
+        headers = {'kid': kid, 'typ': 'JOSE'}
```
lkawka: The spec says `typ: SHOULD be set to "JOSE" for JWS`, so let's make it configurable.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+            payload=canonical_payload.encode('utf-8'),
+            key=signing_key,
+            headers=headers,
+            algorithm=alg,
+        )
+
+        # The result of jws.sign is a compact serialization: HEADER.PAYLOAD.SIGNATURE
+        protected_header, _, signature = jws_string.split('.')
+
+        agent_card_signature = AgentCardSignature(
```
lkawka, line 84: what about the `headers` parameter? I think that we could make it configurable as well.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+    Returns:
+        A callable that takes an AgentCard, and raises an error if none of the signatures are valid.
+    """
+
+    def signature_verifier(
+        agent_card: AgentCard,
+    ) -> None:
+        """The actual signature_verifier function."""
+        if not agent_card.signatures:
+            raise JOSEError('No signatures found on AgentCard')
```
lkawka: You shouldn't use a library's error here. Such usage is misleading, users may incorrectly assume that it's thrown by the underlaying library. This is should be our error.

tests/integration/test_client_server_integration.py
```diff
@@ -1,13 +1,14 @@
...
-from typing import NamedTuple
+from typing import NamedTuple, Any
 from unittest.mock import ANY, AsyncMock, patch
 
 import grpc
 import httpx
 import pytest
 import pytest_asyncio
 from grpc.aio import Channel
+from jose.backends.base import Key
```
lkawka: Since you use this library in tests, it should also be added to dev dependencies in `pyproject.toml`.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+    return agent_card_signer
+
+
+def create_signature_verifier(
+    key_provider: Callable[
+        [str | None, str | None], str | bytes | dict[str, Any] | Key
+    ],
+    algorithms: list[str],
+) -> Callable[[AgentCard], None]:
+    """Creates a function that verifies AgentCard signatures.
```
lkawka: It would be useful to document the behavior we have chosen here i.e. we check if at least one signature matches the key, otherwise expect an exception.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+        algorithms: List of acceptable algorithms for verification used to prevent algorithm confusion attacks.
+
+    Returns:
+        A callable that takes an AgentCard, and raises an error if none of the signatures are valid.
+    """
+
+    def signature_verifier(
+        agent_card: AgentCard,
+    ) -> None:
+        """The actual signature_verifier function."""
```
lkawka: `Verifies agent cards signatures.`

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,165 @@
...
+
+                jws.verify(
+                    token=token,
+                    key=verification_key,
+                    algorithms=algorithms,
+                )
+                # Found a valid signature, exit the loop and function
+                break
+            except JOSEError as e:
+                last_error = e
```
lkawka: Why not just rethrow the error here?

tests/integration/test_client_server_integration.py
```diff
@@ -827,3 +842,266 @@ async def test_base_client_sends_message_with_extensions(
 
     if hasattr(transport, 'close'):
         await transport.close()
+
+
+@pytest.mark.asyncio
+async def test_json_transport_get_signed_base_card_no_initial(
```
lkawka: I don't understand the last part of the name of this test

src/a2a/utils/signing.py
sokoliva: Unfortunately, the only solution I found is adding a function for removing empty strings, sets and dicts as seen in signing.py. I did not find Pydantic solution for removing all falsy values.

src/a2a/utils/signing.py
sokoliva: There is no specific reason behind using `jose` over `PyJWT` library. Based on your reasoning, I agree that `PyJWT` is a better option . I will change it to `PyJWT`.

src/a2a/utils/signing.py
sokoliva: Acknowledged.

src/a2a/utils/signing.py
sokoliva: Acknowledged.

tests/integration/test_client_server_integration.py
sokoliva: "no initial" refers to JsonRpcTransport being initialized without an initial card, forcing it to fetch the base card from the server in get_card method. 

I will rename it to `test_json_transport_get_signed_base_card` and leave the description to avoid confusion.

src/a2a/utils/signing.py
sokoliva: It seems to be a no longer needed leftover from my former version of `canonicalize_agent_card`. I will remove it.

src/a2a/utils/signing.py
sokoliva: Acknowledged.

tests/integration/test_client_server_integration.py
sokoliva: This is no longer needed since switching to PyJWT library.

src/a2a/utils/signing.py
sokoliva: Acknowledged.

src/a2a/utils/signing.py
sokoliva: The current logic is to check if any of the signatures are valid so we don't mind if some of the verifications fail. An error should be raised only if none of the signatures are valid.

src/a2a/utils/signing.py
sokoliva: Acknowledged.

src/a2a/utils/signing.py
lkawka: Ok, but then it doesn't make sense to add error details from just a single check.

src/a2a/utils/signing.py
lkawka: Ok

One more thing, please mark this method as private by renaming it to `_clean_empty`.

tests/integration/test_client_server_integration.py
```diff
@@ -9,6 +9,7 @@
 import pytest_asyncio
 from grpc.aio import Channel
 
+from jwt.api_jwk import PyJWK
```
lkawka, line 12: The same issue as with the previous library applies here as well. You should include this in the dev dependencies list.

src/a2a/utils/signing.py
```diff
@@ -0,0 +1,178 @@
...
+    if isinstance(d, dict):
+        cleaned_dict: dict[Any, Any] = {k: clean_empty(v) for k, v in d.items()}
+        return {k: v for k, v in cleaned_dict.items() if v}
+    if isinstance(d, list):
+        cleaned_list: list[Any] = [clean_empty(v) for v in d]
+        return [v for v in cleaned_list if v]
+    return d if d not in ['', [], {}] else None
+
+
+def canonicalize_agent_card(agent_card: AgentCard) -> str:
```
lkawka: This function doesn't require PyJWT library, I would separate it into a separate file, so that developers wanting to use it without other functions in this file, can do so without depending on PyJWT.

tests/integration/test_client_server_integration.py
sokoliva: Makes sense. Thanks.

src/a2a/utils/signing.py
sokoliva: Makes sense.

src/a2a/utils/signing.py
lkawka: This seems unresolved

sokoliva: I put the behaviour description under signature_verifier because I felt it was more appropriate. WDYT?

sokoliva: Resolved in person.

PR #611, opened 2026-01-08
feat: support disabling OTel instrumentation via env var

### LINKED ISSUE #604: [Feat]: Support disabling A2A OTEL Instrumentation

The A2A server instrumentation for OpenTelemetry spans is too verbose. The current way to disable A2A spans is to disable at the export level. The current library disables instrumentation based on [import error only](https://github.com/a2aproject/a2a-python/blob/86c6759ce209db5575d6cf9c6e596d1cb6bf6aa1/src/a2a/utils/telemetry.py#L77)

What I would like to do is disable this at the instrumentation level rather than the export level. This can be done in a few ways; one that may be most simple to users is to configure an environment variable `A2A_OTEL_INSTRUMENTATION` that supports disabling spans for the a2a python library

- Custom exporter that filters out a2a.server.* spans; this is an antipattern and can cause broken trace trees
- Injecting a custom tracer provider for the a2a-python module

### Additional context

https://github.com/a2aproject/a2a-python/issues/295 identifies that tracing is only done if a tracer is configured. The underlying issue around noisy tracing or disabling tracing is not addressed. 

### Code of Conduct

vinoo999:
# Description

Support disabling OTEL instrumentation via the a2a-sdk using an environment variable `OTEL_A2A_SDK_INSTRUMENTATION_ENABLED`

Fixes #604  🦕

Release-As: 0.3.23

### DIFF
--- a/src/a2a/utils/telemetry.py
+++ b/src/a2a/utils/telemetry.py
@@ -71,6 +71,7 @@ import os
 
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
+
 from typing_extensions import Self
 
 


src/a2a/utils/telemetry.py
```diff
@@ -99,12 +132,9 @@ def __getattr__(self, name: str) -> Any:
     _SpanKind = _NoOp()  # type: ignore
     StatusCode = _NoOp()  # type: ignore
 
-SpanKind = _SpanKind
+SpanKind = _SpanKind  # type: ignore
```
vinoo999, line 137: was receiving unbound linting errors since i moved the assignment to line 132 rather than exception block.

src/a2a/utils/telemetry.py
```diff
@@ -74,11 +85,33 @@ def internal_method(self):
...
+    otel_installed = True
+
 except ImportError:
     logger.debug(
         'OpenTelemetry not found. Tracing will be disabled. '
         'Install with: \'pip install "a2a-sdk[telemetry]"\''
     )
+    otel_installed = False
+
+ENABLED_ENV_VAR = 'OTEL_A2A_SDK_INSTRUMENTATION_ENABLED'
```
benkawecki:
```suggestion
ENABLED_ENV_VAR = 'OTEL_INSTRUMENTATION_A2A_SDK_ENABLED'
```
Setting this to match other instrumentation env vars within the OTEL community which are prefixed by `OTEL_INSTRUMENTATION_`.

src/a2a/utils/telemetry.py
```diff
@@ -18,6 +18,16 @@
 - Automatic recording of exceptions and setting of span status.
 - Selective method tracing in classes using include/exclude lists.
 
+Configuration:
+- Environment Variable Control: OpenTelemetry instrumentation can be
+  disabled using the `OTEL_INSTRUMENTATION_A2A_SDK_ENABLED` environment
```
lkawka, line 23: I would prefer that the default value for the flag is 0/false which are usually the defaults for unset values. 

How about renaming it to `A2A_DISABLE_OTEL_INSTRUMENTATION`?

pyproject.toml
```diff
@@ -72,9 +72,10 @@ exclude = ["tests/"]
 testpaths = ["tests"]
 python_files = "test_*.py"
 python_functions = "test_*"
-addopts = "-ra --strict-markers"
+addopts = "-ra --strict-markers --dist loadgroup"
 markers = [
   "asyncio: mark a test as a coroutine that should be run by pytest-asyncio",
+  "xdist_group: mark a test to run in a specific sequential group for isolation",
```
lkawka, line 78: Please explain why you need this extra dependency. Can we do without it?

vinoo999: The new tests need to reload the telemetry module with different `OTEL_INSTRUMENTATION_A2A_SDK_ENABLED` environment variable values. The module-level code in `telemetry.py` checks this env var once at import time, so changing it mid-test-run doesn't work unless you reload the module. xdist prevents race conditions when modifying the module level code vars so that these tests don't set/re-set the env var accordingly. 

This dependency is only added for `dev` is there an issue here?

src/a2a/utils/telemetry.py
vinoo999: This aligns with OTEL conventions

example: 
- https://github.com/elastic/elasticsearch-py/blob/063e88fe0a4d642df803f37142c2e30a74f20b91/elasticsearch/_otel.py#L34
- https://docs.datadoghq.com/opentelemetry/config/environment_variable_support/#java-specific-configuration

vinoo999: I don't feel too strongly on the name though but it is enabled in code by default. 

it could be `OTEL_INSTRUMENTATION_A2A_SDK_DISABLED` but wrapping your head around 1 for the env var disabling instrumentation while 0 enabling it is awkward. I generally prefer to avoid negation env vars

PR #644, opened 2026-01-25
feat: add auth header to push notifications

### LINKED ISSUE #585: [Feat]: PushNotificationConfig.authentication is ignored; no Authorization header is sent in push notifications

The A2A protocol spec states that when a client provides a `PushNotificationConfig` with an authentication scheme (e.g. `"schemes": ["Bearer"]`), the A2A server **must authenticate** when sending push notifications to the client’s webhook.
Example config:

```json
"configuration": {
  "pushNotificationConfig": {
    "url": "CALLBACK-URL",
    "token": "secure-client-token-for-task-aaa",
    "authentication": {
      "schemes": ["Bearer"]
    }
  }
}
```

However, the Python implementation (`BasePushNotificationSender`) **completely ignores `authentication`** and sends **no `Authorization` header**.
It only attaches:

```
X-A2A-Notification-Token: <token>
```

This means that webhook endpoints cannot authenticate the caller and cannot follow the security model described in the spec.

This appears to be a **spec compliance gap**: push notification authentication is described by the protocol but not implemented in the Python server.

I would like the Python server to:

* Honor `PushNotificationConfig.authentication`
* Support at least the `"Bearer"` scheme
* Automatically add the appropriate `Authorization` header
* Match the spec examples by sending both:

  * `X-A2A-Notification-Token`
  * `Authorization: Bearer <token_or_jwt>`

As a workaround, we currently:

* **Subclass `BasePushNotificationSender`**
* Override `_dispatch_notification`
* Inject our own `Authorization: Bearer <jwt>` header

This works, but:

* It duplicates logic that should be part of the framework
* It breaks consistency between Python and other A2A implementations
* It makes spec-compliant webhook security non-standard and harder to maintain

A built-in implementation would make push notification authentication reliable, consistent, and aligned with the A2A spec.

wondr-wclabs:
## Description
- Add Authorization header for push notifications using PushNotificationConfig.authentication (prefers Bearer when present)
- Keep existing X-A2A-Notification-Token behavior
- Add unit test covering auth header behavior

## Testing
- uv run pytest tests/server/tasks/test_push_notification_sender.py

Fixes #585

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
   "googleapis-common-protos>=1.70.0",
   "culsans>=0.11.0 ; python_full_version < '3.13'",
   "packaging>=24.0",
+  "typing-extensions>=4.0.0",
 ]
 
 classifiers = [
--- a/src/a2a/server/tasks/base_push_notification_sender.py
+++ b/src/a2a/server/tasks/base_push_notification_sender.py
@@ -82,9 +82,7 @@ class BasePushNotificationSender(PushNotificationSender):
     ) -> bool:
         url = push_info.url
         try:
-            headers = None
-            if push_info.token:
-                headers = {'X-A2A-Notification-Token': push_info.token}
+            headers = self._build_headers(push_info)
 
             response = await self._client.post(
                 url,
@@ -103,3 +101,23 @@ class BasePushNotificationSender(PushNotificationSender):
             )
             return False
         return True
+
+    @staticmethod
+    def _authorization_header(
+        push_info: TaskPushNotificationConfig,
+    ) -> str | None:
+        auth = push_info.authentication
+        if not auth.scheme or not auth.credentials:
+            return None
+        return f'{auth.scheme} {auth.credentials}'
+
+    def _build_headers(
+        self, push_info: TaskPushNotificationConfig
+    ) -> dict[str, str] | None:
+        headers: dict[str, str] = {}
+        if push_info.token:
+            headers['X-A2A-Notification-Token'] = push_info.token
+        authorization = self._authorization_header(push_info)
+        if authorization:
+            headers['Authorization'] = authorization
+        return headers or None
--- a/tests/server/tasks/test_push_notification_sender.py
+++ b/tests/server/tasks/test_push_notification_sender.py
@@ -8,6 +8,7 @@ from a2a.server.tasks.base_push_notification_sender import (
     BasePushNotificationSender,
 )
 from a2a.types.a2a_pb2 import (
+    AuthenticationInfo,
     StreamResponse,
     Task,
     TaskArtifactUpdateEvent,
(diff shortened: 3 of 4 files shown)


wondr-wclabs:
Follow-up: CI lint flagged _NoOp.__enter__ return typing in src/a2a/utils/telemetry.py. Updated to use Self with a py3.10-safe typing_extensions import and added typing-extensions to core deps; uv.lock updated accordingly.

wondr-wclabs:
Added edge-case coverage for _authorization_header: no credentials, empty schemes, non-Bearer scheme selection, empty-string scheme filtering, and validation error for None scheme entries. All tests in tests/server/tasks/test_push_notification_sender.py pass locally.

wondr-wclabs:
All checks are green ✅ This PR is ready for review/merge when you have a moment. Happy to make any adjustments.

wondr-wclabs:
Rebased this onto current `main` and resolved the merge conflict against the newer protobuf-based push notification sender.

The original implementation was written against the older Pydantic `PushNotificationConfig` model with `authentication.schemes: list[str]`. Current `main` now dispatches `TaskPushNotificationConfig` protobuf messages, where authentication is a single `scheme` plus `credentials`, so I adapted the header builder to that shape rather than preserving the old "prefer Bearer from a list" logic. The behavior after the rebase is:

- keep the existing `X-A2A-Notification-Token` header for legacy token configs;
- add `Authorization: <scheme> <credentials>` only when both protobuf auth fields are present;
- leave headers as `None` when neither token nor complete auth is configured.

Local checks run after the rebase:
- `uv lock --check`
- `uv run pytest tests/server/tasks/test_push_notification_sender.py -q`
- `uv run ruff check`
- `uv run ruff format --check`
- `uv run ty check --output-format=github`
- `uv run pytest -q --ignore=tests/integration/cross_version/client_server/test_client_server.py`

The full unfiltered pytest run only failed in `tests/integration/cross_version/client_server/test_client_server.py` because the external 0.3 server process did not open its port within that test fixture’s 3-second local timeout. The same file fails when run alone locally; the sender tests and the rest of the suite pass.

wondr-wclabs:
Rebased this branch onto current `main` (`75991c1`) after the install-smoke matrix update landed. The rebase was clean; the PR diff is still limited to the push-notification auth header implementation, its tests, and the existing `typing-extensions` dependency metadata.

I reran the local checks that cover this change:

- `uv lock --check`
- `uv run pytest tests/server/tasks/test_push_notification_sender.py -q` — 10 passed
- `uv run ruff check`
- `uv run ruff format --check`
- `uv run ty check --output-format=github`
- `uv run pytest -q --ignore=tests/integration/cross_version/client_server/test_client_server.py` — 1736 passed, 90 skipped, 3 xfailed, 1 xpassed

The ignored cross-version fixture is the same external-server timeout case I noted earlier; it is outside the push sender code path. GitHub checks have restarted on the rebased branch.

PR #654, opened 2026-01-31
feat: support async card modifiers

### LINKED ISSUE #647: [Feat]: allow async extended card modifier

We'd like to be able to modify the authenticated agent card based on data we fetch via an HTTP request. That request uses `async`, so right now we are unable to use an extended agent card modifier to do the job (since we'd need to block on the `async` call and thus block the entire server).

`extended_card_modifier` should be an async method, and inside of `handle_authenticated_agent_card` it should be called with `await`. This way we can make async calls inside of the extended card modifier.

Our alternative is to extend the application class and the adapter class so that we have a custom adapter and implementation of the `handle_authenticated_agent_card` method. This is quite cumbersome and not very future proof.

### Additional context

N/A

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

cbugwadia32:
# Description

Fixes #647 🦕

Release-As: 0.3.23

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -23,7 +23,7 @@ jobs:
         run: |
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - name: Install dependencies
-        run: uv sync --locked --dev
+        run: uv sync --locked
 
       - name: Run Ruff Linter
         id: ruff-lint
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -53,7 +53,7 @@ jobs:
         run: |
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - name: Install dependencies
-        run: uv sync --locked --dev --extra all
+        run: uv sync --locked
       - name: Run tests and check coverage
         run: uv run pytest --cov=a2a --cov-report term --cov-fail-under=88
       - name: Show coverage summary in log
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,7 +89,6 @@ style = "pep440"
 dev = [
   "datamodel-code-generator>=0.30.0",
   "mypy>=1.15.0",
-  "PyJWT>=2.0.0",
   "pytest>=8.3.5",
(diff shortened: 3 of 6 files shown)


cbugwadia32:
@ishymko (since you are assigned to the corresponding issue) - let me know if you want me to do it how Gemini suggests, or if it's fine to have this be a breaking change. I'm not sure what the expectation is since [we're on major version 0 currently](https://semver.org/#spec-item-4).

tests/utils/test_signing.py
ishymko, line 1: May I ask to revert this file in the current PR (unless it failed CI - which is a problem on its own)?

I also noticed inconsistent imports in this file, but it'd rather fix separately and also ideally tune ruff linter to prevent them from being merged in the first place (and if you're interested in fixing this it'd be appreciated!).

Thank you!

src/a2a/utils/helpers.py
```diff
@@ -368,3 +373,13 @@ def canonicalize_agent_card(agent_card: AgentCard) -> str:
     # Recursively remove empty values
     cleaned_dict = _clean_empty(card_dict)
     return json.dumps(cleaned_dict, separators=(',', ':'), sort_keys=True)
+
+
+async def apply_optional_awaitable(
```
ishymko: What do you think about something like

```python
async def maybe_await(value: Union[T, Awaitable[T]]) -> T:
    if inspect.isawaitable(value):
        return await value
    return value
```

so that usages continue to call the function and only result is postprocessed? It feels a bit more explicit and you don't need args/kwargs boilerplate to proxy parameters:

```diff
-    self.extended_card_modifier(base_card, context)
+    await maybe_await(self.extended_card_modifier(base_card, context))
```

tests/utils/test_signing.py
cbugwadia32: Yeah this happened because I initially had modified this file (to make the modifiers async) and then ultimately changed it to leave it as is, but the reformatting remained. I've reset it to latest main.

cbugwadia32:
Should be good now @ishymko - thanks!

ishymko (approved):
🚀

PR #635, opened 2026-01-20
fix: Update invocation metadata from `grpc.aio.ServicerContext` to use the correct syntax

jubonni:
# Description
invocation_metadata from grpc.aio.ServicerContext was treated as a property instead of a method which prevented extracting the invocation metadata

- [x] Follow the [`CONTRIBUTING` Guide](https://github.com/a2aproject/a2a-python/blob/main/CONTRIBUTING.md).
- [x] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
  - Important Prefixes for [release-please](https://github.com/googleapis/release-please):
    - `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
    - `feat:` represents a new feature, and correlates to a SemVer minor.
    - `feat!:`, or `fix!:`, `refactor!:`, etc., which represent a breaking change (indicated by the `!`) and will result in a SemVer major.
- [x] Ensure the tests and linter pass (Run `bash scripts/format.sh` from the repository root to format)
- [x] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕

PR #676, opened 2026-02-09
fix: gRPC metadata header casing and invocation_metadata() call

### LINKED ISSUE #656: [Bug]: Metadata key 'X-A2A-Extensions' is invalid: INTERNAL: Illegal header key ( gRPC metadata key X-A2A-Extensions is invalid )
### What happened?

When using the A2A Python SDK with gRPC transport, the client injects metadata key X-A2A-Extensions. gRPC requires metadata keys to be lowercase, so this causes INTERNAL: Illegal header key at runtime.
### Environment
OS: macOS (darwin 24.6.0)
Python: 3.12
grpcio: 1.76.0
a2a-sdk: 0.3.22
### Steps to Reproduce
Use ClientFactory (or any SDK gRPC client) to send a message over gRPC.
The SDK automatically adds X-A2A-Extensions metadata.
gRPC fails with Illegal header key.
### Actual Result
Metadata key 'X-A2A-Extensions' is invalid: INTERNAL: Illegal header key
### Expected Result
SDK should use a lowercase metadata key (e.g. x-a2a-extensions), which is valid for gRPC.
### Temporary Workaround
Override the header constant at runtime:
from a2a.extensions import common as a2a_ext_commonfrom a2a.client.transports import grpc as a2a_grpc_transporta2a_ext_common.HTTP_EXTENSION_HEADER = "x-a2a-extensions"a2a_grpc_transport.HTTP_EXTENSION_HEADER = "x-a2a-extensions"
Or disable extensions by default in gRPC metadata.
### Suggested Fix
Use a lowercase metadata key for gRPC transport, or ensure the header key is normalized to lowercase before attaching metadata.

### Relevant log output

```shell

```

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

guglielmo-san:
This PR fixes two issues related to the grpc support:

1. The extension headers key in the metadata are supposed to be lower key ([ref](https://github.com/grpc/grpc/blob/ce463633548e47c05343bc1243f9ea95b700a908/src/core/lib/surface/validate_metadata.cc#L99))
2. Invokes correctly the `servicer_context.invocation_metadata` as a [method](https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.ServicerContext.invocation_metadata) instead of as a property

For each of these issues was respectively opened a community PR (#635 , #673). To speed up the integration of these changes, this new PR has been created to integrate both changes at the same time.

Fixes #656


src/a2a/client/transports/grpc.py
```diff
@@ -64,10 +64,13 @@ def _get_grpc_metadata(
         extensions: list[str] | None = None,
     ) -> list[tuple[str, str]] | None:
         """Creates gRPC metadata for extensions."""
-        if extensions is not None:
-            return [(HTTP_EXTENSION_HEADER, ','.join(extensions))]
-        if self.extensions is not None:
-            return [(HTTP_EXTENSION_HEADER, ','.join(self.extensions))]
+        extensions_to_use = (
+            extensions if extensions is not None else self.extensions
+        )
```
holtskinner:
```suggestion
        extensions_to_use = extensions or self.extensions
```

This should handle the same cases, unless you want `[]` to be handled differently?

src/a2a/client/transports/grpc.py
```diff
@@ -64,10 +64,13 @@ def _get_grpc_metadata(
...
+        if extensions_to_use is not None:
```
holtskinner:
```suggestion
        if extensions_to_use:
```

src/a2a/client/transports/grpc.py
guglielmo-san: makes sense

PR #698, opened 2026-02-17
fix: fix bad "list tasks" merge for JSON-RPC

ishymko:
Fixes https://github.com/a2aproject/a2a-python/pull/697#discussion_r2817860175, bad merge in #696.

Cover "list tasks" in client-server integration tests which would prevent it.

Re #559.

### DIFF
--- a/src/a2a/server/request_handlers/jsonrpc_handler.py
+++ b/src/a2a/server/request_handlers/jsonrpc_handler.py
@@ -26,7 +26,6 @@ from a2a.types.a2a_pb2 import (
     GetTaskRequest,
     ListTaskPushNotificationConfigRequest,
     ListTasksRequest,
-    ListTasksResponse,
     Message,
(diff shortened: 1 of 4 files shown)


src/a2a/server/request_handlers/jsonrpc_handler.py
holtskinner: You might need to add another option for an error. Note, the other methods like `on_get_task()` return dicts.

```suggestion
    ) -> ListTasksResponse | JSONRPCError:
```

Double check what the correct type is for `_build_error_response`

holtskinner (changes requested):

src/a2a/server/request_handlers/jsonrpc_handler.py
ishymko: Yes I noticed it already and drafted the PR, leftovers from a bad merge.

PR #709, opened 2026-02-18
feat(server): implement `Resource Scoping` for tasks and push notifications

### LINKED ISSUE #610: [Feat]: Resource Scoping

The new task/list method introduced a concept of resource scoping and the [specification](http://screen/7T6DKaPLiRJUrg8) dictates limiting the scope of returned tasks to the some authorized scope. It is not yet implemented.

Limit the scope of returned tasks to some authorized scope.

_No response_

### Additional context

_No response_

### Code of Conduct

sokoliva:
## Description
Introduces caller indentity isolation to ensure clients only access authorized resources, as mandated by the [A2A spec](https://a2a-protocol.org/latest/specification/#131-data-access-and-authorization-scoping).

- Add `owner`field to `TaskMixin` and `PushNotificationConfig` database models.
 - Add `last_updated` field to `TaskMixin` for optimized sorting and indexing.
- Update `DatabaseTaskStore`, `InMemoryTaskStore` and `DatabasePushNotificationConfigStore` to use `OwnerResolver`.
- Add Resource Scoping related Unit tests.
- Add Alembic configuration to enable users to update their own databases with non-optional `owner` column in `tasks` and `push_notification_configs` table and optional `last_updated` and index `(owner, last_updated)` in `tasks`  .
- Distribute alembic configuration, enable CLI commands such as `uv run a2a-db` for database updating.

## Note
- In `src/a2a/server/tasks/database_task_store.py` `list` method, Gemini suggested a refactor of pagination.  I thoroughly reviewed it and comfirmed that the logic is the same and that readability of code improved so I decided to accept it.
- It seems there was a functional bug in [InMemoryPushNotificationConfigStore](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/inmemory_push_notification_config_store.py) `delete_info` method. When `config_id` is None and only `task_id` was provided it would search for configs mapped to `task_id` with `config.id=task_id`, contrary to `delete_info` method of  [DatabasePushNotificationConfigStore](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/database_push_notification_config_store.py) where if config_id is None, all configurations for the task are deleted. Unfortunately, I did not find intended behavior defined in the spec, but behavior of `DatabasePushNotificationConfigStore's` `delete_info` seems more logical. 
## Breaking changes
- added non-optional owner field to the Task Model. Use alembic configuration to update your database.

Fixes #610 🦕


scutuatua-crypto (approved):

src/a2a/server/owner_resolver.py
```diff
@@ -0,0 +1,18 @@
...
+
+# Example Default Implementation
+def resolve_user_scope(context: ServerCallContext) -> str:
+    """Resolves the owner scope based on the user in the context."""
+    if not context:
+        return 'unknown'
+    if not context.user:
+        raise ValueError('User not found in context.')
+    # Example: Basic user name. Adapt as needed for your user model.
+    return context.user.user_name
```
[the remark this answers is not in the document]
sokoliva: Outdated

src/a2a/server/tasks/database_task_store.py
```diff
@@ -104,12 +109,14 @@ async def _ensure_initialized(self) -> None:
...
 
-    def _to_orm(self, task: Task) -> TaskModel:
+    def _to_orm(self, task: Task, owner: str) -> TaskModel:
         """Maps a Pydantic Task to a SQLAlchemy TaskModel instance."""
         return self.task_model(
             id=task.id,
             context_id=task.context_id,
             kind=task.kind,
+            owner=owner,
+            last_updated=task.status.timestamp,
```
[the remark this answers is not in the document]
sokoliva: fixed

pyproject.toml
```diff
@@ -323,3 +323,89 @@ docstring-code-format = true
...
+# module = "ruff"
+# options = "check --fix REVISION_SCRIPT_FILENAME"
+#
+# [[tool.alembic.post_write_hooks]]
+# Alternatively, use the exec runner to execute a binary found on your PATH
+# name = "ruff"
+# type = "exec"
+# executable = "ruff"
+# options = "check --fix REVISION_SCRIPT_FILENAME"
+
```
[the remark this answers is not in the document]
sokoliva: Not true. This setup was generated by the alembic official setup command for [using pyproject.toml for configuration ](https://alembic.sqlalchemy.org/en/latest/tutorial.html#using-pyproject-toml-for-configuration)"alembic init --template pyproject alembic"

The logging setup in `env.py` does not fail.

src/a2a/server/models.py
```diff
@@ -127,6 +129,8 @@ class TaskMixin:
     kind: Mapped[str] = mapped_column(
         String(16), nullable=False, default='task'
     )
+    owner: Mapped[str] = mapped_column(String(255), nullable=False, index=True)
+    last_updated: Mapped[datetime] = mapped_column(String(22), nullable=True)
```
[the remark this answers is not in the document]
sokoliva: outdated

alembic/README
```diff
@@ -0,0 +1,56 @@
...
+uv run alembic history --verbose
+```
+
+### Running Migrations
+```bash
+# Upgrade to the latest version
+uv run alembic upgrade head
+
+# Downgrade by one version
+uv run alembic downgrade base
```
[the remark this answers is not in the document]
sokoliva: Done.

src/a2a/server/models.py
```diff
@@ -16,7 +18,7 @@ def override(func):  # noqa: ANN001, ANN201
 
 
 try:
-    from sqlalchemy import JSON, Dialect, LargeBinary, String
+    from sqlalchemy import JSON, Dialect, Index, LargeBinary, String
```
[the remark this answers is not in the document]
sokoliva: Obsolete. Will use string.

src/a2a/server/tasks/push_notification_config_store.py
```diff
@@ -8,16 +9,26 @@ class PushNotificationConfigStore(ABC):
 
     @abstractmethod
     async def set_info(
-        self, task_id: str, notification_config: PushNotificationConfig
+        self,
+        task_id: str,
+        notification_config: PushNotificationConfig,
+        context: ServerCallContext | None = None,
```
ishymko: Here and below for `context` parameter: seems like usages were not updated to provide this parameter and it's going to be `None` always. Given that it's going to be for 1.0 breaking release either way, I'd make it mandatory to make sure it's always passed and owner can be checked.

src/a2a/server/tasks/inmemory_task_store.py
```diff
@@ -15,45 +18,70 @@
...
     ) -> a2a_pb2.ListTasksResponse:
-        """Retrieves a list of tasks from the store."""
+        """Retrieves a list of tasks from the store, for the given owner."""
+        owner = self.owner_resolver(context)
+        logger.debug('Listing tasks for owner %s with params %s', owner, params)
+
         async with self.lock:
-            tasks = list(self.tasks.values())
+            owner_tasks = self.tasks.get(owner, {})
+            tasks = list(owner_tasks.values())
```
ishymko: nit: this code uses a bit different pattern compared to `get`, I suggest to either use fallback to an empty dict in both places or explicit `if` here as well.

Same for `delete`. Maybe a helper private `get_owner_tasks` can be useful here.

pyproject.toml
```diff
@@ -347,3 +348,89 @@ docstring-code-format = true
 docstring-code-line-length = "dynamic"
 quote-style = "single"
 indent-style = "space"
+
+
+[tool.alembic]
```
ishymko, line 354: nit: I suggest linking original docs via comment here and remove commented out values as they add some noise to the file.

alembic/README
ishymko, line 1: Question on the developer experience here: if one needs to migrate their database - which steps are required? Currently readme mixes information for the package developers on how to add more migrations with the usage for integrations (how to run migrations).

I also believe that it assumes cloning the repository in order to run them, as it usually doesn't happen when one just uses package as a dependency. This creates some risk as one should make sure they switched to the appropriate commit to avoid applying migrations from `main` branch not yet released to PyPi?

Is it possible to distribute migrations together with the package itself and provide instructions on running them against the installed package version? We can also consider optional parameter which one may use to automatically apply migrations at runtime (default to `False`).

References:
- https://stackoverflow.com/q/73633063
- https://github.com/sqlalchemy/alembic/discussions/1104

src/a2a/server/tasks/push_notification_config_store.py
sokoliva: Done

src/a2a/server/tasks/inmemory_task_store.py
sokoliva: Agreed, it is better to be consistent with syntax. Done. I used `get(owner, {})`.

sokoliva: I also added the `get_owner_tasks` private helper.

alembic/README
sokoliva: Done and tested locally.

pyproject.toml
sokoliva: Done

src/a2a/migrations/README.md
ishymko, line 1: nit: `README.md`?

src/a2a/migrations/versions/6419d2d130f6_add_owner_to_task.py
ishymko, line 1: We should use migration(s) for all schema updates made in this PR: `last_updated` and indexes as well.

Currently we have a [flag](https://github.com/a2aproject/a2a-python/blob/cced34d4eb22f7ec78422b2d04ce54924e132215/src/a2a/server/tasks/database_task_store.py#L39) to create tables via [`create_all`](https://docs.sqlalchemy.org/en/21/core/metadata.html#sqlalchemy.schema.MetaData.create_all) which is not going to create new columns inside existing tables. * - maybe it will create new index, but it requires checking.

src/a2a/migrations/env.py
```diff
@@ -0,0 +1,102 @@
...
+
+    await connectable.dispose()
+
+
+def run_migrations_online() -> None:
+    """Run migrations in 'online' mode."""
+    asyncio.run(run_async_migrations())
+
+
+if context.is_offline_mode():
```
ishymko, line 118: Is it possible to use it via CLI? Can't find it in the `cli.py` or in the docs. It could be beneficial for environments where DB updates are done only via scripts with manual DBA reviews to generate SQL instead of running it.

src/a2a/a2a_db_cli.py
ishymko, line 1: Let's add `--verbose -v` flag for debug logs, not sure about exact log levels in sqlalchemy but at least printing executed SQL statements would be good.

src/a2a/migrations/README.md
```diff
@@ -0,0 +1,84 @@
...
+# MySQL example
+export DATABASE_URL="mysql+aiomysql://user:pass@localhost/your_database_name"
+```
+
+### 2. Apply Migrations
+Always run this command after installing or upgrading the SDK to ensure your database matches the required schema.
+
+```bash
+# Bring the database to the latest version
+a2a-db
```
ishymko: Running this command in a fresh environment will fail with `alembic` package missing error. `alembic` is a dev dependency currently so for package users it won't be installed.

You can reproduce it by creating a "fake" consumer project and installing `whl` there:

```
uv build
```
which is going to output something like `dist/a2a_python-*.whl`.

And in the new folder
```
uv venv
uv pip install <whl path from uv build>
uv run a2a-db
```

I believe we have two options:

1. Instruct users to manually install `alembic` as their dev dependency if they want to use this CLI.
2. Create a separate "extra" which will be used for this.

src/a2a/server/tasks/inmemory_task_store.py
```diff
@@ -15,45 +18,72 @@
...
+    ) -> None:
         """Initializes the InMemoryTaskStore."""
         logger.debug('Initializing InMemoryTaskStore')
-        self.tasks: dict[str, Task] = {}
+        self.tasks: dict[str, dict[str, Task]] = defaultdict(dict)
         self.lock = asyncio.Lock()
+        self.owner_resolver = owner_resolver
+
+    def _get_owner_tasks(self, owner: str) -> dict[str, Task]:
+        return self.tasks.get(owner, {})
```
ishymko, line 34: I don't think that fallback to `{}` is required after changing `tasks` to `defaultdict(dict)`.

src/a2a/server/tasks/database_task_store.py
ishymko, line 1: Let's consider if some check can be added to fail fast on non-applied migrations. Currently it's going to fail during request execution, maybe it can be checked in `initialize` method to fail fast?

We can follow-up with this in a separate PR/issue as it's rather a safety net, but given that we're establishing DB migrations infrastructure it can be beneficial to research this aspect as well.

src/a2a/migrations/README.md
sokoliva: Done.

src/a2a/a2a_db_cli.py
sokoliva: Done

src/a2a/server/tasks/inmemory_task_store.py
sokoliva: Without fallback to `{ }` `owner_tasks = self._get_owner_tasks(owner)` become of type `dict[str, Task] | None` which makes `task = owner_tasks.get(task_id)` fail with "AttributeError: 'NoneType' object has no attribute 'get'".

src/a2a/migrations/versions/6419d2d130f6_add_owner_to_task.py
sokoliva: By running [tests/server/tasks/test_database_task_store.py](https://github.com/a2aproject/a2a-python/blob/main/tests/server/tasks/test_database_task_store.py) against a local database without the index and with engine disposal commented out, I confirmed that it does not create the index. I added the schema updates for `last_updated` and `index`.

src/a2a/migrations/env.py
sokoliva: Alembic uses [--sql](https://alembic.sqlalchemy.org/en/latest/offline.html#:~:text=Auto%20Generating%20Migrations-,Generating%20SQL%20Scripts%20(a.k.a.%20%E2%80%9COffline%20Mode%E2%80%9D),-The%20Importance%20of) flag for this, so I also added `--sql` flag to run migrations in offline mode.

src/a2a/server/tasks/database_task_store.py
sokoliva: Noted.

src/a2a/migrations/versions/6419d2d130f6_add_columns_owner_last_updated.py
```diff
@@ -0,0 +1,90 @@
...
+                sa.Column(
+                    'owner',
+                    sa.String(128),
+                    nullable=False,
+                    server_default=owner,
+                ),
+            )
+        if column_exists(
+            table, 'kind'
+        ):  # Check to differentiate between table of tasks and push_notification_configs. Only tasks table should have last_updated column and index.
```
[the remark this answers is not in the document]
sokoliva: That wouldn't work because table with Tasks could have a different name than `tasks`.

src/a2a/server/owner_resolver.py
```diff
@@ -0,0 +1,18 @@
...
+OwnerResolver = Callable[[ServerCallContext | None], str]
+
+
+# Example Default Implementation
+def resolve_user_scope(context: ServerCallContext | None) -> str:
+    """Resolves the owner scope based on the user in the context."""
+    if not context:
+        return 'unknown'
+    if not context.user:
+        raise ValueError('User not found in context.')
```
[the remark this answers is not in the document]
sokoliva: Since user is a non-optional field in `ServerCallContext` I decided to remove this check all together.

src/a2a/cli.py
```diff
@@ -0,0 +1,176 @@
...
+        dest='sub_verbose',
+        help='Enable verbose output (sets sqlalchemy.engine logging to INFO)',
+        action='store_true',
+    )
+    down_parser.add_argument(
+        '--sql',
+        dest='sub_sql',
+        help='Run migrations in sql mode (generate SQL instead of executing)',
+        action='store_true',
+    )
```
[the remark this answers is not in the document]
sokoliva: Done. Added a private helper `_add_shared_args`

src/a2a/migrations/README.md
sokoliva: Added a new optional dependency `cli = ["alembic>=1.14.0"]` and `ImportError` handling to cli.py, env.py and `6419d2d130f6_add_columns_owner_last_updated.py`. WDYT?

src/a2a/migrations/versions/6419d2d130f6_add_columns_owner_last_updated.py
```diff
@@ -0,0 +1,95 @@
...
+    return index_name in indexes
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Get the default value from the config (passed via CLI)
+    owner = context.config.get_main_option('owner', 'unknown')
+    tables_str = context.config.get_main_option(
+        'tables', 'tasks,push_notification_configs'
+    )
```
ishymko: Let's have two explicit different flags, one for each table and handle each table explicitly even if it introduces some code duplication.

src/a2a/migrations/README.md
```diff
@@ -0,0 +1,100 @@
...
+### 3. Apply Migrations
+Always run this command after installing or upgrading the SDK to ensure your database matches the required schema. This will upgrade the tables `tasks` and `push_notification_configs` in your database by adding columns `owner` and `last_updated` and an index `(owner, last_updated)` to the `tasks` table and a column `owner` to the `push_notification_configs` table.
+
+```bash
+uv run a2a-db
+```
+
+### 4. Customizing Defaults with Flags
+#### -o
+Allows you to pass custom values for the new `owner` column. If not set, it will default to the value `unknown`.
```
ishymko: Consider renaming to something more specific without short option. Going forward with more migrations coming we may want to introduce other parameters for other migrations.

One option is to prefix it with some disambiguated migration name.

src/a2a/migrations/README.md
```diff
@@ -0,0 +1,100 @@
...
+uv run a2a-db -t "my_table1" -t "my_table2"
+```
+#### -v
+Enables verbose output by setting `sqlalchemy.engine` logging to `INFO`.
+
+```bash
+uv run a2a-db -v
+```
+#### --sql
+Enables running migrations in `offline` mode. Migrations are generated as SQL scripts instead of being run against the database.
```
ishymko: Let's specify exactly how "Migrations are generated as SQL scripts" - printed to stdout or some file.

src/a2a/server/owner_resolver.py
```diff
@@ -0,0 +1,16 @@
...
+OwnerResolver = Callable[[ServerCallContext | None], str]
+
+
+# Example Default Implementation
+def resolve_user_scope(context: ServerCallContext | None) -> str:
+    """Resolves the owner scope based on the user in the context."""
+    if not context:
+        return 'unknown'
+    # Example: Basic user name. Adapt as needed for your user model.
+    return context.user.user_name
```
ishymko, line 16: We need to be careful with defaults here and how those defaults interact with DB migration defaults.

Using `unknown` in both places will lead to allowing everyone to access prior data.

Maybe DB default should be something more specific like `legacy_v03_no_user_info`.

src/a2a/server/owner_resolver.py
```diff
@@ -0,0 +1,16 @@
...
+
+# Definition
+OwnerResolver = Callable[[ServerCallContext | None], str]
+
+
+# Example Default Implementation
+def resolve_user_scope(context: ServerCallContext | None) -> str:
+    """Resolves the owner scope based on the user in the context."""
+    if not context:
+        return 'unknown'
```
ishymko, line 14: In which cases we won't have `context` here?

pyproject.toml
```diff
@@ -39,6 +39,7 @@ postgresql = ["sqlalchemy[asyncio,postgresql-asyncpg]>=2.0.0"]
 mysql = ["sqlalchemy[asyncio,aiomysql]>=2.0.0"]
 signing = ["PyJWT>=2.0.0"]
 sqlite = ["sqlalchemy[asyncio,aiosqlite]>=2.0.0"]
+cli = ["alembic>=1.14.0"]
```
ishymko: Let's consolidate naming with `db` part, the CLI is named `a2a-db` as a command, while it is in `cli.py` and also has just `cli` here.

To give us more room in the future I suggest to use `db` everywhere: extra name, file name, command name.

src/a2a/server/tasks/inmemory_task_store.py
ishymko: Is it a runtime or a type error? Accessing an unknown key for `defaultdict` should produce `dict`. `.get(Key)` returns `dict[str, Task] | None` but for `defaultdict` it's safe to use indexer:

```suggestion
        return self.tasks[owner]
```

src/a2a/migrations/README.md
ishymko: SG, let's update the readme as well.

sokoliva: Done.

src/a2a/server/owner_resolver.py
sokoliva: I think `if not context: return 'unknown'` should be removed when we make `ServerCallContext` non-optional everywhere. https://github.com/a2aproject/a2a-python/issues/718 

`resolve_user_scope` is being called by methods that have `ServerCallContext` as optional and they pass that same value to `resolve_user_scope`.

src/a2a/server/tasks/inmemory_task_store.py
sokoliva: Discussed offline: we decided to remove defaultdict to avoid the possibility of unnecessary clutter in database.

src/a2a/server/owner_resolver.py
sokoliva: Great point. I will change DB default to `legacy_v03_no_user_info` as you recommended.

src/a2a/migrations/README.md
sokoliva: Done. They are printed to stdout.

pyproject.toml
sokoliva: Done. Changed to `a2a-db-cli`.

src/a2a/migrations/versions/6419d2d130f6_add_columns_owner_last_updated.py
sokoliva: Great idea. Added two flags `--tasks-table` and `--push-notification-table` as suggested and removed `-t` flag.

src/a2a/migrations/README.md
sokoliva: Done. Added renamed to `--add_columns_owner_last_updated-default-owner` as suggested.

pyproject.toml
```diff
@@ -49,6 +50,7 @@ all = [
   "a2a-sdk[grpc]",
   "a2a-sdk[telemetry]",
   "a2a-sdk[signing]",
+  "a2a-sdk[a2a-db-cli]",
```
ishymko: For extra name `a2a-` prefix seems redundant and is not used in other names above, can be just `db-cli`.

src/a2a/migrations/versions/6419d2d130f6_add_columns_owner_last_updated.py
```diff
@@ -0,0 +1,157 @@
...
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Get the default value from the config (passed via CLI)
+    owner = context.config.get_main_option(
+        'add_columns_owner_last_updated_default_owner',
+        'legacy_v03_no_user_info',
+    )
+    tasks_tables = ['tasks']
+    push_notification_tables = ['push_notification_configs']
```
ishymko: There is no need for multiple tables per each type as the code uses only one. By default we should use our default table names and flag should just override name if needed (if one customized it in the code).

src/a2a/server/models.py
```diff
@@ -153,6 +148,8 @@ class TaskMixin:
     kind: Mapped[str] = mapped_column(
         String(16), nullable=False, default='task'
     )
+    owner: Mapped[str] = mapped_column(String(128), nullable=False)
+    last_updated: Mapped[str] = mapped_column(String(22), nullable=True)
```
ishymko: Why is it `String(22)`- is it a length of a date serialized to certain format? Is it possible to use DB-native date type?

src/a2a/a2a_db_cli.py
ishymko, line 1: Seems like in the end we need a dedicated [`versions_table`](https://alembic.sqlalchemy.org/en/latest/api/runtime.html#alembic.runtime.environment.EnvironmentContext.configure.params.version_table) for A2A migrations.

If one already has alembic migrations defined, both runs are going to share same table and alembic is going to fail when it finds "unknown" migration ids (either our utility when finding existing application migrations or vice versa). I tested it locally (set up a clean project with alembic and installed our package there) and it results in

```
alembic.util.exc.CommandError: Can't locate revision identified by '408f446640d1'
```

Let's customize `versions_table` for all our runs to `a2a_alembic_version` instead of the default `alembic_version`.

See https://github.com/sqlalchemy/alembic/issues/722#issuecomment-670273154.

src/a2a/migrations/versions/6419d2d130f6_add_columns_owner_last_updated.py
sokoliva: Makes sense. Done.

src/a2a/a2a_db_cli.py
sokoliva: Thank you for pointing this out and for the useful links. I updated the migration version table name.

src/a2a/server/models.py
sokoliva: changed to sqlalchemy DateTime(timezone=True).

ishymko: As we use it for sorting only, I don't think timezone is necessary here, just UTC time should be fine. Also I imagine it can be tricky in some databases, so not using timezone should make it simpler and compatible with a wider range of DBs.

ishymko (approved):

src/a2a/server/tasks/database_task_store.py
```diff
@@ -194,21 +228,20 @@ async def list(
...
             total_count = (await session.execute(count_stmt)).scalar_one()
 
-            # Use coalesce to treat NULL timestamps as empty strings,
+            # Use coalesce to treat NULL timestamps as datetime.min,
             # which sort last in descending order
             stmt = base_stmt.order_by(
-                func.coalesce(timestamp_col, '').desc(),
+                func.coalesce(
+                    timestamp_col,
+                    datetime.min.replace(tzinfo=timezone.utc),
```
lkk7, line 243: I know it's late, but I caught this now:

[MySQL docs](https://dev.mysql.com/doc/refman/9.7/en/datetime.html) say:
> The supported range is '1000-01-01 00:00:00' to '9999-12-31 23:59:59'.

Using `datetime.min` here pushes year `0001` into the query which is against the specification.

sokoliva: Thank you for pointing this out! I opened an issue about this: #1219. I was unable to replicate the issue, could you tell me more if this cause any problems to you?

lkk7: Honestly, not really. Just a spec issue that I noticed. It's not critical.
In terms of just storing data, it works on modern MySQL versions.
I'm just not sure about what will happen when you start sorting that date, doing arithmetic operations on it, etc. This could lead to an unfindable problem in the future.

PR #719, opened 2026-02-20
feat(client): expose close() and async context manager support on abstract Client

### LINKED ISSUE #689: [Feat]: Expose close() and async context manager support on abstract Client

`ClientTransport` supports the async context manager protocol since #682, and BaseClient since #688. However, the abstract `Client` class (`src/a2a/client/client.py`) — which defines the public interface for all client implementations — does not declare `close()`, `__aenter__`, or `__aexit__`.

This means code that works with the `Client` type cannot use the async context manager or call `close()`:

```python
client = await ClientFactory.connect('https://agent.example.com')
# client is typed as Client, not BaseClient

async with client:
    ...
# TypeError: 'Client' object does not support the asynchronous context manager protocol

await client.close()
# AttributeError: 'Client' object has no attribute 'close'
```

The integration tests reflect this gap — they rely on `hasattr` checks instead of the type system:

```python
if hasattr(transport, 'close'):
    await transport.close()
```

Add `close()` as an abstract method and `__aenter__`/`__aexit__` as concrete methods on `Client` in `src/a2a/client/client.py`:

```python
@abstractmethod
async def close(self) -> None:
    """Closes the client and releases any underlying resources."""

async def __aenter__(self) -> 'Client':
    return self

async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
    await self.close()
```

`BaseClient` already implements `close()` delegating to `self._transport.close()`, so it satisfies the new abstract method with no changes needed.

This enables the idiomatic pattern at the `Client` abstraction level:

```python
async with await ClientFactory.connect('https://agent.example.com') as client:
    async for event in client.send_message(msg):
        ...
    # close() called automatically, even on exceptions
```

Making `close()` abstract rather than a concrete no-op is consistent with how `Client` is designed — it already declares 7 abstract methods. Any implementation managing resources (transports, connections) should be required to handle cleanup explicitly. The SDK is pre-1.0, and the fix for any hypothetical external subclass would be a single-line `async def close(self): pass`.

- Change `ClientFactory.connect()` return type to `BaseClient`: This would expose `close()` and `async with` to factory users, but narrows the return type to a concrete class. Code with `Client` type hints (e.g. function signatures, dependency injection) would still lack access.
- Keep relying on `hasattr(client, 'close')`: This is what the integration tests currently do. It works but is fragile, bypasses static analysis, and doesn't guarantee that all implementations handle cleanup.
- Make `close()` a concrete no-op instead of abstract: This avoids forcing subclasses to implement it, but loses the guarantee that resource-holding implementations explicitly handle cleanup. Given that `Client` already requires 7 abstract methods, one more is not a meaningful burden.

### Additional context

After contributing #682 (`ClientTransport`) and #688 (`BaseClient`), I saw the opportunity to extend async context manager support to the abstract `Client` interface, making the pattern consistent across the entire client hierarchy — from transport to concrete client to the abstract interface that `ClientFactory.connect()` returns.

Adding an abstract method to `Client` is technically a breaking change for any external subclass, but given the SDK is pre-1.0 and `Client` already requires 7 abstract methods, the impact is minimal (a single-line `async def close(self): pass` fix). Existing code that calls `close()` manually or uses `try/finally` continues to work as before.

If there's interest, `EventQueue` on the server side could also benefit from the same pattern — it already has a sophisticated `close()` with graceful/immediate modes, child propagation, and cross-version handling, but doesn't support `async with`. Happy to open a separate issue and submit PRs for both.

### Code of Conduct

cchinchilla-dev:
# Description

Fixes #689 🦕

### Problem

The abstract `Client` class (`src/a2a/client/client.py`) — the public interface returned by `ClientFactory.connect()` — does not declare `close()`, `__aenter__`, or `__aexit__`. Code typed against `Client` cannot use the async context manager pattern or call `close()`:

```python
client = await ClientFactory.connect(card)
# client is typed as Client, not BaseClient

async with client:  # TypeError
    ...

await client.close()  # AttributeError
```

`ClientTransport` already supports this protocol since #682, and `BaseClient` since #688, but the gap at the abstract interface level means the pattern is unavailable to consumers that depend on the `Client` type.

The integration tests reflect this — they rely on `hasattr` checks instead of the type system:

```python
if hasattr(transport, 'close'):
    await transport.close()
```

### Fix

- Add `close()` as an abstract method on `Client`, consistent with the existing 7 abstract methods.
- Add `__aenter__` and `__aexit__` as concrete methods on `Client`, delegating to `close()`. `__aenter__` returns `Self` (via `typing_extensions`), matching the convention established in `ClientTransport` (#682).

`BaseClient` already implements `close()` (delegating to `self._transport.close()`), so it satisfies the new abstract method with no additional changes.

This enables the idiomatic pattern at the `Client` abstraction level:

```python
async with await ClientFactory.connect(card) as client:
    async for event in client.send_message(msg):
        ...
    # close() called automatically, even on exceptions
```

### Tests

No new tests needed. The existing `test_base_client_async_context_manager` and `test_base_client_async_context_manager_on_exception` in `test_base_client.py` already exercise the `__aenter__`/`__aexit__`/`close()` chain through `BaseClient`, which inherits these methods from `Client`. Happy to add dedicated tests if maintainers prefer a different approach.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -92,6 +92,7 @@ openapiv2
 opensource
 otherurl
 pb2
+poolclass
 postgres
 POSTGRES
 postgresql
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ postgresql = ["sqlalchemy[asyncio,postgresql-asyncpg]>=2.0.0"]
 mysql = ["sqlalchemy[asyncio,aiomysql]>=2.0.0"]
 signing = ["PyJWT>=2.0.0"]
 sqlite = ["sqlalchemy[asyncio,aiosqlite]>=2.0.0"]
+db-cli = ["alembic>=1.14.0"]
 
 sql = ["a2a-sdk[postgresql,mysql,sqlite]"]
 
@@ -49,6 +50,7 @@ all = [
   "a2a-sdk[grpc]",
   "a2a-sdk[telemetry]",
   "a2a-sdk[signing]",
+  "a2a-sdk[db-cli]",
 ]
 
 [project.urls]
@@ -347,3 +349,17 @@ docstring-code-format = true
 docstring-code-line-length = "dynamic"
 quote-style = "single"
 indent-style = "space"
+
+
+[tool.alembic]
+
+# path to migration scripts.
+script_location = "src/a2a/migrations"
+
+# additional paths to be prepended to sys.path. defaults to the current working directory.
+prepend_sys_path = [
+    "src"
+]
+
+[project.scripts]
+a2a-db = "a2a.a2a_db_cli:run_migrations"
--- a/src/a2a/a2a_db_cli.py
+++ b/src/a2a/a2a_db_cli.py
@@ -1,0 +1,156 @@
+import argparse
+import logging
+import os
+
+from importlib.resources import files
+
+
+try:
+    from alembic import command
+    from alembic.config import Config
+
+except ImportError as e:
+    raise ImportError(
+        "CLI requires Alembic. Install with: 'pip install a2a-sdk[db-cli]'."
+    ) from e
+
+
+def _add_shared_args(
+    parser: argparse.ArgumentParser, is_sub: bool = False
+) -> None:
+    """Add common arguments to the given parser."""
+    prefix = 'sub_' if is_sub else ''
+    parser.add_argument(
+        '--database-url',
+        dest=f'{prefix}database_url',
+        help='Database URL to use for the migrations. If not set, the DATABASE_URL environment variable will be used.',
+    )
+    parser.add_argument(
+        '--tasks-table',
+        dest=f'{prefix}tasks_table',
+        help='Custom tasks table to update. If not set, the default is "tasks".',
+    )
+    parser.add_argument(
+        '--push-notification-configs-table',
+        dest=f'{prefix}push_notification_configs_table',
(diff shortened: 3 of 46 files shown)


holtskinner:
Thanks for the contribution, what do you think about reverting the changes to `BaseClient` in #688 ? Not sure they are needed anymore

cchinchilla-dev:
> Thanks for the contribution, what do you think about reverting the changes to `BaseClient` in #688 ? Not sure they are needed anymore

You're right — with #719 adding `__aenter__`/`__aexit__` to `Client`, the overrides on `BaseClient` are redundant since it inherits them. Happy to revert those changes. Should I push the revert to this PR or open a separate one?

cchinchilla-dev:
> Thank you @cchinchilla-dev!
> 
> I believe we should target it towards `1.0-dev` branch. Adding a new abstract method is technically a breaking change for other implementation.
> 
> While I acknowledge that the chance of non-SDK implementations of this class is low, given that 1.0 release is in progress already, I'd prefer to execute this change there.

Thanks @ishymko — absolutely, that makes sense. The issue description already noted this is technically a breaking change, so targeting `1.0-dev` alongside the `v1.0` work is the right call. Want me to rebase onto `1.0-dev`? If so, I can also include the `BaseClient` revert that @holtskinner mentioned, or keep that as a separate PR — whichever you prefer.

ishymko:
@cchinchilla-dev you can combine both changes and rebase the PR onto `1.0-dev`.

PR #732, opened 2026-02-23
refactor(client): remove URL resolution logic from transports

### LINKED ISSUE #703: [Task]: Refine transport selection logic based on the new data model
1. REST transport [tries to use JSON-RPC transport interface](https://github.com/a2aproject/a2a-python/blob/b306e442b35787dcf88fc28a0fd845c07a3703e3/src/a2a/client/transports/rest.py#L64-L69).
2. JSON-RPC transport [uses the first supported interface](https://github.com/a2aproject/a2a-python/blob/b306e442b35787dcf88fc28a0fd845c07a3703e3/src/a2a/client/transports/jsonrpc.py#L61-L67), while there is no requirement of this [in the spec](https://a2a-protocol.org/latest/specification/#831-supported-interfaces-declaration), first one is the default but it can be for other protocol.

ishymko:
Rely on the `ClientFactory` to resolve proper URL and do not duplicate logic in transports. `AgentCard` is still passed to transports as it's used for capabilities inspection.

Make both `agent_card` and `url` mandatory, transports are mainly used from the `ClientFactory` and `| None` are likely non-breaking leftovers.

Fixes #703

### DIFF
--- a/src/a2a/client/transports/jsonrpc.py
+++ b/src/a2a/client/transports/jsonrpc.py
@@ -11,7 +11,6 @@ from google.protobuf import json_format
 from httpx_sse import SSEError, aconnect_sse
 from jsonrpc.jsonrpc2 import JSONRPC20Request, JSONRPC20Response
 
-from a2a.client.card_resolver import A2ACardResolver
 from a2a.client.errors import (
     A2AClientHTTPError,
     A2AClientJSONError,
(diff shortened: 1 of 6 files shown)


src/a2a/client/transports/rest.py
```diff
@@ -50,37 +46,19 @@ class RestTransport(ClientTransport):
...
-                    f'AgentCard does not support {TransportProtocol.HTTP_JSON} '
-                    f'or {TransportProtocol.JSONRPC}'
-                )
-        else:
-            raise ValueError('Must provide either agent_card or url')
-        if self.url.endswith('/'):
-            self.url = self.url[:-1]
+        if url.endswith('/'):
+            url = url[:-1]
+        self.url = url
```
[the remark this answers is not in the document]
ishymko: or `.removesuffix('/')`

src/a2a/client/transports/jsonrpc.py
```diff
@@ -50,31 +49,18 @@ class JsonRpcTransport(ClientTransport):
     def __init__(
         self,
         httpx_client: httpx.AsyncClient,
-        agent_card: AgentCard | None = None,
-        url: str | None = None,
+        url: str,
```
holtskinner: Nit: is there a reason to change this order? I understand most users should be using `ClientFactory` instead.

ishymko: Indeed not that important, it just seemed to be more logical to have `url` first as it's used for connecting. Will revert for now, will think about it separately.

PR #738, opened 2026-02-25
test: improve test_end_to_end.py

ishymko:
# Description

- Test artifacts.
- Add more assertions for streaming: validate all events.
- Fix non-streaming tests which were actually streaming.

### DIFF
--- a/tests/integration/test_client_server_integration.py
+++ b/tests/integration/test_client_server_integration.py
@@ -1,52 +1,54 @@
 import asyncio
-
 from collections.abc import AsyncGenerator
-from typing import Any, NamedTuple
-from unittest.mock import AsyncMock, patch
+from typing import NamedTuple, Any
+from unittest.mock import ANY, AsyncMock, patch
 
 import grpc
 import httpx
 import pytest
 import pytest_asyncio
-
-from cryptography.hazmat.primitives.asymmetric import ec
 from google.protobuf.json_format import MessageToDict
+from grpc.aio import Channel
 
+from jwt.api_jwk import PyJWK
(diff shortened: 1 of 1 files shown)


tests/integration/test_end_to_end.py
```diff
@@ -211,16 +214,19 @@ async def test_end_to_end_send_message_blocking(transport_setups):
...
-    assert task
-    assert task.id
-    assert task.status.state == TaskState.TASK_STATE_COMPLETED
+    assert len(events) == 1
+    response, _ = events[0]
+    assert response.task.id
+    assert response.task.status.state == TaskState.TASK_STATE_COMPLETED
+    assert len(response.task.artifacts) == 1
+    assert response.task.artifacts[0].name == 'test-artifact'
+    assert response.task.artifacts[0].parts[0].text == 'artifact content'
```
sokoliva, line 223: Nit: consider performing the lookup once and then reuse the result to not waste CPU cycles:
```
assert len(artifacts := response.task.artifacts) == 1
assert artifacts[0].name == 'test-artifact'
assert artifacts[0].parts[0].text == 'artifact content'
```

[ToT](https://engdoc.corp.google.com/eng/doc/tott/episodes/770.md?cl=head)

ishymko: I believe this operator can be used here, but usually it's done for maps rather than just regular fields (unless they are expensive to compute which is not the case here as it's just a proto generated type).

I'd keep it as is unless you insist.

sokoliva: Makes sense. Approved.

PR #744, opened 2026-02-27
fix: use default_factory for mutable field defaults in ServerCallContext

greysonlalonde:
## summary
- replace `default={}` and `default=UnauthenticatedUser()` with `default_factory=dict` and `default_factory=UnauthenticatedUser` in `ServerCallContext` to prevent shared mutable default instances across all instances.

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler.py
+++ b/src/a2a/server/request_handlers/default_request_handler.py
@@ -322,11 +322,16 @@ class DefaultRequestHandler(RequestHandler):
             (
                 result,
                 interrupted_or_non_blocking,
+                bg_consume_task,
             ) = await result_aggregator.consume_and_break_on_interrupt(
                 consumer,
                 blocking=blocking,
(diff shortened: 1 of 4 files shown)


greysonlalonde:
@ishymko bumping

src/a2a/server/context.py
```diff
@@ -19,7 +19,7 @@ class ServerCallContext(BaseModel):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    state: State = Field(default={})
-    user: User = Field(default=UnauthenticatedUser())
+    state: State = Field(default_factory=dict)
+    user: User = Field(default_factory=UnauthenticatedUser)
```
ishymko, line 23: I believe it shouldn't matter for `UnauthenticatedUser` as it's readonly.

greysonlalonde: Agreed, more of a pedantic/consistency change for `user`.

PR #752, opened 2026-03-02
feat: Implement a vertex based task store

### LINKED ISSUE #751: [Feat]: add support for the vertex managed task store service

_No response_

It will be similar to src/a2a/server/tasks/database_task_store.py but use the vertex backend instead.

_No response_

### Additional context

_No response_

### Code of Conduct

gaborfeher:
Create a task store adapter that uses the Vertex Managed Task Store as a backend.

Fixes #751

BEGIN_COMMIT_OVERRIDE
feat: Implement a vertex based task store (#752)

Release-As: 0.3.25
END_COMMIT_OVERRIDE

### DIFF
--- a/tests/contrib/tasks/test_vertex_task_store.py
+++ b/tests/contrib/tasks/test_vertex_task_store.py
@@ -1,13 +1,14 @@
 """
 Tests for the VertexTaskStore.
 
-These tests interact with the Vertex AI Agent Engine. They are skipped by default
-unless the necessary environment variables are set, which prevents them from failing in GitHub Actions.
+These tests can be run with a real or fake Vertex AI Agent Engine as a backend.
+The real ones are skipped by default unless the necessary environment variables\
+are set, which prevents them from failing in GitHub Actions.
 
 To run these tests locally, you can use the provided script:
     ./run_vertex_tests.sh
 
-Be sure to set the following environment variables:
+The following environment variables are required for the real backend:
     VERTEX_PROJECT="your-project" \
     VERTEX_LOCATION="your-location" \
     VERTEX_BASE_URL="your-base-url" \


pyproject.toml
```diff
@@ -62,6 +64,9 @@ build-backend = "hatchling.build"
 [tool.hatch.version]
 source = "uv-dynamic-versioning"
 
+[tool.hatch.metadata]
+allow-direct-references = true
+
```
ishymko: Is it still required? I assume it was for referencing a non-released version of Google Cloud SDK?

src/a2a/contrib/tasks/vertex_task_converter.py
```diff
@@ -0,0 +1,161 @@
...
+        TaskState.unknown: types.State.STATE_UNSPECIFIED,
+        TaskState.submitted: types.State.SUBMITTED,
+        TaskState.working: types.State.WORKING,
+        TaskState.completed: types.State.COMPLETED,
+        TaskState.canceled: types.State.CANCELLED,
+        TaskState.failed: types.State.FAILED,
+        TaskState.rejected: types.State.REJECTED,
+        TaskState.input_required: types.State.INPUT_REQUIRED,
+        TaskState.auth_required: types.State.AUTH_REQUIRED,
+    }.get(task_state, types.State.STATE_UNSPECIFIED)
```
ishymko: nit: maybe two dictionaries as module variables, one computed from another?

```python
_STORED_TO_SDK_STATE = {
    types.State.STATE_UNSPECIFIED: TaskState.unknown,
    types.State.SUBMITTED: TaskState.submitted,
    types.State.WORKING: TaskState.working,
    types.State.COMPLETED: TaskState.completed,
    types.State.CANCELLED: TaskState.canceled,
    types.State.FAILED: TaskState.failed,
    types.State.REJECTED: TaskState.rejected,
    types.State.INPUT_REQUIRED: TaskState.input_required,
    types.State.AUTH_REQUIRED: TaskState.auth_required,
}

_SDK_TO_STORED_STATE = {v: k for k, v in _STORED_TO_SDK_STATE.items()}
```

src/a2a/contrib/tasks/vertex_task_converter.py
```diff
@@ -0,0 +1,161 @@
+try:
+    from vertexai import types
```
ishymko: super nit: I'd import vertexai types under some alias as module name matches `a2a.types`, so it's not immediately clear when reading the code which one uses `types.Xxx` (vertexai) vs just `Xxx` (a2a.types).

uv.lock
```diff
@@ -2,8 +2,10 @@ version = 1
 revision = 3
 requires-python = ">=3.10"
 resolution-markers = [
-    "python_full_version >= '3.13'",
-    "python_full_version < '3.13'",
+    "python_full_version >= '3.14'",
+    "python_full_version == '3.13.*'",
+    "python_full_version >= '3.11' and python_full_version < '3.13'",
+    "python_full_version < '3.11'",
```
ishymko, line 8: Was it auto-updated after installing vertexai?

src/a2a/contrib/tasks/vertex_task_converter.py
```diff
@@ -0,0 +1,161 @@
...
+    """Converts a SDK Part to a proto Part."""
+    if isinstance(part.root, TextPart):
+        return types.Part(text=part.root.text)
+    if isinstance(part.root, DataPart):
+        data_bytes = json.dumps(part.root.data).encode('utf-8')
+        return types.Part(
+            inline_data=types.Blob(
+                mime_type='application/json', data=data_bytes
+            )
+        )
```
ishymko, line 64: I believe this isn't going to roundtrip back to the A2A DataPart? Maybe `application/json` MIME type should be used as a signal when converting back (one can have a `FileWithBytes` representing the same though but it feels that this use case is going to be less popular)?

src/a2a/contrib/tasks/vertex_task_converter.py
```diff
@@ -0,0 +1,161 @@
...
+        return Part(
+            root=FilePart(
+                file=FileWithUri(
+                    mime_type=stored_part.file_data.mime_type,
+                    uri=stored_part.file_data.file_uri,
+                )
+            )
+        )
+
+    return Part(root=TextPart(text=''))
```
ishymko: Is is by design that the method for other direction throws on unknown and this one defaults to an empty text?

tests/contrib/tasks/test_vertex_task_converter.py
ishymko, line 1: Maybe some roundtrip tests can be added here (or existing replaced with roundtrip)?

I see that `test_vertex_task_store.py` does save-read tests but as it requires real Vertex it's not going to be executed in CI.

tests/contrib/tasks/test_vertex_task_store.py
```diff
@@ -0,0 +1,483 @@
+"""
+Tests for the VertexTaskStore.
+
+These tests interact with the Vertex AI Agent Engine. They are skipped by default
+unless the necessary environment variables are set, which prevents them from failing in GitHub Actions.
+
+To run these tests locally, you can use the provided script:
+    ./scripts/run_vertex_tests.sh
```
ishymko: nit: outdated path? The script is not under `scripts` anymore.

tests/contrib/tasks/test_vertex_task_store.py
ishymko, line 1: Should we have a dedicated test suite which mocks external dependencies given that this one can't be run in the CI? While I agree that having e2e tests is important if we can't run them without our own project it's problematic.

I am wondering what are best practices for testing against Cloud SDK without paying for the actual infra.

pyproject.toml
gaborfeher: Good catch, removed.

uv.lock
gaborfeher: Yes. I tried replaying my original steps and adding the vertexai dependency triggers this change. I mean running `uv lock` after adding the dep in `pyproject.toml`

src/a2a/contrib/tasks/vertex_task_converter.py
gaborfeher: Yes, that's correct. So are you saying that it's better to turn `FileWithBytes` of type `application/json` into `DataPart` than vice versa?

tests/contrib/tasks/test_vertex_task_store.py
gaborfeher: I added an in-memory based fake backend, I don't have any better ideas, but this might be good enough.

src/a2a/contrib/tasks/vertex_task_converter.py
ishymko: It will be lossy either way I believe (unless we use a custom MIME type defined specifically here but I am not sure how well it will work with the Task Store backend) but for me it feels like this approach will cover more common use-case and auto-conversion `FileWithBytes` of `application/json` to `DataPart` will rarely occur in practice.

That being said we can adjust it separately as it's only on read. Resolving.

tests/contrib/tasks/test_vertex_task_store.py
ishymko: Nice, I like it!

Issue #742, opened 2026-02-26
[Feat]: 0.3 compatibility


ishymko:
Umbrella issue for defining compatibility between 1.0 SDK with 0.3 spec (note: it's not about Python SDK API compatibility, but spec compatibility for requests/responses payloads and behavior).

See https://github.com/a2aproject/A2A/discussions/1416 for more context.

### Summary
1. 1.0 server to serve agent card which contains union of 1.0 and 0.3 fields - based on the analysis in the discussion above 1.0 agent card was adjusted to make it possible.
2. 1.0 client to consume 0.3 agent cards.
3. 1.0 server to expose 0.3 interface (per transport). Can be done via a separate path for HTTP or via `A2A-Version` header. Will require generating two sets of types for gRPC, given that package name of `a2a.proto` was updated for 1.0 spec to avoid collision (https://github.com/a2aproject/A2A/pull/1474) both services can be exported on the same port.
4. 1.0 client to consume 0.3 interface (per transport). [`AgentInterface`](https://a2a-protocol.org/latest/specification/#446-agentinterface) now contains `protocolVersion`. Old agent card defines transports URLs in different fields, so client will be able to distinguish in non-ambiguous way.

### Non-goals
1. Dynamically selecting push notifications payload: [0.3 sends `Task`](https://a2a-protocol.org/v0.3.0/specification/#95-push-notification-setup-and-usage) and [1.0 sends same events as streaming](https://a2a-protocol.org/latest/specification/#433-push-notification-payload).
2. Sending or not sending `Task` as a first event of `resubscribe` method - requirement added in 1.0 ([3.1.6. Subscribe to Task](https://a2a-protocol.org/latest/specification/#316-subscribe-to-task))

### Implementation
Given that data model changes made in the 1.0 spec mostly reshuffled existing fields the idea is to rely on transport-level mapping logic for compatibility while reusing existing domain code. a2a-go implementation: https://github.com/a2aproject/a2a-go/pull/225.

#### Client
Client already has [`transport` concept](https://github.com/a2aproject/a2a-python/blob/1.0-dev/src/a2a/client/transports/base.py) which can contain mapping logic via a dedicated transport or a wrapper one. `ClientFactory` will require updating transport selection logic to take protocol versions and agent card differences into account to properly select transport: [client_factory.py](https://github.com/a2aproject/a2a-python/blob/1.0-dev/src/a2a/client/client_factory.py#L183).

yuyu123-yuyu:
[...]

Issue #706, opened 2026-02-18
[Task]: Research removing buf from build process


ishymko:
Currently [`buf`](https://github.com/a2aproject/a2a-python/blob/1.0-dev/scripts/gen_proto.sh) is executed [on each build](https://github.com/a2aproject/a2a-python/blob/b306e442b35787dcf88fc28a0fd845c07a3703e3/pyproject.toml#L67-L69).

Given that the spec isn't going to be updated often after the release 99% of runs will be no-op while it requires more dependencies to install for initial `uv sync`/`uv run pytest`. We should consider extracting it into a separate step which is going to be run manually when needed (when new spec update is coming).

In addition see https://github.com/a2aproject/a2a-python/pull/754#discussion_r2877534713 - currently one needs to run `uv build` in a cloned repo to generate 0.3 files as they are ignored. Considerations above apply to these files as well.

**BUT:** seems like this process is also required currently for `a2a.json` [here](https://github.com/a2aproject/a2a-python/blob/b306e442b35787dcf88fc28a0fd845c07a3703e3/src/a2a/server/apps/jsonrpc/fastapi_app.py#L56). So maybe this file should be committed to the repo same way as the generated code to avoid regenerating each time (now it's in [`.gitignore`](https://github.com/a2aproject/a2a-python/blob/1.0-dev/.gitignore)). It's unclear if it works/matches `main` branch.

Given the above **but** proposed plan is:

1) Remove `buf` from build process and commit all required files.
2) Deal with **but**.

PR #754, opened 2026-03-03
feat(compat): legacy v0.3 protocol models, conversion logic and utilities

bartek-w:
* Isolate legacy v0.3 Pydantic models in `src/a2a/compat/v0_3/types.py`.
* Add dynamic generation script `scripts/gen_proto.sh` for pulling legacy v0.3 Protobuf specifications directly from GitHub, ensuring no `DescriptorPool` namespace collisions.
* Introduce strict conversion boundaries (`proto_utils.py` and `conversions.py`) to elegantly translate between legacy byte-formats, intermediate Pydantic models, and the modern v1.0 SDK architecture.
* Add comprehensive round-trip tests to guarantee zero data loss during conversion bridging.

Re #742


src/a2a/compat/v0_3/proto_utils.py
```diff
@@ -0,0 +1,1103 @@
...
+    @classmethod
+    def file(
+        cls, file: types.FileWithUri | types.FileWithBytes
+    ) -> a2a_pb2.FilePart:
+        if isinstance(file, types.FileWithUri):
+            return a2a_pb2.FilePart(
+                file_with_uri=file.uri, mime_type=file.mime_type, name=file.name
+            )
+        return a2a_pb2.FilePart(
+            file_with_bytes=file.bytes.encode('utf-8'),
```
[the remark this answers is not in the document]
bartek-w: src/a2a/compat/v0_3/proto_utils.py is copied from 0.3
This is legacy code copied as is.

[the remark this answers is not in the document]
bartek-w: src/a2a/compat/v0_3/proto_utils.py is copied from 0.3
This is legacy code copied as is.

.github/actions/spelling/expect.txt
ishymko, line 1: nit: Not sure maybe we used [`allow.txt`](https://github.com/a2aproject/a2a-python/blob/1.0-dev/.github/actions/spelling/allow.txt) wrong but this is where all unknown words are put right now, let's put this one there as well and investigate later. 

I see that [CI](https://github.com/a2aproject/a2a-python/actions/runs/22614325557?pr=754) offers this file so maybe our current way is wrong

Image

scripts/gen_proto.sh
```diff
@@ -19,3 +19,15 @@ fi
 # Fix imports in generated grpc file
 echo "Fixing imports in src/a2a/types/a2a_pb2_grpc.py"
 sed 's/import a2a_pb2 as a2a__pb2/from . import a2a_pb2 as a2a__pb2/g' src/a2a/types/a2a_pb2_grpc.py > src/a2a/types/a2a_pb2_grpc.py.tmp && mv src/a2a/types/a2a_pb2_grpc.py.tmp src/a2a/types/a2a_pb2_grpc.py
+
+# Download legacy v0.3 compatibility protobuf code
+echo "Downloading legacy v0.3 proto file..."
+curl -o src/a2a/compat/v0_3/a2a_v0_3.proto https://raw.githubusercontent.com/a2aproject/A2A/a8b45dcc429a5571ef8a24c36336bf84b89bbd7f/specification/grpc/a2a.proto
```
ishymko: nit: Can you add a comment how `a8b45dcc429a5571ef8a24c36336bf84b89bbd7f` commit selected?

Seems like there is/was no good way of tracking the source of generated types. I assume it was done based on [`a2a_pb2.py`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/grpc/a2a_pb2.py) history and #455 was the last PR regenerating it. The description there mentions [`a8b45dcc429a5571ef8a24c36336bf84b89bbd7f`](https://github.com/a2aproject/A2A/commit/a8b45dcc429a5571ef8a24c36336bf84b89bbd7f) and [`b3b266d127dde3d1000ec103b252d1de81289e83`](https://github.com/a2aproject/A2A/commit/b3b266d127dde3d1000ec103b252d1de81289e83), although `b3b266d127dde3d1000ec103b252d1de81289e83` is a newer one it only updates comment, so `a8b45` was used?

src/a2a/compat/v0_3/.gitignore
ishymko, line 1: I'd say we don't need to ignore those files. Currently `gen_proto.sh` is executed on each build which I'd like to avoid, as in fact data model doesn't change that often (#706). Same code for 1.0 data model is committed and other SDKs do the same.

We can put them in a separate PR as this one is large enough already.

src/a2a/compat/v0_3/conversions.py
```diff
@@ -0,0 +1,1336 @@
...
+        state_str = compat_status.state.value.upper()
+        if state_str == 'UNKNOWN':
+            core_status.state = pb2_v10.TaskState.TASK_STATE_UNSPECIFIED
+        else:
+            # Handle 'input-required', 'auth-required'
+            state_str = state_str.replace('-', '_')
+
+            core_status.state = cast(
+                'Any', pb2_v10.TaskState.Value(f'TASK_STATE_{state_str}')
+            )
```
ishymko: nit: maybe an explicit map between all states can be used here for explicitness? (same for the opposite direction in `to_compat_task_status`)

src/a2a/compat/v0_3/conversions.py
```diff
@@ -0,0 +1,1336 @@
...
+    return types_v03.TaskStatusUpdateEvent(
+        task_id=core_event.task_id,
+        context_id=core_event.context_id,
+        status=to_compat_task_status(core_event.status)
+        if core_event.HasField('status')
+        else types_v03.TaskStatus(state=types_v03.TaskState.unknown),
+        metadata=MessageToDict(core_event.metadata)
+        if core_event.HasField('metadata')
+        else None,
+        final=False,  # 'final' was removed in v1.0
```
ishymko: Let's set it to "state is a terminal state": [6.1. Task Object (0.3 spec)](https://a2a-protocol.org/v0.3.0/specification/#61-task-object):

> A task which has reached a terminal state (completed, canceled, rejected, or failed) can't be restarted

(a2a-go also does it: [`conversions.go#L230`](https://github.com/a2aproject/a2a-go/blob/b964514821857e315bd127808f757034a40a8287/a2acompat/a2av0/conversions.go#L230))

src/a2a/compat/v0_3/conversions.py
```diff
@@ -0,0 +1,1336 @@
...
+    )
+
+    if which == 'text':
+        return types_v03.Part(
+            root=types_v03.TextPart(text=core_part.text, metadata=metadata)
+        )
+
+    if which == 'data':
+        # core_part.data is a google.protobuf.Value. It can be converted to dict.
+        data_dict = MessageToDict(core_part.data)
```
ishymko, line 89: Hmm, based on my testing `MessageToDict` can return primitives for `google.protobuf.Value`:

```python
from google.protobuf import json_format
from google.protobuf.struct_pb2 import Value, Struct, ListValue, NullValue
    
# 1. String Value
v_str = Value(string_value="Hello World")
out_str = json_format.MessageToDict(v_str)
print(f"Original: Value(string_value='Hello World')")
print(f"Result Type: {type(out_str)}")
print(f"Result Value: {out_str!r}")

# 2. Number Value
v_num = Value(number_value=42.5)
out_num = json_format.MessageToDict(v_num)
print(f"Original: Value(number_value=42.5)")
print(f"Result Type: {type(out_num)}")
print(f"Result Value: {out_num!r}")

# 3. Bool Value
v_bool = Value(bool_value=True)
out_bool = json_format.MessageToDict(v_bool)
print(f"Original: Value(bool_value=True)")
print(f"Result Type: {type(out_bool)}")
print(f"Result Value: {out_bool!r}")

# 4. Null Value
v_null = Value(null_value=NullValue.NULL_VALUE)
out_null = json_format.MessageToDict(v_null)
print(f"Original: Value(null_value=NULL_VALUE)")
print(f"Result Type: {type(out_null)}")
print(f"Result Value: {out_null!r}")

# 5. List Value
list_val = ListValue()
list_val.values.add(string_value="item1")
list_val.values.add(number_value=123)
v_list = Value(list_value=list_val)

out_list = json_format.MessageToDict(v_list)
print(f"Original: Value(list_value=[...])")
print(f"Result Type: {type(out_list)}")
print(f"Result Value: {out_list!r}")

# 6. Struct Value (Map) - The only case that returns a dict!
struct_val = Struct()
struct_val.fields["key"].string_value = "value"
v_struct = Value(struct_value=struct_val)

out_struct = json_format.MessageToDict(v_struct)
print(f"Original: Value(struct_value={{'key': 'value'}})")
print(f"Result Type: {type(out_struct)}")
print(f"Result Value: {out_struct!r}")
```

gives

```
Original: Value(string_value='Hello World')
Result Type: <class 'str'>
Result Value: 'Hello World'
Original: Value(number_value=42.5)
Result Type: <class 'float'>
Result Value: 42.5
Original: Value(bool_value=True)
Result Type: <class 'bool'>
Result Value: True
Original: Value(null_value=NULL_VALUE)
Result Type: <class 'NoneType'>
Result Value: None
Original: Value(list_value=[...])
Result Type: <class 'list'>
Result Value: ['item1', 123.0]
Original: Value(struct_value={'key': 'value'})
Result Type: <class 'dict'>
Result Value: {'key': 'value'}
```

Unless I misunderstand something it even contradicts its own type of `-> Dict[Text, Any]`...

scripts/gen_proto.sh
bartek-w: You are right. It was based on https://github.com/a2aproject/a2a-python/pull/455 but somehow I copied wrong commit. Fixed.

src/a2a/compat/v0_3/.gitignore
bartek-w: Let's keep it as part of #706

src/a2a/compat/v0_3/conversions.py
ishymko: UPD:

What I meant is that doing

```python
def test_data_part_conversion_non_struct():
    v10_part = pb2_v10.Part()
    v10_part.data.string_value = "primitive data"

    to_compat_part(v10_part)
```

throws

```
            return types_v03.Part(
>               root=types_v03.DataPart(data=data_dict, metadata=metadata)
            )
E           pydantic_core._pydantic_core.ValidationError: 1 validation error for DataPart
E           data
E             Input should be a valid dictionary [type=dict_type, input_value='primitive data', input_type=str]
E               For further information visit https://errors.pydantic.dev/2.12/v/dict_type
```

In [a2a-go](https://github.com/a2aproject/a2a-go/blob/5f7a594a85a1653ba8d41a7fc0f0924cef8f6b3d/a2acompat/a2av0/conversions.go#L172-L178) primitives are wrapped into a dict with `value` key and are marked with a metadata value. Old code won't support this metadata but with this data model change I believe it's the best what we can do.

PR #758, opened 2026-03-03
feat: handle tenant in Client

### LINKED ISSUE #672: [Task]: 1.0: Support multi-tenancy
https://github.com/a2aproject/A2A/pull/1195 updates core proto with extra HTTP bindings prefixed with `/{tenant}`. It should be supported.

It's not fully clear how to represent it for JSON-RPC, also HTTP+JSON/REST protocol binding in the spec doesn't mention it, see https://github.com/a2aproject/A2A/issues/1273 and https://a2a-protocol.org/latest/specification/#53-method-mapping-reference.

https://github.com/a2aproject/a2a-go/pull/237 can be used as an inspiration.

sokoliva:
## Changes
- Rest client transport prepends path with tenant if provided.
- added `TenantTransportDecorator` to `client/transports/base.py` which adds default tenant to requests in for provided
- updated `client_factory` to create `TenantTransportDecorator` if tenant is provided in `AgentInterface`

## Contributing

Fixes #672 🦕


src/a2a/client/transports/__init__.py
```diff
@@ -16,4 +16,5 @@
     'GrpcTransport',
     'JsonRpcTransport',
     'RestTransport',
+    'TenantTransportDecorator',
```
ishymko: Let's not export it for now, we may do it later if needed (just want to keep public API reasonable in size, each new export is a contract we'll have to maintain in non-major versions).

src/a2a/client/transports/rest.py
```diff
@@ -264,8 +272,11 @@ async def list_tasks(
             modified_kwargs,
             extensions if extensions is not None else self.extensions,
         )
+
+        path = self._get_path('/v1/tasks', request.tenant)
```
ishymko: Can we put this logic into `_send_XXX_request` methods instead and make `tenant` a mandatory parameter in all those internal methods? This way we ensure that it's not forgotten, as now skipping `path = self._get_path` step before passing path won't trigger any type errors.

src/a2a/client/transports/base.py
```diff
@@ -158,3 +158,171 @@ async def get_extended_agent_card(
...
+
+    def update_tenant(self, tenant: str) -> str:
+        """If tenant is not provided, use the default tenant.
+
+        Returns:
+            The tenant used for the request.
+        """
+        if tenant != '':
+            return tenant
+        return self._tenant or ''
```
ishymko: Consider

```suggestion
    def _resolve_tenant(self, tenant: str) -> str:
        """If tenant is not provided, use the default tenant.

        Returns:
            The tenant used for the request.
        """
        return tenant or self._tenant or ''
```

"Update" is a bit inaccurate here as it's a pure function which doesn't modify any state, also `_` added as it's internal.

Body is updated to be an equivalent one-liner which makes "priorities" more explicit.

src/a2a/client/transports/base.py
```diff
@@ -158,3 +158,171 @@ async def get_extended_agent_card(
     @abstractmethod
     async def close(self) -> None:
         """Closes the transport."""
+
+
+class TenantTransportDecorator(ClientTransport):
```
ishymko: Maybe let's move into a separate file? Current one doesn't have any implementation now, so this addition will change this invariant.

src/a2a/client/transports/base.py
sokoliva: Great point. Also, I changed it to `return tenant or self._tenant` since both `tenant` and `self._tenant` are mandatory. If `tenant==''` it will return `self._tenant` and if both of them are empty strings it will return an empty string (confirmed in tests `test_resolve_tenant_logic` and `test_resolve_tenant_logic_empty_tenant` in `tests/client/transports/test_tenant_decorator.py`).

src/a2a/client/transports/rest.py
sokoliva: Done. I will revisit `get_extended_agent_card` once [767](https://github.com/a2aproject/a2a-python/pull/767) is merged.

sokoliva: I removed `| None`, but should empty strings be allowed? They are currently allowed. @ishymko 

Edit: in [A2A Protocol Operations](https://a2a-protocol.org/latest/specification/#3-a2a-protocol-operations) tenant is defined as optional, hence I defined `_get_path` as

```
 def _get_path(self, base_path: str, tenant: str) -> str:
        return f'/{tenant}{base_path}' if tenant else base_path
```

 Please confirm.

ishymko (approved):

PR #760, opened 2026-03-04
feat(compat): AgentCard backward compatibility helpers and tests

bartek-w:
This commit implements the backwards compatibility helpers for exchanging legacy v0.3 Agent Cards across the v1.0 protocol bounds.

Also resolves linting and validation constraints.

# Description

### DIFF
--- a/src/a2a/client/helpers.py
+++ b/src/a2a/client/helpers.py
@@ -1,8 +1,116 @@
 """Helper functions for the A2A client."""
 
+from typing import Any
 from uuid import uuid4
 
-from a2a.types.a2a_pb2 import Message, Part, Role
+from google.protobuf.json_format import ParseDict
+
+from a2a.types.a2a_pb2 import AgentCard, Message, Part, Role
+
+
+def parse_agent_card(agent_card_data: dict[str, Any]) -> AgentCard:
+    """Parse AgentCard JSON dictionary and handle backward compatibility."""
+    _handle_extended_card_compatibility(agent_card_data)
+    _handle_connection_fields_compatibility(agent_card_data)
+    _handle_security_compatibility(agent_card_data)
+
+    return ParseDict(agent_card_data, AgentCard(), ignore_unknown_fields=True)
+
+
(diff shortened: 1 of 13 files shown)


src/a2a/client/helpers.py
ishymko, line 1: To the best of my knowledge `snake_case` shouldn't be required here, both 1.0 and 0.3 use `camelCase` for field names.

src/a2a/client/helpers.py
```diff
@@ -1,8 +1,137 @@
...
+def _map_legacy_security(
+    sec_list: list[dict[str, list[str]]],
+) -> list[dict[str, Any]]:
+    """Convert a legacy security requirement list into the 1.0.0 Protobuf format."""
+    return [
+        {
+            'schemes': {
+                scheme_name: {'list': scopes}
+                for scheme_name, scopes in sec_dict.items()
+            }
```
ishymko, line 70: What's the source of `schemes` key here?

Note that in [0.3](https://a2a-protocol.org/v0.3.0/specification/#json-rpc) the actual data model used is JSON Schema based, not proto based (and have some differences).

src/a2a/server/apps/jsonrpc/jsonrpc_app.py
ishymko, line 518: Given that compat mode was applied in REST, I believe it should be applied here as well. It happens after version negotiation though, so it doesn't have to be a "union", can be handled in follow-up PR(s) for other conversions.

tests/integration/cross_version/validate_agent_cards_030.py
ishymko, line 1: nit: if it meant to be committed maybe some short readme on how to use it? We can skip if in the end (after compat work) it's going to be removed completely and fully replaced with automated tests.

bartek-w: I've added a comment in the file to explain how it is used.

ishymko (approved):

PR #764, opened 2026-03-04
feat(compat): set a2a-version header to 1.0.0

bartek-w:
# Description


src/a2a/utils/constants.py
```diff
@@ -22,3 +22,7 @@ class TransportProtocol(str, Enum):
 
 DEFAULT_MAX_CONTENT_LENGTH = 10 * 1024 * 1024  # 10MB
 JSONRPC_PARSE_ERROR_CODE = -32700
+VERSION_HEADER = 'X-A2A-Version'
```
ishymko: Should it be
```suggestion
VERSION_HEADER = 'A2A-Version'
```

There is no `X-` in 1.0: [14.2.1. A2A-Version Header](https://a2a-protocol.org/latest/specification/#1421-a2a-version-header)

src/a2a/client/client_factory.py
```diff
@@ -65,6 +67,12 @@ def __init__(
     ):
         if consumers is None:
             consumers = []
+
+        if config.httpx_client is not None:
+            config.httpx_client.headers.setdefault(
+                VERSION_HEADER, PROTOCOL_VERSION_CURRENT
+            )
```
ishymko: nit: maybe something like


```suggestion
        client = config.httpx_client or httpx.AsyncClient()
        client.headers.setdefault(
            VERSION_HEADER, PROTOCOL_VERSION_CURRENT
        )
```

so that below we can just do

```diff
-config.httpx_client or self._default_httpx_client(),
+client
```

src/a2a/client/client_factory.py
```diff
@@ -155,7 +169,7 @@ async def connect(  # noqa: PLR0913
         client_config = client_config or ClientConfig()
         if isinstance(agent, str):
             if not client_config.httpx_client:
-                async with httpx.AsyncClient() as client:
+                async with cls._default_httpx_client() as client:
```
ishymko: Should be safe to use the normal one, agent card fetching is a bit different compared to the actual RPC interaction and doesn't rely on the version header.

(I also noticed that this code feels strange in the first place with `else` branch holding similar code, I will try to refactor it separately)

src/a2a/utils/constants.py
bartek-w: Good catch. It looks like I've missed that x- prefixes were deprecated like 15 years ago :D

PR #775, opened 2026-03-06
fix: return background task from consume_and_break_on_interrupt to prevent GC

### LINKED ISSUE #774: [Bug]: ResultAggregator._continue_consuming background task is not referenced and may be garbage collected
### What happened?

In `ResultAggregator.consume_and_break_on_interrupt()`, when a non-blocking `message/send` request interrupts early (e.g., `blocking=False`), the remaining event consumption is delegated to a background task via `asyncio.create_task()`. However, the returned Task object is not saved, making it eligible for garbage collection.
```
# src/a2a/server/tasks/result_aggregator.py, line 162
asyncio.create_task(  # noqa: RUF006
    self._continue_consuming(event_stream, event_callback)
)
```
Per the [Python docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may get garbage collected at any time, even before it's done.

When this task is garbage collected:

1. The event_stream async generator is finalized
2. Remaining events (artifacts, completed/failed status updates) are never processed
3. The `event_callback` (push notification sender) never fires for the final task state
4. The task remains stuck in working state permanently from the client's perspective

Steps to Reproduce

1. Configure an A2A server with push notifications enabled
2. Send a `message/send` request with MessageSendConfiguration(blocking=False, push_notification_config=...)
3. Have the agent executor take a non-trivial amount of time (>10 seconds) to complete
4. Observe that the task is acknowledged with working status but never transitions to completed
5. The push notification webhook for the final state is never sent

The longer the agent takes, the more likely the GC collects the task. For agents running 60+ seconds, this is essentially 100% reproducible.

Expected Behavior

The background _continue_consuming task should complete, processing all remaining events and firing the push notification callback for each, including the terminal state.

Root Cause

The noqa: RUF006 comment on line 162 suppresses Ruff's [asyncio-dangling-task](https://docs.astral.sh/ruff/rules/asyncio-dangling-task/) rule, which specifically warns about this pattern. The TODO on line 161 also acknowledges the issue:
```
# TODO: We should track all outstanding tasks to ensure they eventually complete.
```
DefaultRequestHandler already has a `_background_tasks: set[asyncio.Task]` and `_track_background_task()` method for exactly this purpose, but ResultAggregator doesn't use it.

Suggested Fix

Return the background task from consume_and_break_on_interrupt() so the caller can track it, or accept a task-tracking callback. For example:
```
# Return the background task in the tuple
async def consume_and_break_on_interrupt(
    self, consumer, blocking=True, event_callback=None
) -> tuple[Task | Message | None, bool, asyncio.Task | None]:
    ...
    bg_task = None
    if should_interrupt:
        bg_task = asyncio.create_task(
            self._continue_consuming(event_stream, event_callback)
        )
        interrupted = True
        break
    return await self.task_manager.get_task(), interrupted, bg_task
```
Then in DefaultRequestHandler.on_message_send:

```
result, interrupted_or_non_blocking, bg_task = await result_aggregator.consume_and_break_on_interrupt(...)
if bg_task:
    bg_task.set_name(f'continue_consuming:{task_id}')
    self._track_background_task(bg_task)
```

Environment

- a2a-sdk version: 0.3.22 (also present on latest main)
- Python: 3.12.12
- OS: macOS (ARM64)

Related

- PR #394 introduced the event_callback mechanism but also introduced this untracked task
- Issue #239 (original non-blocking push notification bug)

### Relevant log output

```shell

```

### Code of Conduct

kevinlu310:
# Description

ResultAggregator.consume_and_break_on_interrupt creates a background asyncio.Task to continue consuming events after an interruption (non-blocking or auth_required), but discards the task reference. On Python 3.12+ the event loop only holds weak references to tasks, so the garbage collector can silently collect the task before it completes — dropping remaining events (completed/failed status) and push notification callbacks.

Return the background task as a third tuple element so callers can hold a strong reference. DefaultRequestHandler.on_message_send now tracks it via _track_background_task(), the same mechanism already used for other background work.

Fixes #774 🦕

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ postgresql = ["sqlalchemy[asyncio,postgresql-asyncpg]>=2.0.0"]
 mysql = ["sqlalchemy[asyncio,aiomysql]>=2.0.0"]
 signing = ["PyJWT>=2.0.0"]
 sqlite = ["sqlalchemy[asyncio,aiosqlite]>=2.0.0"]
+vertex = ["google-cloud-aiplatform>=1.140.0"]
 
 sql = ["a2a-sdk[postgresql,mysql,sqlite]"]
 
@@ -47,6 +48,7 @@ all = [
   "a2a-sdk[grpc]",
   "a2a-sdk[telemetry]",
   "a2a-sdk[signing]",
+  "a2a-sdk[vertex]",
 ]
 
 [project.urls]
--- a/src/a2a/contrib/tasks/vertex_task_converter.py
+++ b/src/a2a/contrib/tasks/vertex_task_converter.py
@@ -1,0 +1,158 @@
+try:
+    from vertexai import types as vertexai_types
+except ImportError as e:
+    raise ImportError(
+        'vertex_task_converter requires vertexai. '
+        'Install with: '
+        "'pip install a2a-sdk[vertex]'"
+    ) from e
+
+import base64
+import json
+
+from a2a.types import (
+    Artifact,
+    DataPart,
+    FilePart,
+    FileWithBytes,
+    FileWithUri,
+    Part,
+    Task,
+    TaskState,
+    TaskStatus,
+    TextPart,
+)
+
+
+_TO_SDK_TASK_STATE = {
+    vertexai_types.State.STATE_UNSPECIFIED: TaskState.unknown,
+    vertexai_types.State.SUBMITTED: TaskState.submitted,
+    vertexai_types.State.WORKING: TaskState.working,
+    vertexai_types.State.COMPLETED: TaskState.completed,
+    vertexai_types.State.CANCELLED: TaskState.canceled,
+    vertexai_types.State.FAILED: TaskState.failed,
+    vertexai_types.State.REJECTED: TaskState.rejected,
+    vertexai_types.State.INPUT_REQUIRED: TaskState.input_required,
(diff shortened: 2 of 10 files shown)


tests/server/tasks/test_result_aggregator.py
```diff
@@ -260,17 +262,21 @@ async def mock_consume_generator():
 
         # Mock _continue_consuming to check if it's called by create_task
         self.aggregator._continue_consuming = AsyncMock()
+        sentinel_task = asyncio.ensure_future(asyncio.sleep(0))
+        mock_create_task.return_value = sentinel_task
```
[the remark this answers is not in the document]
kevinlu310: addressed

src/a2a/server/request_handlers/default_request_handler.py
guglielmo-san, line 260: I believe that this comment should be also removed now

kevinlu310: This one is about managing the producer task (_run_event_stream) in non-blocking flows -- i.e., lifecycle management of the agent execution itself (when does it get cleaned up, what if multiple requests hit the same task, etc.). That's a broader concern unrelated to the GC bug we fixed.

PR #783, opened 2026-03-06
feat(server): add v0.3 legacy compatibility for database models

### LINKED ISSUE #715: [Task]: Database stores compatibility and migration
Need to research how new version is going to behave against old data based on Pydantic types and define migration path.

Applies to push notification and task stores. Ideally new code should support reading all data and map it to proto based data model without extra migrations.

sokoliva:
## Description
Implements a mechanism to handle legacy v0.3 data stored in the database. When a `Task` or a `TaskPushNotificationConfig` does not have a `protocol_version` set to `0.1`, it validates and converts the data using v0.3 Pydantic models and conversion utilities. This ensures backward compatibility for existing records containing string-based enums and old field structures.

## Changes
- The `status`, `artifacts`, and `history` fields in `TaskMixin` now use standard SQLAlchemy JSON columns with explicit Python type hints (Mapped[Any], Mapped[list[Any]]).
- Removed `PydanticType` and `PydanticListType`: These custom SQLAlchemy types are no longer needed as serialization is now handled at the Store level.
- Updated `_to_orm` to use MessageToDict on the entire Task object
- Updated `_from_orm`:
   - v1.0: Uses ParseDict
   - Legacy (v0.3): Uses Pydantic's model_validate to reconstruct the legacy Task tree before converting to core types.
 - Updated Tests: 
    - Removed obsolete tests for the deleted Pydantic type 
    - updated the Task Store integration tests to verify the new `0.3 type to 1.0 type` conversion logic

## Contributing Guide

Fixes #715 🦕

### DIFF
--- a/src/a2a/client/transports/rest.py
+++ b/src/a2a/client/transports/rest.py
@@ -109,6 +109,8 @@ class RestTransport(ClientTransport):
         params = MessageToDict(request)
         if 'id' in params:
             del params['id']  # id is part of the URL path
+        if 'tenant' in params:
+            del params['tenant']
 
         response_data = await self._execute_request(
             'GET',
@@ -127,12 +129,16 @@ class RestTransport(ClientTransport):
         context: ClientCallContext | None = None,
     ) -> ListTasksResponse:
         """Retrieves tasks for an agent."""
+        params = MessageToDict(request)
+        if 'tenant' in params:
+            del params['tenant']
+
         response_data = await self._execute_request(
             'GET',
             '/tasks',
             request.tenant,
             context=context,
-            params=MessageToDict(request),
+            params=params,
         )
         response: ListTasksResponse = ParseDict(
             response_data, ListTasksResponse()
@@ -185,8 +191,10 @@ class RestTransport(ClientTransport):
         params = MessageToDict(request)
         if 'id' in params:
             del params['id']
-        if 'task_id' in params:
-            del params['task_id']
+        if 'taskId' in params:
+            del params['taskId']
+        if 'tenant' in params:
+            del params['tenant']
 
         response_data = await self._execute_request(
             'GET',
@@ -208,8 +216,10 @@ class RestTransport(ClientTransport):
     ) -> ListTaskPushNotificationConfigsResponse:
         """Lists push notification configurations for a specific task."""
         params = MessageToDict(request)
-        if 'task_id' in params:
-            del params['task_id']
+        if 'taskId' in params:
+            del params['taskId']
+        if 'tenant' in params:
+            del params['tenant']
 
         response_data = await self._execute_request(
             'GET',
@@ -233,8 +243,10 @@ class RestTransport(ClientTransport):
         params = MessageToDict(request)
         if 'id' in params:
             del params['id']
-        if 'task_id' in params:
-            del params['task_id']
+        if 'taskId' in params:
+            del params['taskId']
+        if 'tenant' in params:
+            del params['tenant']
 
         await self._execute_request(
             'DELETE',
--- a/src/a2a/server/request_handlers/rest_handler.py
+++ b/src/a2a/server/request_handlers/rest_handler.py
@@ -7,7 +7,6 @@ from google.protobuf.json_format import (
     MessageToDict,
     MessageToJson,
     Parse,
-    ParseDict,
 )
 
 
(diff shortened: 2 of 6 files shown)


tests/server/tasks/test_database_task_store.py
```diff
@@ -683,4 +683,139 @@ async def test_owner_resource_scoping(
...
+async def test_get_0_3_task_detailed(
+    db_store_parameterized: DatabaseTaskStore,
+) -> None:
+    """Test retrieving a detailed legacy v0.3 task from the database.
+
+    This test simulates a database that already contains legacy v0.3 JSON data
+    (string-based enums, different field names) and verifies that the store
+    correctly converts it to the modern Protobuf-based Task model.
+    """
+    from a2a.compat.v0_3 import types as types_v03
```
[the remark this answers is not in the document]
sokoliva: It is not missing it

src/a2a/server/models.py
```diff
@@ -292,3 +324,25 @@ class PushNotificationConfigModel(PushNotificationConfigMixin, Base):
     """Default push notification config model with standard table name."""
 
     __tablename__ = 'push_notification_configs'
+
+
+_LEGACY_CONVERSIONS: dict[type, tuple[type[BaseModel], Callable]] | None = None
+
+
+def _get_legacy_conversions() -> dict[type, tuple[type[BaseModel], Callable]]:
```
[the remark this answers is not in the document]
sokoliva: Outdated

src/a2a/server/models.py
ishymko, line 1: Let's add `protocol_version` column (#715) for future upgrades so that we can differentiate data model, it can be nullable so that existing data it not touched and new code should write `1.0` to this column. It's going to be another DB migration but with 1.0 it's a good chance to set it up for longer term. With this we don't need to fallback but we can rely on this column instead (no value - read as pydantic and convert, `1.0` - read as new).

sokoliva: Done: https://github.com/a2aproject/a2a-python/pull/789

src/a2a/server/models.py
```diff
@@ -153,25 +48,17 @@ class TaskMixin:
...
-    # Using declared_attr to avoid conflict with Pydantic's metadata
-    @declared_attr
-    @classmethod
-    def task_metadata(cls) -> Mapped[dict[str, Any] | None]:
-        """Define the 'metadata' column, avoiding name conflicts with Pydantic."""
-        return mapped_column(JSON, nullable=True, name='metadata')
+    # Using 'task_metadata' to avoid conflict with SQLAlchemy's 'Base.metadata'
+    task_metadata: Mapped[dict[str, Any] | None] = mapped_column(
+        JSON, nullable=True, name='metadata'
+    )
```
ishymko: Is this change required? I believe that original comment about `declared_attr` is not accurate but maybe it's better to address it in a separate PR.

src/a2a/server/tasks/database_task_store.py
```diff
@@ -129,33 +129,27 @@ def _to_orm(self, task: Task, owner: str) -> TaskModel:
...
-            artifacts=list(task.artifacts) if task.artifacts else [],
-            history=list(task.history) if task.history else [],
-            task_metadata=(
-                MessageToDict(task.metadata) if task.metadata.fields else None
-            ),
+            status=task_dict.get('status'),
+            artifacts=task_dict.get('artifacts', []),
+            history=task_dict.get('history', []),
+            task_metadata=task_dict.get('metadata'),
+            protocol_version='1.0',
```
ishymko: I'd stick with the typed approach for safety here unless it was required to switch to dictionary and string keys.

src/a2a/server/tasks/database_task_store.py
```diff
@@ -129,33 +129,27 @@ def _to_orm(self, task: Task, owner: str) -> TaskModel:
...
+            'context_id': task_model.context_id,
+            'status': task_model.status,
+            'artifacts': task_model.artifacts,
+            'history': task_model.history,
+            'metadata': task_model.task_metadata,
+        }
+        if task_model.protocol_version == '1.0':
+            return ParseDict(task_dict, Task())
+        legacy_task = types_v03.Task.model_validate(task_dict)
+        return conversions.to_core_task(legacy_task)
```
ishymko: I'd suggest to write explicit mappings for both legacy and proto versions: `Task(id=tak_model.id...` and `types_v03.Task(id=...` to avoid raw string usage in strings keys.

sokoliva: Done.

src/a2a/server/tasks/database_task_store.py
sokoliva: It was not necessary. Fixed.

PR #784, opened 2026-03-06
refactor(client)!: introduce ServiceParameters for extensions and include it in ClientCallContext

guglielmo-san:
# Description

This PR refactors the client API definitions to streamline extension handling and unify transport logic:

- A new class to store extensions, integrated into the ClientCallContext. This reduces the need for having separate extension fields across the API definition.
- Extracted common HTTP argument parsing logic into shared helper functions used by both REST and JSON-RPC transports.
- Interceptor logic has been temporarily removed, as it will be redesigned and reintroduced in an upcoming PR.

### DIFF
--- a/src/a2a/compat/v0_3/grpc_transport.py
+++ b/src/a2a/compat/v0_3/grpc_transport.py
@@ -85,11 +85,7 @@ def _handle_grpc_stream_exception(
 class CompatGrpcTransport(ClientTransport):
     """A backward compatible gRPC transport for A2A v0.3."""
 
-    def __init__(
-        self,
-        channel: Channel,
-        agent_card: a2a_pb2.AgentCard | None
-    ):
+    def __init__(self, channel: Channel, agent_card: a2a_pb2.AgentCard | None):
         """Initializes the CompatGrpcTransport."""
         self.agent_card = agent_card
         self.channel = channel


src/a2a/client/transports/grpc.py
```diff
@@ -295,36 +292,32 @@ async def close(self) -> None:
...
+                metadata.append((key.lower(), value))
         return metadata
 
     def _get_grpc_timeout(
         self, context: ClientCallContext | None
     ) -> float | None:
-        return context.timeout if context else None
+        if context:
+            return context.timeout
+        return None
```
ishymko: Why was it changed?

src/a2a/client/transports/rest.py
ishymko: Not used anymore?

src/a2a/client/base_client.py
```diff
@@ -116,11 +89,29 @@ async def send_message(
...
+    def _apply_client_config(self, request: SendMessageRequest) -> None:
+        if not request.configuration.blocking and self._config.polling:
+            request.configuration.blocking = not self._config.polling
+        if (
+            not request.configuration.HasField('push_notification_config')
+            and self._config.push_notification_configs
+        ):
+            request.configuration.push_notification_config.CopyFrom(
+                self._config.push_notification_configs[0]
+            )
```
ishymko, line 106: Maybe not in this PR, but I feel like plural `push_notification_configs` is some legacy, as we only take `[0]`.

src/a2a/client/transports/grpc.py
guglielmo-san: reverted

src/a2a/client/base_client.py
guglielmo-san: Yeah there is still a lot of clean up to do

PR #795, opened 2026-03-10
feat(compat): REST and JSONRPC servers compatible with 0.3 clients.

### LINKED ISSUE #742: [Feat]: 0.3 compatibility

bartek-w:
* Unified v0.3 Handling (grpc_handler.py): Refactored the CompatGrpcHandler to delegate directly to
     RequestHandler03, removing ad-hoc data transformations and unifying translation logic across all transports.
   * REST & JSONRPC Adapters: Implemented and tested legacy HTTP adapters. Added comprehensive tests in
     test_rest_handler.py and test_rest_fastapi_app_compat.py to validate exact JSON payload shapes and legacy
     routing (e.g., /v0.3/v1/message:send).
   * Feature Flag Integration: Introduced the enable_v0_3_compat flag for A2AStarletteApplication and
     A2ARESTFastAPIApplication. When disabled, legacy requests are securely rejected with standard errors to
     prevent overhead.
   * Cross-Version Integration: Upgraded the v1.0 reference test server (server_1_0.py) to announce and mount
     legacy REST interfaces. The automated test matrix now explicitly validates jsonrpc and rest paths alongside
     grpc.

Fixes #742

### DIFF
--- a/src/a2a/server/apps/rest/fastapi_app.py
+++ b/src/a2a/server/apps/rest/fastapi_app.py
@@ -24,6 +24,7 @@ else:
         _package_fastapi_installed = False
 
 
+from a2a.compat.v0_3.rest_adapter import REST03Adapter
 from a2a.server.apps.jsonrpc.jsonrpc_app import CallContextBuilder
 from a2a.server.apps.rest.rest_adapter import RESTAdapter
 from a2a.server.context import ServerCallContext
@@ -55,6 +56,7 @@ class A2ARESTFastAPIApplication:
             [AgentCard, ServerCallContext], Awaitable[AgentCard] | AgentCard
         ]
         | None = None,
+        enable_v0_3_compat: bool = False,
     ):
         """Initializes the A2ARESTFastAPIApplication.
 
@@ -72,6 +74,8 @@ class A2ARESTFastAPIApplication:
             extended_card_modifier: An optional callback to dynamically modify
               the extended agent card before it is served. It receives the
               call context.
+            enable_v0_3_compat: If True, mounts backward-compatible v0.3 protocol
+              endpoints under the '/v0.3' path prefix using REST03Adapter.
         """
         if not _package_fastapi_installed:
             raise ImportError(
@@ -87,6 +91,18 @@ class A2ARESTFastAPIApplication:
             card_modifier=card_modifier,
             extended_card_modifier=extended_card_modifier,
         )
+        self.enable_v0_3_compat = enable_v0_3_compat
+        self._v03_adapter = None
+
+        if self.enable_v0_3_compat:
+            self._v03_adapter = REST03Adapter(
+                agent_card=agent_card,
+                http_handler=http_handler,
+                extended_agent_card=extended_agent_card,
+                context_builder=context_builder,
+                card_modifier=card_modifier,
+                extended_card_modifier=extended_card_modifier,
+            )
 
     def build(
         self,
@@ -98,7 +114,7 @@ class A2ARESTFastAPIApplication:
 
         Args:
             agent_card_url: The URL for the agent card endpoint.
-            rpc_url: The URL for the A2A JSON-RPC endpoint.
+            rpc_url: The URL for the A2A REST endpoint base path.
             **kwargs: Additional keyword arguments to pass to the FastAPI constructor.
 
         Returns:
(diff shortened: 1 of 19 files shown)


src/a2a/server/apps/rest/fastapi_app.py
```diff
@@ -72,6 +74,8 @@ def __init__(  # noqa: PLR0913
             extended_card_modifier: An optional callback to dynamically modify
               the extended agent card before it is served. It receives the
               call context.
+            enable_v0_3_compat: If True, mounts backward-compatible v0.3 protocol
+              endpoints under the '/v0.3' path prefix using REST03Adapter.
```
ishymko, line 78: Let's allow customizing this path as well, I believe it should be also possible for one to use same base URL from the agent card perspective as due to `v1` removal in 1.0 (🥲) paths shouldn't overlap either way: `https://example.com/a2a/rest` with `https://example.com/a2a/rest/message:send` and `https://example.com/a2a/rest/v1/message:send`.

src/a2a/server/apps/rest/fastapi_app.py
```diff
@@ -117,4 +133,20 @@ async def get_agent_card(request: Request) -> Response:
...
+            v03_router = APIRouter()
+            for route, callback in v03_adapter.routes().items():
+                v03_router.add_api_route(
+                    f'{rpc_url}/v0.3{route[0]}', callback, methods=[route[1]]
+                )
+
+            @v03_router.get(f'{rpc_url}/v0.3{agent_card_url}')
+            async def get_v03_agent_card(request: Request) -> Response:
+                card = await v03_adapter.handle_get_agent_card(request)
+                return JSONResponse(card)
```
ishymko, line 148: Shouldn't be needed for compat, this thing is supposed to be a "static" transport agnostic well-know path ([14.3. Well-Known URI Registration](https://a2a-protocol.org/latest/specification/#143-well-known-uri-registration)) common for 1.0 and 0.3 and serve this "compatible" union agent card (already done here I believe).

src/a2a/server/apps/rest/fastapi_app.py
bartek-w: due to v1 vs empty prefix, I think it should be possible to remove v0.3 prefix completely in another PR. I will keep it as it is and make it customizable if this will be required.

src/a2a/server/apps/rest/fastapi_app.py
bartek-w: As above. I think we will drop the prefix completely. If not I will adjust it.

Issue #757, opened 2026-03-03
[Task]: Revisit client interceptors


ishymko:
Client-side interceptors are only implemented for JSON-RPC transport ([`jsonrpc.py`](https://github.com/a2aproject/a2a-python/blob/8b647bdc8ad38551a19ba6bdd566ac9f9c714ef8/src/a2a/client/transports/jsonrpc.py#L86-L107)) and the data model itself is defined around JSON-RPC and not in a transport-agnostic way: [`middleware.py`](https://github.com/a2aproject/a2a-python/blob/8b647bdc8ad38551a19ba6bdd566ac9f9c714ef8/src/a2a/client/middleware.py#L24).

REST contains a no-op function ([`rest.py`](https://github.com/a2aproject/a2a-python/blob/8b647bdc8ad38551a19ba6bdd566ac9f9c714ef8/src/a2a/client/transports/rest.py#L72-L81) and there is nothing in [`grpc.py`](https://github.com/a2aproject/a2a-python/blob/1.0-dev/src/a2a/client/transports/grpc.py).

For 1.0 breaking release there is a chance to do it at the client level in a transport agnostic way. Current implementation is likely a leftover from pre-multitransport era.

PR #796, opened 2026-03-10
feat(client): transport agnostic interceptors

guglielmo-san:
# Description

This PR refactors the client interceptors architecture, centralizing their execution within the BaseClient rather than delegating them to the underlying transport implementations.
These interceptors allow to modify request before being sent to the server, and server responses before are sent back to the caller, with an early return mechanism.
The Authentication interceptor is updated as well to store authentication values in the ServiceParameters class of the ClientCallContext.

Fix: #757

### DIFF
--- a/src/a2a/client/interceptors.py
+++ b/src/a2a/client/interceptors.py
@@ -12,6 +12,7 @@ from a2a.types.a2a_pb2 import (  # noqa: TC001
     AgentCard,
 )
 
+
 @dataclass
 class BeforeArgs:
     """Arguments passed to the interceptor before a method call."""
--- a/src/a2a/client/base_client.py
+++ b/src/a2a/client/base_client.py
@@ -456,9 +456,7 @@ class BaseClient(Client):
             await self._intercept_after(after_args, before_result['executed'])
 
             tracker = ClientTaskManager()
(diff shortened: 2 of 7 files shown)


src/a2a/client/interceptors.py
```diff
@@ -0,0 +1,65 @@
...
+    method: str
+    value: Any
+
+
+@dataclass
+class ClientCallResult:
+    """Represents the method and its associated result payload."""
+
+    method: str
+    value: Any
```
ishymko: Maybe let's inline those into `XxxArgs`? Without extensive typing like we had in a2a-js I think there is not much value in extracting as a separate types.

src/a2a/client/interceptors.py
```diff
@@ -0,0 +1,65 @@
...
+class AfterArgs:
+    """Arguments passed to the interceptor after a method call completes."""
+
+    result: ClientCallResult
+    agent_card: AgentCard
+    context: ClientCallContext | None = None
+    early_return: bool = False
+
+
+class ClientCallInterceptor:
```
ishymko: Should probably be inherited from `ABC`? I see we do it for pure abstract classes here (`ClientTransport`, `IDGenerator`).

PR #810, opened 2026-03-11
chore: incorporate latest proto changes (blocking -> return_immediately)

ishymko:
Original spec PR: https://github.com/a2aproject/A2A/pull/1507.

Required a bunch of inversions including compat conversion updates. The gotcha here is that 0.3 SDK **already** uses blocking mode by default (although it's not according to the spec), so this spec change is about applying a de-facto standard.


tests/compat/v0_3/test_conversions.py
```diff
@@ -1767,8 +1764,7 @@ def test_to_core_send_message_request_no_configuration():
         ),
     )
     core_req = to_core_send_message_request(v03_req)
-    # Default is True if configuration is absent
-    assert core_req.configuration.blocking is True
+    assert not core_req.HasField('configuration')
```
bartek-w: Please confirm this change is intentional.

ishymko: Yeah it is due to removal of 

```python
else:
        core_req.configuration.blocking = True  # Default for A2A
```

so `configuration` is not set anymore, but updated to reflect intention a bit better (assert that `return_immediately is False`).

PR #814, opened 2026-03-11
fix: taskslist error on invalid page token and response serialization

guglielmo-san:
# Description

This PR addresses two main issues related to the list_tasks JSON-RPC endpoint:

- Fixed how list_tasks formats the ListTasksResponse to ensure all fields are explicitly printed, even when they carry default values (such as an empty next_page_token).
-  Replaced generic ValueError exceptions with A2A SDK's standard InvalidParamsError when handling malformed or invalid pagination tokens. This correctly surfaces -32602 Invalid params via JSON-RPC.


src/a2a/server/request_handlers/jsonrpc_handler.py
```diff
@@ -373,7 +373,11 @@ async def list_tasks(
             response = await self.request_handler.on_list_tasks(
                 request, context
             )
-            result = MessageToDict(response, preserving_proto_field_name=False)
+            result = MessageToDict(
+                response,
+                preserving_proto_field_name=False,
+                always_print_fields_with_no_presence=True,
+            )
```
ishymko, line 380: Can we have a test asserting the full payload to see exactly what's there?

ishymko (approved):

src/a2a/server/request_handlers/jsonrpc_handler.py
guglielmo-san: Done

PR #821, opened 2026-03-12
fix(server): handle unwrapped legacy push notification configurations

sokoliva:
# Description
Fix the database PushNotificationConfig store when reading legacy data from version 0.3. The store now correctly handles cases when legacy PushNotificationConfig was stored rather than the full wrapped model.

### DIFF
--- a/.jscpd.json
+++ b/.jscpd.json
@@ -1,5 +1,13 @@
 {
-  "ignore": ["**/.github/**", "**/.git/**", "**/tests/**", "**/src/a2a/grpc/**", "**/.nox/**", "**/.venv/**"],
+  "ignore": [
+    "**/.github/**",
+    "**/.git/**",
+    "**/tests/**",
+    "**/src/a2a/grpc/**",
+    "**/src/a2a/compat/**",
+    "**/.nox/**",
+    "**/.venv/**"
+  ],
   "threshold": 3,
   "reporters": ["html", "markdown"]
 }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 [project.optional-dependencies]
 http-server = ["fastapi>=0.115.2", "sse-starlette", "starlette"]
 encryption = ["cryptography>=43.0.0"]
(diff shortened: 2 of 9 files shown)


src/a2a/server/tasks/database_push_notification_config_store.py
```diff
@@ -341,19 +344,23 @@ async def delete_info(
                 )
 
     def _parse_config(
-        self, json_payload: str, protocol_version: str | None = None
+        self,
+        json_payload: str,
+        task_id: str,
```
holtskinner: This field should be Optional since it's only required for 0.3
```suggestion
        task_id: str | None = None,
```

Also update the docstring to indicate this.

src/a2a/server/tasks/database_push_notification_config_store.py
holtskinner, line 370:
```suggestion
        inner_config = types_v03.PushNotificationConfig.model_validate_json(
            json_payload
        )
return conversions.to_core_task_push_notification_config(
types_v03.TaskPushNotificationConfig(
            task_id=task_id,
            push_notification_config=inner_config,
)
)
```

nit: Fix the formatting for my suggestion, but this can be simpler to read?

src/a2a/server/tasks/database_push_notification_config_store.py
```diff
@@ -341,19 +344,23 @@ async def delete_info(
...
-            types_v03.TaskPushNotificationConfig.model_validate_json(
-                json_payload
-            )
+        inner_config = types_v03.PushNotificationConfig.model_validate_json(
+            json_payload
+        )
+        legacy_instance = types_v03.TaskPushNotificationConfig(
+            task_id=task_id,
+            push_notification_config=inner_config,
         )
```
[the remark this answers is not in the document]
sokoliva: there is no legacy data wrapped as TaskPushNotificationConfig

src/a2a/server/tasks/database_push_notification_config_store.py
sokoliva: Agreed. Done.

PR #823, opened 2026-03-13
fix(compat): allow DataPart.data to accept Any to align with spec

### LINKED ISSUE #714: [Bug]: `DataPart` should accept objects as data, not just dicts.
### What happened?

According to the [A2A spec's proto file](https://github.com/a2aproject/A2A/blob/629190aea9fe5838bb712008e6d24156680af197/specification/a2a.proto#L228), the `DataPart` accepts any object, not just maps, whereas the [Python implementation](https://github.com/a2aproject/a2a-python/blob/9968f9c07f105bae8a6b296aeb6dea873b3b88b0/src/a2a/types.py#L277) specifically only allows dicts.

Perhaps the Python implementation is still in the process of adopting the updated proto?

Akshat8510:
**Closes #714**

**Changes:**
- Updated `DataPart.data` type hint from `dict[str, Any]` to `Any` in `src/a2a/compat/v0_3/types.py`.

**Context:**
Following the 1.0-dev refactor, the `DataPart` class now resides in the `compat/v0_3` layer. This change ensures that the compatibility layer aligns with the A2A specification (where `data` can be an object, array, string, etc.). 

I verified that the new 1.0 `types` (generated from Protobuf) use the `Part` message, which already correctly handles these types.

### DIFF
--- a/src/a2a/compat/v0_3/types.py
+++ b/src/a2a/compat/v0_3/types.py
@@ -274,7 +274,7 @@ class DataPart(A2ABaseModel):
     Represents a structured data segment (e.g., JSON) within a message or artifact.
     """
 
-    data: dict[str, Any]
+    data: Any
     """
     The structured data content.
     """


Akshat8510:
@gemini-code-assist/gemini review

ishymko:
Hey @Akshat8510, thanks for taking a look, please see https://github.com/a2aproject/a2a-python/issues/714#issuecomment-4053563521, this is working as intended.

PR #837, opened 2026-03-16
fix: Avoid sending a single json message in multiple chunks. 

### LINKED ISSUE #742: [Feat]: 0.3 compatibility

bartek-w:
Fixes #742

### DIFF
--- a/src/a2a/compat/v0_3/rest_adapter.py
+++ b/src/a2a/compat/v0_3/rest_adapter.py
@@ -1,4 +1,5 @@
 import functools
+import json
 import logging
 
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable
@@ -103,9 +104,9 @@ class REST03Adapter(RESTAdapterInterface):
 
         async def event_generator(
             stream: AsyncIterable[Any],
-        ) -> AsyncIterator[dict[str, dict[str, Any]]]:
+        ) -> AsyncIterator[str]:
             async for item in stream:
-                yield {'data': item}
+                yield json.dumps(item)
 
         return EventSourceResponse(
             event_generator(method(request, call_context))
--- a/src/a2a/compat/v0_3/rest_handler.py
+++ b/src/a2a/compat/v0_3/rest_handler.py
@@ -1,9 +1,9 @@
 import logging
 
-from collections.abc import AsyncIterable, AsyncIterator
+from collections.abc import AsyncIterator
 from typing import TYPE_CHECKING, Any
 
-from google.protobuf.json_format import MessageToDict, MessageToJson, Parse
+from google.protobuf.json_format import MessageToDict, Parse
 
 
 if TYPE_CHECKING:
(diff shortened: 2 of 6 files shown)


src/a2a/compat/v0_3/rest_adapter.py
```diff
@@ -103,9 +104,9 @@ async def _handle_streaming_request(
 
         async def event_generator(
             stream: AsyncIterable[Any],
-        ) -> AsyncIterator[dict[str, dict[str, Any]]]:
+        ) -> AsyncIterator[str]:
             async for item in stream:
-                yield {'data': item}
+                yield json.dumps(item)
```
ishymko, line 109: Similar code for JSON-RPC uses `{'data': json.dumps(item)}` ([source](https://github.com/a2aproject/a2a-python/blob/7e6b3c2332fca4a56ea521fc12223ff2422de19c/src/a2a/server/apps/jsonrpc/jsonrpc_app.py#L593)). Are both equivalent? 

From the added tests I can see that it works as intended but just curious why two are different.

bartek-w: Yes, It looks like this is equivalent (tested with manual change).

PR #839, opened 2026-03-16
refactor: disable buf generate on build

guglielmo-san:
# Description

This PR disables the automatic generation of A2A types for v1.0 and for compatibility with v0.3 at build phase.
`gen_proto.sh` and `gen_proto_compat.sh` can be used as scripts to generate those types

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -22,8 +22,6 @@ jobs:
       - name: Add uv to PATH
         run: |
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
-      - name: Install Buf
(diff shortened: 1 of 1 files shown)


src/a2a/types/a2a.json
holtskinner, line 1: Should we be maintaining a separate copy of this here? If we care about fixed versions, we can use the git tags from https://github.com/a2aproject/A2A to reference the a2a.json file

guglielmo-san: This json file is generated from the new `a2a.proto`, and currently used by FastAPI in jsonrpc app. However the usage is wrong as this a2a.json contains the rest targets, as it was generated from the .proto. So I would remove it entirely in the next PR

PR #843, opened 2026-03-17
fix: Use POST method for REST endpoint /tasks/{id}:subscribe

### LINKED ISSUE #840: [Bug]: REST `/tasks/{id}:subscribe` uses GET method instead of POST
It should be POST as per the spec: [0.3](https://a2a-protocol.org/v0.3.0/specification/#356-method-mapping-reference-table), [1.0](https://a2a-protocol.org/latest/specification/#53-method-mapping-reference).

Applies to both 0.3 version from `main` and 1.0 RC from `1.0-dev`, to both client and server.

 **BUT** fixing it for 0.3 is a bit dangerous as one may had assumptions around wrong implementation, so the proposal is:

1) For 1.0 server and client it should just be fixed.
2) 1.0 client in compat 0.3 mode should use POST method as per the spec to be compatible with other SDKs which implement it properly. We may have a flag in the legacy compat transport to use GET if one needs to talk to old Python server which uses wrong method.
3) 0.3 server should accept both to work with old clients.

bartek-w:
POST should be always use for /tasks/{id}:subscribe.
Decisions for backward compatibility with invalid protocol implementations:
1.0 server: Accept both POST and GET
1.0 client: Always use POST
0.3 server: Accept both POST and GET
0.3 client: Try POST first, on HTTP 405 error retry with GET. Cache the retry state to ensure that there is at most one retry attempt per transport instance.

Fixes #840

### DIFF
--- a/src/a2a/client/transports/rest.py
+++ b/src/a2a/client/transports/rest.py
@@ -262,7 +262,6 @@ class RestTransport(ClientTransport):
             f'/tasks/{request.id}:subscribe',
             request.tenant,
             context=context,
-            json=MessageToDict(request),
         ):
             yield event
 
--- a/src/a2a/compat/v0_3/rest_transport.py
+++ b/src/a2a/compat/v0_3/rest_transport.py
@@ -64,13 +64,14 @@ class CompatRestTransport(ClientTransport):
         httpx_client: httpx.AsyncClient,
         agent_card: AgentCard | None,
         url: str,
+        subscribe_method_override: str | None = None,
     ):
         """Initializes the CompatRestTransport."""
         self.url = url.removesuffix('/')
         self.httpx_client = httpx_client
         self.agent_card = agent_card
-        self._subscribe_method = 'POST'
-        self._subscribe_retry_attempted = False
+        self._subscribe_method_override = subscribe_method_override
+        self._subscribe_auto_method_override = subscribe_method_override is None
 
     async def send_message(
         self,
@@ -285,17 +286,12 @@ class CompatRestTransport(ClientTransport):
         on this transport instance. If both fail with 405, it will default back
         to POST for next calls but will not retry again.
         """
-        if self._subscribe_method == 'POST':
-            json_body = MessageToDict(request, preserving_proto_field_name=True)
-        else:
-            json_body = None
-
+        subscribe_method = self._subscribe_method_override or 'POST'
         try:
             async for event in self._send_stream_request(
-                self._subscribe_method,
+                subscribe_method,
                 f'/v1/tasks/{request.id}:subscribe',
                 context=context,
-                json=json_body,
             ):
                 yield event
         except A2AClientError as e:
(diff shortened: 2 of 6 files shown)


src/a2a/compat/v0_3/rest_transport.py
```diff
@@ -273,13 +276,45 @@ async def subscribe(
...
+
+        This method implements backward compatibility logic for the subscribe
+        endpoint. It first attempts to use POST, which is the official method
+        for A2A subscribe endpoint. If the server returns 405 Method Not Allowed,
+        it falls back to GET and remembers this preference for future calls
+        on this transport instance. If both fail with 405, it will default back
+        to POST for next calls but will not retry again.
+        """
+        if self._subscribe_method == 'POST':
+            json_body = MessageToDict(request, preserving_proto_field_name=True)
```
ishymko: I think the body should be empty here, `id` is encoded in the path and there is nothing else.

[0.3 has some `name`](https://a2a-protocol.org/v0.3.0/specification/#79-tasksresubscribe) in the payload but I am sure it's a mistake.

src/a2a/compat/v0_3/rest_transport.py
```diff
@@ -273,13 +276,45 @@ async def subscribe(
...
+                if self._subscribe_retry_attempted:
+                    self._subscribe_method = 'POST'
+                    raise
+                else:
+                    self._subscribe_method = 'GET'
+                    self._subscribe_retry_attempted = True
+                    async for event in self.subscribe(request, context=context):
+                        yield event
+            else:
+                raise
```
ishymko, line 314: Nice! I don't think we can do anything better than this given the situation we're in.

Can we have a config to force one specific method just as a safety net?

src/a2a/server/request_handlers/rest_handler.py
```diff
@@ -159,9 +159,15 @@ async def on_subscribe_to_task(
         Yields:
             JSON serialized objects containing streaming events
         """
-        task_id = request.path_params['id']
+        params = SubscribeToTaskRequest()
+        if request.method == 'POST':
+            body = await request.body()
+            if body:
+                Parse(body, params)
```
ishymko: Do we need it? Both `id` and `tenant` should be taken from path.

src/a2a/client/transports/rest.py
```diff
@@ -262,10 +262,11 @@ async def subscribe(
     ) -> AsyncGenerator[StreamResponse]:
         """Reconnects to get task updates."""
         async for event in self._send_stream_request(
-            'GET',
+            'POST',
             f'/tasks/{request.id}:subscribe',
             request.tenant,
             context=context,
+            json=MessageToDict(request),
```
ishymko: I believe we don't need it, both `id` and `tenant` are in the path.

src/a2a/compat/v0_3/rest_transport.py
bartek-w: As discussed offline: added flag to override behaviour.

Issue #797, opened 2026-03-10
[Task]: Revisit server "apps" and handlers


ishymko:
### Duplicated well-known agent card handler
JSON-RPC and REST transports are defined via separate "apps" each including well-known agent card handler, supporting them both on one instance requires manual work for serving agent card at well-known path as otherwise it'd be registered twice under transport specific routes. In 0.3 the code serializing agent card for a well-known path is also different: JSON-RPC one uses JSON Schema based model and REST uses ProtoJSON, so depending on which app is used one gets different agent card format - it may not manifest itself for simple agent cards, example of the difference: https://github.com/a2aproject/a2a-js/pull/344.

Current approach also doesn't support gRPC well, if one wants to serve agent card without any other HTTP based transports and only use gRPC.

### FastAPI and Starlette

There is also independent FastAPI and Starlette code (likely due to OpenAPI UI integration in FastAPI, but with proto-based types it doesn't work out of the box either way). The logic is fragmented across different concepts which are not symmetric across transports: adapters, handlers and applications.

#### JSON-RPC
- `JSONRPCApplication` [`jsonrpc_app.py`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/apps/jsonrpc/jsonrpc_app.py) - base class for both FastAPI and Starlette, operates on Starlette data model, parses incoming request into JSON-RPC data model and invokes appropriate methods of `JSONRPCHandler` (see below).
- `A2AFastAPIApplication` [`fastapi_app.py`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/apps/jsonrpc/fastapi_app.py) - FastAPI implementation of the above. Has `build` to create an app or `add_routes_to_app` to update existing instance created externally.
- `A2AStarletteApplication` [`starlette_app.py`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/apps/jsonrpc/starlette_app.py) - Starlette implementation of the above.
- `JSONRPCHandler` [`jsonrpc_handler.py`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/request_handlers/jsonrpc_handler.py) - operates on A2A data model for input and uses `dict` as output, doesn't have a dependency on starlette, invokes domain transport agnostic `RequestHandler`.

#### HTTP+JSON/REST
[...]

PR #848, opened 2026-03-17
chore: add sample client and server

ishymko:
There is a dedicated repo for samples ([a2a-samples](https://github.com/a2aproject/a2a-samples)), however having something runnable in the repository is useful for manual testing and ad-hoc experimentation. Add trivial hello-world/echo agent and a simple client CLI to connect to it.

It exposes all transports and both 1.0 and 0.3 version.

1. Run server
```bash
$ uv run samples/hello_world_agent.py
```
2. (_separate terminal_) Run client (supports `--transport` as well)
```bash
$ uv run samples/cli.py
Connecting to http://127.0.0.1:41241 (preferred transport: Any)

✓ Agent Card Found:
  Name: Sample Agent
  Picked Transport: JsonRpcTransport

Connected! Send a message or type /quit to exit.
You: test
TaskStatusUpdate [TASK_STATE_WORKING]: Processing your question... 
TaskArtifactUpdate [response]: Hello World! You said: 'test'. Thanks for your message! 
TaskStatusUpdate [TASK_STATE_COMPLETED]: 
--- Task Completed ---
You: 
```


samples/hello_world_agent.py
```diff
@@ -0,0 +1,254 @@
...
+    )
+    jsonrpc_app = jsonrpc_app_builder.build()
+
+    main_app = FastAPI()
+    main_app.mount('/a2a/jsonrpc', jsonrpc_app)
+    main_app.mount('/a2a/rest', rest_app)
+
+    @main_app.get('/.well-known/agent-card.json')
+    def get_agent_card() -> Any:
+        """Return the agent card metadata used for automatic client connection."""
```
guglielmo-san: why this is needed

ishymko: Changed to use `add_routes_to_app` from JSON-RPC to rely on it's agent card handler. There is no good way to do it now as both REST and JSON-RPC have their own handlers for well-known path. Hopefully should become better with #797.

PR #859, opened 2026-03-17
fix(client): prevent `ClientFactory.create()` from mutating shared config extensions

### LINKED ISSUE #858: ClientFactory.create() mutates shared ClientConfig.extensions across calls
## Bug Description

`ClientFactory` is designed for reuse — create one factory, call `create()` multiple times for different agents. However, when `create()` is called with per-client `extensions`, the merged list is written back to `self._config.extensions`, permanently mutating the shared config. Extensions silently accumulate across calls.

## Reproduction

```python
config = ClientConfig(
    httpx_client=httpx.AsyncClient(),
    extensions=['base-ext'],
)
factory = ClientFactory(config)

factory.create(card, extensions=['ext-a'])
factory.create(card, extensions=['ext-b'])

# BUG: config.extensions is now ['base-ext', 'ext-a', 'ext-b']
# Expected: config.extensions == ['base-ext']
```

## Root Cause

In `src/a2a/client/client_factory.py`, the `create()` method:
```python
all_extensions = self._config.extensions.copy()
if extensions:
    all_extensions.extend(extensions)
    self._config.extensions = all_extensions  # <-- mutates shared config
```

Line 242 writes the merged list back to `self._config.extensions`.

## Expected Behavior

`create()` should produce a per-call config (using `dataclasses.replace`) and never mutate the factory's shared `ClientConfig`.

## Related

This is the same class of bug as #744 (mutable field defaults in `ServerCallContext`).

matanweks:
## Summary

- **Fix**: `ClientFactory.create()` was mutating `self._config.extensions` when called with per-client extensions, causing extensions to silently accumulate across calls.
- **Solution**: Use `dataclasses.replace()` to create a per-call config copy instead of mutating the shared `ClientConfig`. This matches the immutability pattern from #744.
- **Test**: Added regression test verifying that `config.extensions` remains unchanged after multiple `create()` calls with different extensions.

Fixes #858

## Test plan

- [x] New test `test_client_factory_create_does_not_mutate_config_extensions` verifies no mutation
- [x] All existing `test_client_factory.py` tests pass
- [x] `ruff check` and `ruff format` pass
- [x] `mypy` passes with no issues

### DIFF
--- a/src/a2a/client/client_factory.py
+++ b/src/a2a/client/client_factory.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import logging
 
 from collections.abc import Callable
@@ -239,15 +240,20 @@ class ClientFactory:
         all_extensions = self._config.extensions.copy()
         if extensions:
             all_extensions.extend(extensions)
-            self._config.extensions = all_extensions
+
+        config = (
+            dataclasses.replace(self._config, extensions=all_extensions)
+            if extensions
+            else self._config
+        )
 
         transport = self._registry[transport_protocol](
-            card, transport_url, self._config, interceptors or []
+            card, transport_url, config, interceptors or []
         )
 
         return BaseClient(
             card,
-            self._config,
+            config,
             transport,
             all_consumers,
             interceptors or [],
--- a/tests/client/test_client_factory.py
+++ b/tests/client/test_client_factory.py
@@ -266,3 +266,20 @@ async def test_client_factory_connect_with_consumers_and_interceptors(
         call_args = mock_base_client.call_args[0]
         assert call_args[3] == [consumer1]
         assert call_args[4] == [interceptor1]
+
+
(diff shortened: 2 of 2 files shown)


guglielmo-san:
@matanweks Thank you for reporting this issue. In the new [alpha0](https://github.com/a2aproject/a2a-python/releases/tag/v1.0.0-alpha.0) release we changed the way extensions are handled, and this behavior won't occur anymore. 
Can you try the new version?

matanweks:
Thanks for the quick reply! I verified against alpha0 and can confirm the bug no longer applies, so there's nothing left to mutate.
 Happy to close both the issue and the PR. Thanks!

matanweks:
Closing — the underlying bug was resolved in v1.0.0-alpha.0 by moving extensions from `ClientConfig` to a per-call `ServiceParameters` model, eliminating the shared state mutation entirely.

PR #865, opened 2026-03-19
feat: A2A Version Header validation on server side.

bartek-w:


src/a2a/server/request_handlers/jsonrpc_handler.py
ishymko: Why do we need `validate_async_generator` twice?

bartek-w: Looks like gemini mess with the code on final cleanup. Fixed.

PR #873, opened 2026-03-19
refactor(server)!: migrate from Application wrappers to Starlette route-based endpoints for jsonrpc

guglielmo-san:
# Description

This PR refactors the jsonrpc server implementation to expose Starlette Route components directly (via AgentCardRoutes, JsonRpcRoutes) instead of requiring full FastAPI or Starlette application wrappers. 

Ref #797  🦕

### DIFF
--- a/src/a2a/contrib/tasks/vertex_task_converter.py
+++ b/src/a2a/contrib/tasks/vertex_task_converter.py
@@ -1,4 +1,5 @@
 try:
+    from google.genai import types as genai_types
     from vertexai import types as vertexai_types
 except ImportError as e:
     raise ImportError(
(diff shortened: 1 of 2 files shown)


src/a2a/server/routes/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+    StarletteUserProxy,
+)
+from a2a.server.routes.jsonrpc_routes import JsonRpcRoutes
+
+
+__all__ = [
+    'AgentCardRoutes',
+    'CallContextBuilder',
+    'DefaultCallContextBuilder',
+    'JsonRpcDispatcher',
```
ishymko: Let's not export for now.

src/a2a/server/routes/agent_card_routes.py
ishymko, line 1: Here and for the other `XxxRoutes` - maybe can be a function? We just instantiate it to read `routes`.

src/a2a/server/routes/jsonrpc_routes.py
```diff
@@ -0,0 +1,107 @@
...
+        context_builder: CallContextBuilder | None = None,
+        card_modifier: Callable[[AgentCard], Awaitable[AgentCard] | AgentCard]
+        | None = None,
+        extended_card_modifier: Callable[
+            [AgentCard, ServerCallContext], Awaitable[AgentCard] | AgentCard
+        ]
+        | None = None,
+        enable_v0_3_compat: bool = False,
+        rpc_url: str = DEFAULT_RPC_URL,
+        middleware: Sequence[Middleware] | None = None,
```
ishymko: Should be possible with https://starlette.dev/middleware/#applying-middleware-to-groups-of-routes, I'd avoid mirroring Starlette API for this here.

PR #877, opened 2026-03-20
feat: EventQueue - unify implementation between python versions

### LINKED ISSUE #869: [Feat]: Improve server concurrency architecture

The final goal of this effort is to relax or describe limitations on concurrency in AgentExecutor.

In existing version there is multiple hidden assumptions internally between classes as well as in API (e.g. AgentExecutor.cancel() can't wait for AgentExecutor.execute to finish).

The plan is to make incremental changes on concurrency to ensure we can track down invalid change if new concurrency issue is introduced.

_No response_

### Additional context

_No response_

### Code of Conduct

bartek-w:
Introduced a compatibility layer using the culsans library to backport asyncio.Queue.shutdown functionality to Python versions older than 3.13. Previous implementation was broken (deadlocks and inconsistent behaviour with 3.13 implementation). Culsans library allowed for unified code between versions.

 EventConsumer now starts a background task to gracefully wait for queue to finish.

This is one of the steps towards better concurrency model in a2a python sdk.

Fixes #869

### DIFF
--- a/src/a2a/server/events/event_consumer.py
+++ b/src/a2a/server/events/event_consumer.py
@@ -32,7 +32,6 @@ class EventConsumer:
         self.queue = queue
         self._timeout = 0.5
         self._exception: BaseException | None = None
-        self._close_task: asyncio.Task[None] | None = None
         logger.debug('EventConsumer initialized')
 
     async def consume_one(self) -> Event:
@@ -107,25 +106,13 @@ class EventConsumer:
                     )
                 )
 
-                # Initiate a graceful close in the background.
-                # We use immediate=False to ensure tapped child queues can finish
-                # draining their events (preventing data loss for slow consumers).
-                # We use clear_parent_events=True to prevent deadlocks since this
-                # consumer is stopping and won't dequeue any remaining trailing events.
-                # We use create_task instead of awaiting it directly because awaiting
-                # immediate=False would block this consumer until all child queues
-                # are also drained.
                 # Make sure the yield is after the close events, otherwise
                 # the caller may end up in a blocked state where this
                 # generator isn't called again to close things out and the
                 # other part is waiting for an event or a closed queue.
                 if is_final_event:
                     logger.debug('Stopping event consumption in consume_all.')
-                    self._close_task = asyncio.create_task(
-                        self.queue.close(
-                            immediate=False, clear_parent_events=True
-                        )
-                    )
+                    await self.queue.close(True)
                     yield event
                     break
                 yield event
--- a/tests/server/events/test_event_consumer.py
+++ b/tests/server/events/test_event_consumer.py
@@ -461,6 +461,7 @@ async def test_consume_all_handles_validation_error(
         )
 
 
+@pytest.mark.xfail(reason='https://github.com/a2aproject/a2a-python/issues/869')
 @pytest.mark.asyncio
 async def test_graceful_close_allows_tapped_queues_to_drain() -> None:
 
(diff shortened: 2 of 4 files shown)


src/a2a/server/events/event_consumer.py
```diff
@@ -130,7 +135,7 @@ async def consume_all(self) -> AsyncGenerator[Event]:
             except asyncio.TimeoutError:  # pyright: ignore [reportUnusedExcept]
                 # This class was made an alias of built-in TimeoutError after 3.11
                 continue
-            except (QueueClosed, asyncio.QueueEmpty):
+            except (QueueShutDown, asyncio.QueueEmpty):
                 # Confirm that the queue is closed, e.g. we aren't on
                 # python 3.12 and get a queue empty error on an open queue
                 if self.queue.is_closed():
```
yarolegovich, line 128: idk whether this will be a problem, but `is_closed()` reads `self._is_closed` without `with self._lock`

src/a2a/server/events/event_consumer.py
```diff
@@ -114,13 +107,25 @@ async def consume_all(self) -> AsyncGenerator[Event]:
...
+                # are also drained.
                 # Make sure the yield is after the close events, otherwise
                 # the caller may end up in a blocked state where this
                 # generator isn't called again to close things out and the
                 # other part is waiting for an event or a closed queue.
                 if is_final_event:
                     logger.debug('Stopping event consumption in consume_all.')
-                    await self.queue.close(True)
+                    self._close_task = asyncio.create_task(
+                        self.queue.close(
```
yarolegovich: just checking - can self.queue.close raise an exception, what's the behavior of asyncio.create_task in this case and are we fine with it?

tests/server/events/test_event_consumer.py
```diff
@@ -459,3 +459,103 @@ async def test_consume_all_handles_validation_error(
...
+    await producer_task
+    fast_events = await fast_task
+    slow_events = await slow_task
+
+    assert len(fast_events) == 3
+    assert len(slow_events) == 3
+
+
+@pytest.mark.asyncio
+async def test_background_close_deadlocks_on_trailing_events() -> None:
```
yarolegovich, line 524: Trailing events are AgentExecutor implementation bug. Might be easier to guard from those on write?
 
1. Parent and child queues both receive [final, trailing].
2. Parent consumer hits the final event first, fires a background close task that calls child.close(immediate=False) - it blocks on child.queue.join().
3. Child consumer then drains [final] and its own _close_task calls child.close(immediate=False, clear_parent_events=True), but the guard (_is_closed already True) makes it return early - the trailing event is never cleared.
4. child.queue.join() waits forever

<details>

```py
@pytest.mark.asyncio
async def test_child_queue_trailing_events_deadlock() -> None:
    parent_queue = EventQueue(max_queue_size=10)
    child_queue = parent_queue.tap()

    await parent_queue.enqueue_event(Message(message_id='final'))
    await parent_queue.enqueue_event(Message(message_id='trailing'))

    # Intercept child.queue.join() so we can synchronize deterministically
    # instead of relying on asyncio.sleep timing.
    child_join_reached = asyncio.Event()
    original_child_queue = child_queue.queue

    class ChildJoinWrapper:
        def __getattr__(self, name: str) -> Any:
            return getattr(original_child_queue, name)

        async def join(self) -> None:
            child_join_reached.set()
            await original_child_queue.join()

    child_queue.queue = ChildJoinWrapper()

    parent_consumer = EventConsumer(parent_queue)
    child_consumer = EventConsumer(child_queue)

    # Parent consumes fast: [final] -> breaks on final.
    parent_events = [event async for event in parent_consumer.consume_all()]
    assert len(parent_events) == 1

    # Wait until the parent's background _close_task has progressed through
    # child.close() and is now blocked on child.queue.join().
    await asyncio.wait_for(child_join_reached.wait(), timeout=1.0)
    assert child_queue.is_closed(), (
        "Parent's close should have set child._is_closed"
    )

    # Child consumes: dequeues [final] (shutdown(False) still serves
    # existing items).  Its _close_task calls child.close(immediate=False, clear_parent_events=True)
    # but _is_closed is already True -> returns early. Trailing event is never cleared.
    child_events = [event async for event in child_consumer.consume_all()]
    assert len(child_events) == 1

    # Parent's _close_task is stuck on child.queue.join() because the
    # trailing event was never dequeued or task_done'd.
    assert parent_consumer._close_task is not None
    try:
        await asyncio.wait_for(parent_consumer._close_task, timeout=0.5)
    except asyncio.TimeoutError:
        pytest.fail(
            'Deadlock: parent close task hung on child.queue.join() '
            'because trailing event in child queue was never cleared'
        )
```
</details>

src/a2a/server/events/event_consumer.py
bartek-w: Not sure, but added to my list. I'm trying to keep changes focused to simplify reviews.

src/a2a/server/events/event_consumer.py
bartek-w: I've undo the background task change as it is not solving the core issue.

tests/server/events/test_event_consumer.py
bartek-w: You are right. I was planning to add this block on trailing messages, but thought this could solve it for now. I will do that approach while keeping the change focused on EventQueue python version unification.

PR #882, opened 2026-03-20
refactor: Enforce ServerCallContext in request handling

### LINKED ISSUE #718: [Task]: make `ServerCallContext` class parameter mandatory
Currently `ServerCallContext` in an optional parameter in [request handlers](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/request_handlers/request_handler.py#L21) and [task stores](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/task_store.py) methods. Since we use `ServerCallContext` for [mandatory authorization and authentication](https://a2a-protocol.org/latest/specification/#131-data-access-and-authorization-scoping) it should non-optional.

By making the `ServerCallContext` mandatory, we eliminate the risk of un-scoped database queries and ensure that the A2A server acts as a secure gatekeeper at the data layer.

sokoliva:
# Description
- Make `ServerCallContext` a mandatory parameter across all `TaskStore` implementations (`TaskStore` interface, `DatabaseTaskStore`, and `InMemoryTaskStore`) and update tests.
- Make `ServerCallContext` a mandatory parameter in `RequestContext`.

Previously, context defaulted to None, which could allow callers to bypass authorization scoping if the context was set to None. By requiring the context, we guarantee that the `owner_resolver` always has the necessary request context to determine scope boundaries.

Fixes #718 🦕

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler.py
+++ b/src/a2a/server/request_handlers/default_request_handler.py
@@ -49,6 +49,7 @@ from a2a.types.a2a_pb2 import (
 from a2a.utils.errors import (
     InternalError,
     InvalidParamsError,
+    PushNotificationNotSupportedError,
     TaskNotCancelableError,
     TaskNotFoundError,
     UnsupportedOperationError,
@@ -494,7 +495,7 @@ class DefaultRequestHandler(RequestHandler):
         Requires a `PushNotifier` to be configured.
         """
         if not self._push_config_store:
-            raise UnsupportedOperationError
+            raise PushNotificationNotSupportedError
 
         task_id = params.task_id
         task: Task | None = await self.task_store.get(task_id, context)
@@ -520,7 +521,7 @@ class DefaultRequestHandler(RequestHandler):
         Requires a `PushConfigStore` to be configured.
         """
         if not self._push_config_store:
-            raise UnsupportedOperationError
+            raise PushNotificationNotSupportedError
 
         task_id = params.task_id
         config_id = params.id
@@ -592,7 +593,7 @@ class DefaultRequestHandler(RequestHandler):
         Requires a `PushConfigStore` to be configured.
         """
         if not self._push_config_store:
-            raise UnsupportedOperationError
+            raise PushNotificationNotSupportedError
 
         task_id = params.task_id
         task: Task | None = await self.task_store.get(task_id, context)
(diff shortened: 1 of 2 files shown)


src/a2a/server/tasks/task_manager.py
```diff
@@ -31,7 +31,7 @@ def __init__(
         context_id: str | None,
         task_store: TaskStore,
         initial_message: Message | None,
-        context: ServerCallContext | None = None,
+        context: ServerCallContext,
```
guglielmo-san: Let's move all the required parameters at the beginning of the init

src/a2a/server/agent_execution/simple_request_context_builder.py
guglielmo-san: i'm not sure if we should keep is as optional in the `RequestContext`

sokoliva: I agree. I made it mandatory.

src/a2a/server/tasks/task_store.py
ishymko, line 1: There is also `VertexTaskStore` under `contrib` folder which requires similar changes.

src/a2a/server/agent_execution/simple_request_context_builder.py
```diff
@@ -68,19 +68,21 @@ async def build(
         ):
             tasks = await asyncio.gather(
                 *[
-                    self._task_store.get(task_id)
+                    self._task_store.get(
+                        task_id, context or ServerCallContext()
```
ishymko: As `context` can't be `None` this `or` shouldn't be required.

src/a2a/server/request_handlers/default_request_handler.py
ishymko, line 1: Same as https://github.com/a2aproject/a2a-python/pull/882/changes#r2994063488: I see we have some `or ServerCallContext()` in this file.

src/a2a/server/agent_execution/context.py
ishymko: I believe can be done here as well?

```diff
-    def call_context(self) -> ServerCallContext | None:
+    def call_context(self) -> ServerCallContext:
```

ishymko: Here and in some other places below: as it can't be `None` anymore we can skip extra checks.

sokoliva: Thank you.

PR #886, opened 2026-03-22
refactor(client)!: Rename ClientFactory to A2AClientFactory

zeroasterisk:
Completed task 555: Renamed `ClientFactory` to `A2AClientFactory` for backwards compatibility and spec alignment while auto-selecting transport based on `supportedInterfaces`. Exported both `ClientFactory` and `A2AClientFactory`.

### DIFF
--- a/src/a2a/client/__init__.py
+++ b/src/a2a/client/__init__.py
@@ -16,7 +16,11 @@ from a2a.client.client import (
     ClientEvent,
     Consumer,
 )
-from a2a.client.client_factory import ClientFactory, minimal_agent_card
+from a2a.client.client_factory import (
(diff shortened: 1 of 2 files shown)


ishymko (commented):
I am thinking about the opposite: renaming existing `A2AXxx` prefixed things to drop the prefix as we now have more non-prefixed than prefixed: https://github.com/a2aproject/a2a-python/blob/c18fb605786206f522d1d3c14b9b856c950e265d/src/a2a/client/__init__.py#L33-L49

Also for 1.0 should be fine to drop aliases, as it's a breaking release. Given that data model changed completely this renaming shouldn't contribute significantly to the migration.

PR #915, opened 2026-03-30
fix(client): export TenantTransportDecorator and fix docstring

4444J99:
## Summary

`TenantTransportDecorator` is used internally by `ClientFactory` to wrap transports when an `AgentInterface` specifies a tenant, but it is not exported from the `a2a.client.transports` package. This means users who need to compose transport decorators manually (e.g. stacking tenant resolution with a future `RetryTransport`) must import from the private module path.

This PR:

- **Exports `TenantTransportDecorator`** from `a2a.client.transports.__init__` and adds it to `__all__`, consistent with how `ClientTransport`, `JsonRpcTransport`, `RestTransport`, and `GrpcTransport` are exported.
- **Fixes the `send_message()` docstring** which incorrectly read *"Sends a streaming message request to the agent and yields responses as they arrive"* — a copy-paste from `send_message_streaming()`. Corrected to *"Sends a non-streaming message request to the agent."*
- **Adds test coverage for `close()` delegation** and async context manager usage, verifying the decorator properly delegates lifecycle management to the underlying transport.

## Test plan

- [x] All 14 `test_tenant_decorator.py` tests pass (12 existing + 2 new)
- [x] Verified `from a2a.client.transports import TenantTransportDecorator` resolves correctly

### DIFF
--- a/src/a2a/client/transports/__init__.py
+++ b/src/a2a/client/transports/__init__.py
@@ -3,6 +3,7 @@
 from a2a.client.transports.base import ClientTransport
 from a2a.client.transports.jsonrpc import JsonRpcTransport
 from a2a.client.transports.rest import RestTransport
+from a2a.client.transports.tenant_decorator import TenantTransportDecorator
 
 
 try:
@@ -16,4 +17,5 @@ __all__ = [
     'GrpcTransport',
     'JsonRpcTransport',
     'RestTransport',
+    'TenantTransportDecorator',
 ]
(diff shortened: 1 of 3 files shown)


ishymko:
Thank you for the PR!

This type is not exported by design and is considered to be an implementation detail. For the mentioned relation to the future `RetryTransport` we may reconsider it later.

Feel free to factor out other fixes so that we can merge them.

4444J99:
Hi — checking in on this small fix. It exports `TenantTransportDecorator` from `a2a.client.transports` (consistent with how the other transport classes are exported) and fixes a copy-paste docstring in `send_message()`. All 14 tests pass. Happy to rebase against `1.0-dev` if needed.

PR #964, opened 2026-04-13
feat: Enforce `Task` or `Message` as an initial response for Send Streaming Message

sokoliva:
# Description
The A2A spec ([§3.1.2](https://a2a-protocol.org/latest/specification/#312-send-streaming-message)) requires that the first event in a streaming response is always a Task or Message object. However, agents commonly emit TaskStatusUpdateEvent(WORKING) as their first event without explicitly emitting a Task first. Previously the v2 handler forwarded these events as-is, violating the spec.

# Changes
- `default_request_handler_v2.py`: when the v2 handler (`DefaultRequestHandlerV2`) receives a new streaming request and the agent's first event is not a Task or Message, it now injects a `Task(SUBMITTED)` before forwarding the agent's event.
- update `samples/cli.py` to reflect changes in `send_message` return type 
- update `samples/cli.py` to expect `Task` or `Message` as first streaming event
- add a `README.md` for samples
- utilize `get_artifact_text`, `get_message_text` helpers in `samples/cli.py`

# Tested 
Changes seen in `tests/integration/test_scenarios.py`

```
uv run samples/cli.py
Connecting to http://127.0.0.1:41241 (preferred transport: Any)

✓ Agent Card Found:
  Name: Sample Agent
  Picked Transport: JsonRpcTransport

Connected! Send a message or type /quit to exit.
You: hi
Task [state=TASK_STATE_SUBMITTED]
TaskStatusUpdate [state=TASK_STATE_WORKING]: Processing your question...
TaskArtifactUpdate [name=response]: Hello World! Nice to meet you!
TaskStatusUpdate [state=TASK_STATE_COMPLETED]: 
--- Task Completed ---
You: bye
Task [state=TASK_STATE_SUBMITTED]
TaskStatusUpdate [state=TASK_STATE_WORKING]: Processing your question...
TaskArtifactUpdate [name=response]: Goodbye! Have a wonderful day!
TaskStatusUpdate [state=TASK_STATE_COMPLETED]: 
--- Task Completed ---
You: /quit
```

# Note
The legacy handler (`DefaultRequestHandler`) is intentionally left unchanged to avoid breaking existing users who consume its streaming output directly. Also, `spec v0.3` does not define the initial event in send stream message.

Fixes: #965

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -9,7 +9,7 @@ This directory contains runnable examples demonstrating how to build and interac
 | `hello_world_agent.py` | **Server** | A2A agent server |
 | `cli.py` | **Client** | Interactive terminal client |
 
-All three samples are designed to work together out of the box: the agent listens on `http://127.0.0.1:41241`, which is the default URL used by both clients.
+The samples are designed to work together out of the box: the agent listens on `http://127.0.0.1:41241`, which is the default URL used by the client.
 ---
 
 ## `hello_world_agent.py` — Agent Server
@@ -53,21 +53,8 @@ uv run python samples/cli.py
 uv run python samples/cli.py --url http://192.168.1.10:41241 --transport GRPC
 ```
 
+Then type a message like `hello` and press Enter.
+
 Type `/quit` or `/exit` to stop, or press `Ctrl+C`.
 
----
 
-
-## Quick Start
-
-In two separate terminals:
-
-```bash
-# Terminal 1 — start the agent
-uv run python samples/hello_world_agent.py
-
-# Terminal 2 — start the client
-uv run python samples/cli.py
-```
-
-Then type a message like `hello` and press Enter.
--- a/samples/README.md
+++ b/samples/README.md
@@ -1,0 +1,73 @@
+# A2A Python SDK — Samples
+
+This directory contains runnable examples demonstrating how to build and interact with an A2A-compliant agent using the Python SDK.
+
+## Contents
+
+| File | Role | Description |
+|---|---|---|
+| `hello_world_agent.py` | **Server** | A2A agent server |
+| `cli.py` | **Client** | Interactive terminal client |
+
+All three samples are designed to work together out of the box: the agent listens on `http://127.0.0.1:41241`, which is the default URL used by both clients.
+---
+
+## `hello_world_agent.py` — Agent Server
+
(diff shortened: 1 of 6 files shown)


samples/cli.py
```diff
@@ -11,25 +11,22 @@
...
     stream: Any, current_task_id: str | None
 ) -> str | None:
-    async for event, task in stream:
-        if not task:
-            continue
+    async for event in stream:
         if not current_task_id:
-            current_task_id = task.id
-
+            current_task_id = event.task.id
```
guglielmo-san: What if event.task is not set?
imo we should 
1. If event.HasField('message') -> continue (streaming cannot return Message type)
2. if not current_task_id and not event.HasField('Task')
In this case we should throw an error -> the first event from the TaskExecutor should be a Task (https://a2a-protocol.org/latest/specification/#312-send-streaming-message)

samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+This directory contains runnable examples demonstrating how to build and interact with an A2A-compliant agent using the Python SDK.
+
+## Contents
+
+| File | Role | Description |
+|---|---|---|
+| `hello_world_agent.py` | **Server** | A2A agent server |
+| `cli.py` | **Client** | Interactive terminal client |
+
+The samples are designed to work together out of the box: the agent listens on `http://127.0.0.1:41241`, which is the default URL used by the client.
```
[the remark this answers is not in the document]
sokoliva: Fixed

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -302,16 +304,35 @@ async def on_message_send_stream(  # noqa: D102
...
+                and not isinstance(event, (Task, Message))
+            ):
+                # Agent didn't emit a Task/Message first.
+                # The stream MUST begin with a Task or Message.
+                submitted_task = Task(
+                    id=task_id,
+                    context_id=context_id,
+                    status=TaskStatus(state=TaskState.TASK_STATE_SUBMITTED),
+                    history=[params.message],
+                )
```
guglielmo-san, line 332: The task_manager is already covering this use case, we can use the task created and saved on the task store
https://github.com/a2aproject/a2a-python/blob/6b5651102326ae4c7e8936c1109a0f09693c9034/src/a2a/server/tasks/task_manager.py#L150

sokoliva:
Proceeding with a different solution - should emit error.

PR #1016, opened 2026-04-24
fix(server): deliver push notifications across all owners

### LINKED ISSUE #1015: [Bug]: `BasePushNotificationSender` requires request-scoped `ServerCallContext` in constructor
The lifetime of the `BasePushNotificationSender` is singleton as it's a dependency of `DefaultRequestHandler`, however it requires per-request `ServerCallContext` as a mandatory constructor parameter.

`ServerCallContext` is used for `PushNotificationConfigStore` for `owner` resolution. Currently the only possible way is to construct a dummy `ServerCallContext` which won't have any user information.

Re #709 when this change was made.

sokoliva:
# Description
Fix a silent multi-tenant bug where push notifications were dropped for any deployment using authenticated users, and split the `PushNotificationConfigStore` read API into a user-callable owner-scoped read (`get_info`) and an internal cross-owner read for dispatch (`get_info_for_dispatch`).

## The bug
`BasePushNotificationSender` accepted a `ServerCallContext` at construction time and called `config_store.get_info(task_id, self._call_context)` at dispatch. Because the sender is a process-wide singleton, callers passed a dummy `ServerCallContext()` (e.g. `itk/main.py`). The default `OwnerResolver`then resolved the dummy to the empty-string owner, which never matched any real registrar's partition. Result: `get_info` returned `[]` and every notification was silently dropped in any deployment with real authentication.

## The fix
- **New non-abstract method** `PushNotificationConfigStore.get_info_for_dispatch(task_id)` returns every config for the task across all owners. Implemented in the in-memory and database stores. Custom 1.0 subclasses inherit a default implementation that forwards to `get_info(task_id, ServerCallContext())` preserving their 1.0 behavior exactly and emits a `DeprecationWarning` 
- **`BasePushNotificationSender`** no longer takes `context` in `__init__` and now calls `get_info_for_dispatch`. Identity is no longer held on the sender.
- **`get_info(task_id, context)` is unchanged** and remains owner-scoped. Used by the user-callable read endpoints.
  
The split encodes the asymmetry in the type system: the user-callable method requires a context, the dispatch-only method does not. Authorization (check if the user can create a config for a specific task) happens at registration (`set_info`), not at dispatch.

Fixes #1015  🦕


src/a2a/server/tasks/database_push_notification_config_store.py
```diff
@@ -335,6 +338,35 @@ async def get_info(
                     )
             return configs
 
+    async def get_info_for_dispatch(
```
ishymko, line 346: Is there a way to reuse something across owner scoped read and this one? I feel like many things are similar, it's just one more predicate to `where`.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import warnings
```
ishymko: I don't see any other usages of this package in this project, can we do a reguler logger warning instead?

src/a2a/server/tasks/database_push_notification_config_store.py
sokoliva: Yes there is. Done.

ishymko (approved):
🚀

PR #1004, opened 2026-04-21
feat(helpers): add non-text Part, Message, and Artifact helpers

martimfasantos:
## Description

- [x] Follow the [`CONTRIBUTING` Guide](https://github.com/a2aproject/a2a-python/blob/main/CONTRIBUTING.md).
- [x] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [x] Ensure the tests and linter pass (Run `bash scripts/format.sh` from the repository root to format)
- [ ] Appropriate docs were updated (if necessary)

---

### Summary

`proto_helpers.py` provided `new_text_message` and `new_text_artifact` for the text Part variant, but nothing for the three remaining Part types (`data`, `raw`, `url`). This PR completes the set.

**The `data` case is especially awkward without a helper.** `Part.data` is `google.protobuf.Value` in the v1.0 spec, which requires a non-obvious `ParseDict` dance to construct from a plain Python value:

```python
# Without helper
from google.protobuf.json_format import ParseDict
from google.protobuf import struct_pb2
part = Part(data=ParseDict({"answer": "hello"}, struct_pb2.Value()))

# With helper
part = new_data_part({"answer": "hello"})
```

### New helpers

**Part primitives** (building blocks, mirror the existing implicit `Part(text=...)` pattern):

| Helper | Part field | Accepts |
|---|---|---|
| `new_data_part(data)` | `data` (`google.protobuf.Value`) | Any JSON-serializable value (dict, list, str, …) |
| `new_raw_part(raw, media_type, filename)` | `raw` (`bytes`) | Raw bytes with optional MIME type and filename |
| `new_url_part(url, media_type, filename)` | `url` (`str`) | URL with optional MIME type and filename |

**Message helpers** (mirror `new_text_message`):

| Helper | Wraps |
|---|---|
| `new_data_message(data, role, context_id, task_id)` | `new_data_part` |
| `new_raw_message(raw, media_type, filename, role, context_id, task_id)` | `new_raw_part` |
| `new_url_message(url, media_type, filename, role, context_id, task_id)` | `new_url_part` |

**Artifact helpers** (mirror `new_text_artifact`):

| Helper | Wraps |
|---|---|
| `new_data_artifact(name, data, description, artifact_id)` | `new_data_part` |
| `new_raw_artifact(name, raw, media_type, filename, description, artifact_id)` | `new_raw_part` |
| `new_url_artifact(name, url, media_type, filename, description, artifact_id)` | `new_url_part` |

### Changes

[...]

PR #1017, opened 2026-04-24
feat(helpers): add non-text Part extractors and artifact update events

martimfasantos:
## Summary
- Adds `get_data_parts`, `get_raw_parts`, `get_url_parts` extractors to complement the existing `get_text_parts`, completing the Part extraction API for all content types.
- Adds `new_data_artifact_update_event`, `new_raw_artifact_update_event`, `new_url_artifact_update_event` to complement the existing `new_text_artifact_update_event`, completing the streaming event helpers for all content types.
- Re-exports all new helpers and the previously unexposed helpers from #1004 (`new_data_part`, `new_raw_part`, `new_url_part`, `new_text_part`, `new_data_message`, `new_raw_message`, `new_url_message`, `new_data_artifact`, `new_raw_artifact`, `new_url_artifact`) in `a2a.helpers.__init__` so they are accessible via `from a2a.helpers import ...`.

## Changes
- `src/a2a/helpers/proto_helpers.py`: Added 6 new functions (3 extractors + 3 artifact update events).
- `src/a2a/helpers/__init__.py`: Updated imports and `__all__` to expose all 16 new helpers.
- `tests/helpers/test_proto_helpers.py`: Added 12 new test cases covering all new functions.

## Testing
- All mandatory checks pass: `ruff`, `mypy`, `pyright`, `pytest` (50 helper tests pass, 1675 total).

BEGIN_COMMIT_OVERRIDE
feat(helpers): add non-text Part extractors and artifact update events
#1017

Release-As: 1.0.3
END_COMMIT_OVERRIDE


tests/helpers/test_proto_helpers.py
```diff
@@ -4,15 +4,20 @@
 
 from a2a.helpers.proto_helpers import (
```
sokoliva: Helper functions in the helper utility should always be developer-facing (#978).
Let's import directly from a2a.helpers so we can avoid forgetting to export them in the future.

```suggestion
from a2a.helpers import (
```

src/a2a/helpers/proto_helpers.py
```diff
@@ -401,6 +401,45 @@ def get_text_parts(parts: Sequence[Part]) -> list[str]:
     return [part.text for part in parts if part.HasField('text')]
 
 
+def get_data_parts(parts: Sequence[Part]) -> list[Any]:
+    """Extracts structured data from all data Parts.
+
+    Each returned element is the Python object obtained from the
```
sokoliva: Each returned element is a Value proto, not a Python object.

sokoliva (changes requested):

tests/helpers/test_proto_helpers.py
martimfasantos: you're right - this was one of the objectives of this PR, but it completely slipped past me. Thanks for catching it :)

src/a2a/helpers/proto_helpers.py
martimfasantos: thanks for the correction — rather than just fixing the description, I went ahead and implemented the actual conversion using `MessageToDict`, making `get_data_parts` return native Python objects. This keeps it symmetric with `new_data_part` which uses `ParseDict` in the opposite direction.

sokoliva (approved):

Issue #1011, opened 2026-04-22
[Task]: Replace deprecated `FieldDescriptor.label` with `is_repeated` in `proto_utils`


sokoliva:
## Summary
`src/a2a/utils/proto_utils.py` currently uses the deprecated `FieldDescriptor.label` property to check whether a protobuf field is repeated. The non-deprecated alternative is `FieldDescriptor.is_repeated`, but adopting it requires a newer minimum version of `protobuf` than we currently declare, which would be a breaking change for downstream consumers.

## Locations
All occurrences are in `src/a2a/utils/proto_utils.py`:
- `src/a2a/utils/proto_utils.py:177` — `_populate_message_from_params` (or similar)
- `src/a2a/utils/proto_utils.py:211` — `_check_required_field_violation`
- `src/a2a/utils/proto_utils.py:252` — `_recurse_validation`
Each site is marked with a `TODO` comment referencing this issue.

## Proposed change
1. Bump the minimum required `protobuf` version in `pyproject.toml` to one that exposes `FieldDescriptor.is_repeated` as a stable, non-deprecated API.
2. Replace all `field.label == FieldDescriptor.LABEL_REPEATED` checks with `field.is_repeated`.
3. Replace all `field.label != FieldDescriptor.LABEL_REPEATED` checks with `not field.is_repeated`.
4. Remove the `TODO` comments and unused `LABEL_REPEATED` import if no longer needed.
5. Schedule the change as part of the next breaking release.

liujuanjuan1984:
We hit this in downstream integration with `a2a-sdk==1.0.2` on a real server path.

Concrete repro:
- `a2a-sdk==1.0.2`
- `protobuf==7.34.1`
- call path: `DefaultRequestHandler.on_get_task()` -> `@validate_request_params` -> `validate_proto_required_fields()` -> `_check_required_field_violation()`
- runtime error: `AttributeError: 'google._upb._message.FieldDescriptor' object has no attribute 'label'`

A few details from our side:
- `a2a-sdk 1.0.2` currently declares `Requires-Dist: protobuf>=5.29.5`, so a normal resolver can select `7.x`
- with `protobuf 5.29.5` and `6.33.5`, `field.label` is still present
- with `protobuf 7.34.1`, `field.label` is absent while `field.is_repeated` exists
- pinning downstream to `protobuf<7` immediately resolves the crash

So the current `field.label` implementation is not compatible with all versions allowed by the published dependency range.

[...]

PR #1019, opened 2026-04-27
fix(deps): cap protobuf below 7

liujuanjuan1984:
## Summary
- tighten the published protobuf dependency range to `<7`
- keep the declared range aligned with the current `proto_utils` implementation, which still relies on `FieldDescriptor.label`
- update `uv.lock` metadata to match

## Why
`a2a-sdk 1.0.2` currently publishes `protobuf>=5.29.5`, which allows resolvers to select `protobuf 7.x`.

In downstream testing, `a2a-sdk==1.0.2` with `protobuf==7.34.1` crashes on the `DefaultRequestHandler.on_get_task()` path:
- `@validate_request_params`
- `validate_proto_required_fields()`
- `_check_required_field_violation()`
- `AttributeError: 'google._upb._message.FieldDescriptor' object has no attribute 'label'`

This PR applies the minimal packaging fix that matches the current implementation choice from #1010 while broader code-level compatibility is discussed in #1011.

Relates to #1011

## Validation
- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run mypy src`
- `uv run pyright src`
- `uv run pytest`

`pytest` result locally:
- `1671 passed, 90 skipped, 3 xfailed, 4 errors`
- the 4 errors were all in `tests/integration/cross_version/client_server/test_client_server.py`
- they were startup failures for the `server_0_3.py` subprocess path using `uv run --with a2a-sdk[grpc]==0.3.24 --no-project ...`
- these failures do not exercise the changed dependency declaration in this PR


sokoliva:
Hi @liujuanjuan1984,

Thank you for bringing this to our attention and for fixing the issue. I replicated the error you got and confirmed the fix you offered.

sokoliva (approved):

PR #1022, opened 2026-04-27
ci: lint test import hygiene with Ruff

liujuanjuan1984:
## Summary
- add a dedicated Ruff config for low-noise test import hygiene checks
- run that check in the existing linter workflow
- clean up the current test import-order and unused-import findings it surfaces

## Why
The repository currently formats files under `tests/`, but skips linting them entirely in CI. This leaves basic, deterministic issues like unused imports, redefinitions, and import ordering unreported.

This change keeps the main Ruff configuration unchanged and adds a narrow test-only check for `F401`, `F811`, and `I001`.

Includes a small follow-up adjustment to one existing legacy streaming scenario test after CI exposed a flaky Python 3.14 path.

## Validation
- `uv run ruff check --output-format=github`
- `uv run ruff check tests --config ruff-tests.toml --output-format=github`
- `uv run ruff format --check`
- `uv run pytest -q tests/client/test_base_client.py tests/server/events/test_event_consumer.py tests/server/request_handlers/test_default_request_handler.py tests/server/routes/test_jsonrpc_dispatcher.py`


tests/compat/v0_3/test_jsonrpc_app_compat.py
```diff
@@ -1,27 +1,36 @@
...
-    TaskStatus as TaskStatus10,
+)
+from a2a.types.a2a_pb2 import (
     TaskState as TaskState10,
 )
-
-from a2a.compat.v0_3 import a2a_v0_3_pb2
+from a2a.types.a2a_pb2 import (
+    TaskStatus as TaskStatus10,
+)
```
sokoliva, line 18: Let's use "combine-as-imports = true" so that imports like these don't get fragmated.

sokoliva:
Hi @liujuanjuan1984,

Thank you for another helpful PR! 

Instead of adding [ruff-tests.toml](https://github.com/a2aproject/a2a-python/pull/1022/changes#diff-4cd50faf4862d8cbb4fab0ebe9311e32f1f7889326fb2c197f1fb0f030fd9693) to the root, can we add another `pyproject.toml` to the `tests` folder ([Reference](https://docs.astral.sh/ruff/configuration/#config-file-discovery:~:text=file%2C%20like%20so%3A-,pyproject.toml,-ruff.toml))? In that case, we should also remove [tests from the exclude bundle in the root pyproject](https://github.com/a2aproject/a2a-python/blob/7af60502532a44f440885157aaae36c024416b85/pyproject.toml#L295).
That way running "uv run ruff format && uv run ruff --check fix" from the repo root would also reformat tests as you described.

sokoliva (changes requested):

liujuanjuan1984:
@sokoliva Thanks for the review. I've updated this PR to follow your suggestions:

- moved the test-specific Ruff config from `ruff-tests.toml` at the repo root to `tests/pyproject.toml`
- removed `tests` from the root Ruff exclude list so root-level Ruff commands now cover the test suite as well
- enabled `combine-as-imports = true` in the test Ruff config and let Ruff rewrite the affected test imports
- simplified the linter workflow so the root Ruff run covers tests directly, and added `tests/pyproject.toml` to the workflow path triggers

Validation run locally:
- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run mypy src`
- `uv run pyright src`
- `uv run pytest`

Could you please take another look when you have a chance?

tests/compat/v0_3/test_jsonrpc_app_compat.py
liujuanjuan1984: done.

sokoliva (approved):

PR #1044, opened 2026-05-04
ci: Switch Type checker to use ty from astral

holtskinner:
- https://docs.astral.sh/ty/
- Replacement for mypy/pyright with faster performance


holtskinner:
@gemini-code-assist review

ishymko (approved):
Thank you!

PR #1059, opened 2026-05-08
ci: consolidate agent guidance in AGENTS.md and improve mistake-reflection workflow

sokoliva:
## Description
Restructures the AI/agent guidance for this repo around three changes:
1. **Promotes `AGENTS.md` to the source of truth** for agent instructions; reduces `GEMINI.md` to a redirect stub.
2. **Adds a mistake-reflection workflow** — a cross-agent SKILL.md plus an `evidence_rules.md` for graduated learnings.
3. **Trims `mandatory_checks.md`** to use the existing `scripts/lint.sh` instead of restating its commands.
## Why move from GEMINI.md to AGENTS.md
`AGENTS.md` is the cross-agent convention now supported by Gemini, Claude Code, OpenCode, Cursor, Codex CLI, etc. — a single file every modern agent auto-loads. `GEMINI.md` is Gemini-specific.

## Mistake reflection workflow
When you notice the agent made a mistake:
1. Tell it. A simple "no, that's wrong" or "where do you see this?" should auto-trigger the mistake-reflection skill. If it doesn't, ask explicitly: "load the mistake-reflection skill."
   If you're using OpenCode (or a similar agent CLI that supports skills), you'll see mistake-reflection listed under /skills once the session has discovered it.
2. Review the new entry the agent appends to docs/ai/ai_learnings.md (gitignored, local-only). Make sure the root cause and proposed rule actually match the mistake.
3. If the rule is generally useful, ask the agent to promote it to docs/ai/evidence_rules.md (or another tracked file) so future sessions see it. Otherwise leave it in the journal — or delete it if it's a one-off.


.agents/skills/mistake-reflection/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+
+- The user corrects a factual claim, code change, or assumption.
+- A tool result contradicts something you just stated or did
+  (lint failure, test failure, type-check failure, file not found,
+  command exit non-zero on something you said would succeed).
+- You re-read a file or doc and realize a prior statement was wrong
+  or unverified.
+- You realize mid-task that you skipped a required step
+  (e.g. didn't read `docs/ai/coding_conventions.md` /
+  `docs/ai/mandatory_checks.md` at task start).
```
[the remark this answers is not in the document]
sokoliva: ok

GEMINI.md
holtskinner, line 1: Not sure if we actually need a dedicated GEMINI.md file, both Gemini CLI and Antigravity can also read AGENTS.md by default if there is no GEMINI.md

https://antigravity.google/changelog#:~:text=condensed%20chat%20UI-,Added%20support%20for%20reading%20rules%20from%20AGENTS.md%20in%20addition%20to%20GEMINI.md,-One%2Dclick%20chat

holtskinner:
Also, I'd recommend changing the commit message to prefix `ci`, `build` or `chore` instead of `docs` because [`docs` triggers a release message for Python in release-please](https://github.com/googleapis/release-please#:~:text=For%20example%2C%20%22docs%22%20is%20a%20prefix%20for%20releasable%20units%20in%20Java%20and%20Python.) when this probably shouldn't as it's internal development only.

GEMINI.md
sokoliva: Makes sense, I deleted it. Thanks.

ishymko (approved):

Issue #734, opened 2026-02-24
[Task]: Consider refining/removing ClientTaskManager


ishymko:
Currently [`ClientTaskManager`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/client/client_task_manager.py) tries to replicate server logic of maintaining task state on the client applying `TaskXxxUpdateEvent`'s.

This logic already diverges from server when it comes to task history:

1. On the client current status update message is added to the history: https://github.com/a2aproject/a2a-python/blob/cced34d4eb22f7ec78422b2d04ce54924e132215/src/a2a/client/client_task_manager.py#L120-L124

while on the server it only contains messages from previous updates:

https://github.com/a2aproject/a2a-python/blob/cced34d4eb22f7ec78422b2d04ce54924e132215/src/a2a/server/tasks/task_manager.py#L143-L152

2. Initial user message is not added to the history in the `ClientTaskManager`, but is added on the server.

Having `Client` returning `Task` object assembled using this logic can be confusing as it is different from the actual server-side state. Especially with 1.0 which specifies history behavior to be partially server-side defined:

> The Task History field contains Messages exchanged during task execution. However, not all Messages are guaranteed to be persisted in the Task history; for example, transient informational messages may not be stored. Messages exchanged prior to task creation may not be stored in Task history. The agent is responsible to determine which Messages are persisted in the Task History.

(*[3.7 Messages and Artifacts](https://a2a-protocol.org/latest/specification/#37-messages-and-artifacts)*)

In addition, relying heavily on task history is rather a leftover from earlier integrations which used task history to stream the actual LLM response (which should be returned as an artifact instead).

Alternative proposal is to create something like `ArtifactsAgregator.from_stream(iterator from send_message)` as a side utility to interpret artifact `append` flag properly.

PR #1060, opened 2026-05-11
feat(client): add ArtifactsAggregator for assembling artifacts from stream

### LINKED ISSUE #1030: Feat]: Client-side utility for assembling artifacts from update events

1.0 release removed `ClientTaskManager` (see #734) and changed client API in a way that it now closely mirrors server API and uses [`StreamResposne`](https://a2a-protocol.org/latest/specification/#323-stream-response) as a return type.

#734 mentions a possible helper for reconstructing artifacts specifically, this issue is to track it.

As stated in #734 maintaining the whole `Task` state locally can result in divergence with server state and should not be relied on by default (especially due to history behavior specification in 1.0, #734 has more details).

For artifacts though the logic is deterministic and can be replicated locally if one wants to maintain full artifact state by interpreting `append` and `lastChunk` flags. Client API should not be changed here, new helper should operate on a stream of `StreamResponse`.

Alex-Resch:
## Description

  Adds `ArtifactsAggregator`, a client-side utility for reconstructing complete
  `Artifact` objects from a stream of `StreamResponse` events, as proposed in #734
  and tracked in #1030.

  The helper interprets the `append` and `last_chunk` flags of
  `TaskArtifactUpdateEvent` to correctly assemble chunked and interleaved artifacts
  without requiring changes to the existing client API.

  ### Usage

  ```python
  from a2a.client import ArtifactsAggregator

  # Assemble a single artifact by ID
  aggregator = ArtifactsAggregator.from_stream(client.send_message(request))
  artifact = await aggregator.get_artifact('my-artifact-id')

  # Assemble all artifacts
  aggregator = ArtifactsAggregator.from_stream(client.send_message(request))
  artifacts = await aggregator.get_all_artifacts()
  ```

  ### Changes
  
  - `src/a2a/client/artifact_aggregator.py` — new ArtifactsAggregator class
  - `src/a2a/client/__init__.py` — export ArtifactsAggregator
  - `docs/migrations/v1_0/README.md` — document usage under "Client: Send Message"
  - `tests/client/test_artifact_aggregator.py` — unit tests for all methods and edge cases

  Fixes #1030 🦕

### DIFF
--- a/src/a2a/client/artifact_aggregator.py
+++ b/src/a2a/client/artifact_aggregator.py
@@ -14,21 +14,6 @@ class ArtifactsAggregator:
     def __init__(self, stream: AsyncIterator[StreamResponse]) -> None:
         self._stream = stream
 
-    @classmethod
-    def from_stream(
-        cls, stream: AsyncIterator[StreamResponse]
-    ) -> 'ArtifactsAggregator':
-        """Create an ArtifactsAggregator from an async stream of StreamResponse events.
-
-        Args:
-            stream: An async iterator of StreamResponse objects, typically obtained
-                from BaseClient.send_message.
-
-        Returns:
-            A new ArtifactsAggregator instance.
-        """
-        return cls(stream)
-
     async def get_artifact(self, artifact_id: str) -> Artifact | None:
         """Assemble and return a single Artifact by its ID from the stream.
 
--- a/src/a2a/client/artifact_aggregator.py
+++ b/src/a2a/client/artifact_aggregator.py
@@ -1,0 +1,96 @@
+from collections.abc import AsyncIterator
+
+from a2a.types import Artifact, StreamResponse
+
+
+class ArtifactsAggregator:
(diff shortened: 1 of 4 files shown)


src/a2a/client/artifact_aggregator.py
```diff
@@ -0,0 +1,96 @@
...
+        """Create an ArtifactsAggregator from an async stream of StreamResponse events.
+
+        Args:
+            stream: An async iterator of StreamResponse objects, typically obtained
+                from BaseClient.send_message.
+
+        Returns:
+            A new ArtifactsAggregator instance.
+        """
+        return cls(stream)
```
sokoliva: This method duplicates the constructor. Instead of using `artifact_aggregator = ArtifactsAggregator.from_stream(base_client.send_message(request))` we could directly use `artifact_aggregator = ArtifactsAggregator(base_client.send_message(request))`.

Consider removing it or rethinking the design of ArtifactsAggregator.

sokoliva (changes requested):

PR #1075, opened 2026-05-22
fix(telemetry): don't mark QueueShutDown as an error span

### LINKED ISSUE #1065: trace_function: add QueueShutDown to CancelledError-style graceful-exception allowlist
## Summary

``a2a/utils/telemetry.py``'s ``trace_function`` async wrapper (lines ~218-258 on v1.0.3) special-cases ``asyncio.CancelledError`` — records the exception event but does NOT call ``span.set_status(StatusCode.ERROR, …)``. Suggest extending that allowlist to include ``asyncio.QueueShutDown`` (and the a2a back-port shim ``AsyncQueueShutDown`` from ``a2a/server/events/event_queue.py``).

## Why

``EventQueueSource.dequeue_event`` is decorated with ``@trace_class``. On normal task end the producer ``shutdown()``s the underlying queue, the consumer's pending ``dequeue_event()`` raises ``QueueShutDown``, and ``trace_function``'s generic ``except Exception`` arm paints the span ``StatusCode.ERROR``. The consumer (``ActiveTaskEventConsumer.run``, ``a2a/server/agent_execution/active_task.py:~134``) is *designed* to catch ``QueueShutDown`` and log it at DEBUG — this is normal teardown, not an error.

Real-world impact: in a fleet of A2A agents using LangSmith OTel + Langfuse, ≥81 % of ``StatusCode.ERROR`` observations over a 7-day window were this single false alarm. Legitimate ``TaskNotFoundError`` / ``InvalidParamsError`` / ``VersionNotSupportedError`` errors are drowned in the noise.

## Proposed change

Treat ``QueueShutDown`` / ``AsyncQueueShutDown`` the same way ``CancelledError`` is treated today (record exception, leave span status alone, re-raise). One small allowlist tuple in ``trace_function``.

## Workaround we're shipping in the meantime

Custom ``SpanProcessor`` that drops the span before it reaches the OTLP exporter when ``span.name`` is on a small allowlist (today just ``EventQueueSource.dequeue_event``) AND the recorded exception is ``QueueShutDown``-class. Lives in our gideon library (42-com/gideon ``src/gideon/observability.py``); see https://github.com/42-com/gideon/pull/<PR-PENDING> for the implementation.

Happy to open a PR against ``a2a-python`` mirroring the ``CancelledError`` pattern if maintainers are receptive.

## References

* Verified against a2a-sdk v1.0.3 ``a2a/utils/telemetry.py`` + ``a2a/server/events/event_queue_v2.py`` + ``a2a/server/events/event_queue.py``.
* No existing open issue references this telemetry behavior (the two closed ``QueueShutDown`` items #46 and PR #407 are about Python 3.13 ``asyncio.Queue.shutdown`` portability, not telemetry).

sokoliva:
# Description
## Summary
`trace_function`'s async wrapper marks any exception other than `CancelledError` as `StatusCode.ERROR`. This wrongly flags `QueueShutDown`, the normal signal raised when a producer cleanly closes an event queue, as a failure.
## Change
- Add `QueueShutDown` to the same graceful-exception allowlist as`CancelledError` in `src/a2a/utils/telemetry.py`. The class is resolved per Python version (stdlib on 3.13+, `culsans` back-port on <3.13), mirroring `event_queue.py`.
- Add parametrized a test covering both `CancelledError` and `QueueShutDown`.

Fixes #1065🦕

### DIFF
--- a/src/a2a/utils/telemetry.py
+++ b/src/a2a/utils/telemetry.py
@@ -68,12 +68,13 @@ import functools
 import inspect
 import logging
 import os
-import sys
 
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
 from typing_extensions import Self
+
+from a2a.utils._async_queue_compat import QueueShutDown
 
 
 if TYPE_CHECKING:
@@ -145,14 +146,9 @@ SpanKind = _SpanKind
 __all__ = ['SpanKind']
 
 
-if sys.version_info >= (3, 13):
-    from asyncio import QueueShutDown as _QueueShutDown
-else:
-    from culsans import AsyncQueueShutDown as _QueueShutDown
-
-_GRACEFUL_ASYNC_EXCEPTIONS: tuple[type[BaseException], ...] = (
+_NON_ERROR_EXCEPTIONS: tuple[type[BaseException], ...] = (
     asyncio.CancelledError,
-    _QueueShutDown,
+    QueueShutDown,
 )
 
 
@@ -245,9 +241,7 @@ def trace_function(  # noqa: PLR0915
                 # Async wrapper, await for the function call to complete.
                 result = await func(*args, **kwargs)
                 span.set_status(StatusCode.OK)
-            # Graceful end-of-operation signals (currently asyncio.CancelledError,
-            # QueueShutDown).
-            except _GRACEFUL_ASYNC_EXCEPTIONS as ge:
+            except _NON_ERROR_EXCEPTIONS as ge:
                 exception = None
                 logger.debug(
                     '%s in span %s',
--- a/src/a2a/utils/telemetry.py
+++ b/src/a2a/utils/telemetry.py
@@ -68,6 +68,7 @@ import functools
 import inspect
 import logging
 import os
+import sys
 
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
@@ -142,6 +143,17 @@ if not otel_installed or not otel_enabled:
 
 SpanKind = _SpanKind
 __all__ = ['SpanKind']
+
+
+if sys.version_info >= (3, 13):
+    from asyncio import QueueShutDown as _QueueShutDown
+else:
+    from culsans import AsyncQueueShutDown as _QueueShutDown
+
+_GRACEFUL_ASYNC_EXCEPTIONS: tuple[type[BaseException], ...] = (
+    asyncio.CancelledError,
+    _QueueShutDown,
+)
 
 
 def trace_function(  # noqa: PLR0915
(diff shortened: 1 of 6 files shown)


src/a2a/utils/telemetry.py
```diff
@@ -144,6 +145,17 @@ def __getattr__(self, name: str) -> Any:
...
 
+if sys.version_info >= (3, 13):
+    from asyncio import QueueShutDown as _QueueShutDown
+else:
+    from culsans import AsyncQueueShutDown as _QueueShutDown
+
+_GRACEFUL_ASYNC_EXCEPTIONS: tuple[type[BaseException], ...] = (
+    asyncio.CancelledError,
+    _QueueShutDown,
+)
```
ishymko: Given that we have the same check in `event_queue.py` wdyt about extracting it in an internal utility module (i.e. `_async_queue_compat.py`) and share it between two?

src/a2a/utils/telemetry.py
```diff
@@ -233,11 +245,16 @@ async def async_wrapper(*args, **kwargs) -> Any:
                 # Async wrapper, await for the function call to complete.
                 result = await func(*args, **kwargs)
                 span.set_status(StatusCode.OK)
-            # asyncio.CancelledError extends from BaseException
-            except asyncio.CancelledError as ce:
+            # Graceful end-of-operation signals (currently asyncio.CancelledError,
+            # QueueShutDown).
```
ishymko: nit: I'd say this comment is not needed here and the list itself should be used as a source of truth (as we don't want to update this comment for each new item added into the list)

src/a2a/utils/telemetry.py
```diff
@@ -233,11 +245,16 @@ async def async_wrapper(*args, **kwargs) -> Any:
...
+            except _GRACEFUL_ASYNC_EXCEPTIONS as ge:
```
ishymko: nit: `_NON_ERROR_EXCEPTIONS`?

src/a2a/utils/telemetry.py
sokoliva: Done. I extracted 

```
if sys.version_info >= (3, 13):
    from asyncio import Queue as AsyncQueue
    from asyncio import QueueShutDown

    def create_async_queue(maxsize: int = 0) -> AsyncQueue[Any]:
        """Create a backwards-compatible async queue object."""
        return AsyncQueue(maxsize=maxsize)
else:
    import culsans

    from culsans import AsyncQueue  # type: ignore[no-redef]
    from culsans import (
        AsyncQueueShutDown as QueueShutDown,  # type: ignore[no-redef]
    )

    def create_async_queue(maxsize: int = 0) -> AsyncQueue[Any]:
        """Create a backwards-compatible async queue object."""
        return culsans.Queue(maxsize=maxsize).async_q  # type: ignore[no-any-return]
```

from event_queue.py because by extracting just from telemtry.py we would still have code duplication regarding if `sys.version_info >= (3, 13):` check

ishymko (approved):

PR #1077, opened 2026-05-24
fix: raise on append=True for unknown artifact_id (#1038)

### LINKED ISSUE #1038: [Bug]: append=True with unknown artifact_id silently drops the chunk — should fail loudly
### What happened?

When `TaskUpdater.add_artifact(append=True, ...)` is called with an `artifact_id` that doesn't exist on the task yet, the chunk is silently dropped after a `logger.warning`. This is a footgun — the wire delivers a successful response while the streamed content is lost.

#### Reproducer (concept)

```python
from uuid import uuid4
from a2a.types.a2a_pb2 import Part

# Inside an AgentExecutor.execute(...):
async for chunk in some_async_source():
    await updater.add_artifact(
        # No artifact_id — TaskUpdater auto-generates a fresh one on
        # EVERY call. Each call hits the warning path because the
        # newly-generated id has no prior artifact to append to.
        parts=[Part(text=chunk)],
        name="my_response",
        append=True,
        last_chunk=False,
    )
```

Each `add_artifact` call produces:

```
WARNING  a2a.server.tasks.task_manager:task_manager.py:79 Received append=True for nonexistent artifact index <fresh-uuid> in task <task-id>. Ignoring chunk.
```

…and the chunk is dropped. The Task transitions through WORKING → COMPLETED with an empty artifact, no error surfaces upward, and the client sees a successful response with no content.

The fix is to pin `artifact_id = uuid4().hex` once per turn and pass it on every `add_artifact` call — but you only learn this from reading the SDK source or finding a worked example. The first time we hit this, the only signal was a downstream test that asserted on artifact text; if the test had only asserted HTTP status we'd have shipped the regression.

#### Suggested fix

Surface the failure path loudly. Two options, ranked by invasiveness:

1. **Raise instead of warn** in [`task_manager.py:append_artifact_to_task`](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/task_manager.py#L79). A `ValueError` (or a dedicated subclass) makes the misuse impossible to miss. Existing well-formed callers see no change; misusers see a clear stack trace pointing at their `add_artifact` call instead of a silently-empty response.

2. **Or**: have `TaskUpdater.add_artifact` itself remember the most-recently-allocated id when `artifact_id is None`, so a sequence of `add_artifact(append=True, ...)` calls without explicit ids implicitly target the same artifact. This changes default behaviour but matches the natural mental model of "I'm streaming chunks of one thing."

Option 1 is the minimal/safest change. Option 2 is more ergonomic but is a behaviour change, so might want a deprecation cycle.

#### Why it matters

Streaming `artifact_update` events is the main reason a server sets `capabilities.streaming=True`. The current SDK behaviour means the streaming path is a footgun for any server-side executor that doesn't already know the `artifact_id`-pinning trick. Loud failure beats silent drop.

Happy to send a PR for option 1 if maintainers agree on the approach.

#### Environment

- a2a-sdk 1.0.2
- Python 3.12.4

### Relevant log output

```shell
WARNING  a2a.server.tasks.task_manager:task_manager.py:79 Received append=True for nonexistent artifact index 20ba23dd-68e8-4f62-9ff0-42f9c9fa79e4 in task 87481320-93bd-456d-add0-2ccc6fe8e48b. Ignoring chunk.
WARNING  a2a.server.tasks.task_manager:task_manager.py:79 Received append=True for nonexistent artifact index c4569d04-6a52-493a-931d-a70725368721 in task 87481320-93bd-456d-add0-2ccc6fe8e48b. Ignoring chunk.
```

Linux2010:
## Problem
`TaskUpdater.add_artifact(append=True, ...)` with an unknown `artifact_id` logs a warning and drops the chunk. The wire returns success while the streamed content is lost.

## Fix
Raise instead of logging and dropping, with a message pointing the caller to create the artifact first (`append=False`).

## Tests
Updated `test_append_artifact_to_task` to assert `ValueError` is raised and task artifacts are unchanged after the failed append.

Fixes #1038

### DIFF
--- a/src/a2a/server/tasks/task_manager.py
+++ b/src/a2a/server/tasks/task_manager.py
@@ -12,7 +12,7 @@ from a2a.types.a2a_pb2 import (
     TaskStatus,
     TaskStatusUpdateEvent,
 )
-from a2a.utils.errors import InvalidParamsError
+from a2a.utils.errors import InvalidAgentResponseError, InvalidParamsError
 from a2a.utils.telemetry import trace_function
 
 
@@ -74,12 +74,15 @@ def append_artifact_to_task(task: Task, event: TaskArtifactUpdateEvent) -> None:
             dict(new_artifact_data.metadata.items())
         )
     else:
-        # We received a chunk to append, but we don't have an existing artifact.
-        # we will ignore this chunk
-        logger.warning(
-            'Received append=True for nonexistent artifact index %s in task %s. Ignoring chunk.',
-            artifact_id,
-            task.id,
+        # A2A spec (Section 4.22 TaskArtifactUpdateEvent) states:
+        #   "append — If true, the content of this artifact should be
+        #    appended to a previously sent artifact."
+        # append=True with no prior artifact of this ID is an agent
+        # protocol violation (#1038).
+        raise InvalidAgentResponseError(
+            f'append=True for nonexistent artifact_id={artifact_id!r} '
+            f'in task {task.id!r}. The artifact must be created (append=False) '
+            f'before appending parts to it.'
         )
 
 
--- a/src/a2a/server/tasks/task_manager.py
+++ b/src/a2a/server/tasks/task_manager.py
@@ -74,7 +74,8 @@ def append_artifact_to_task(task: Task, event: TaskArtifactUpdateEvent) -> None:
             dict(new_artifact_data.metadata.items())
         )
     else:
-        # A2A spec (Section 4.22 TaskArtifactUpdateEvent) states:
+        # A2A spec (Section 4.2.2 TaskArtifactUpdateEvent,
+        # https://a2a-protocol.org/v1.0.0/specification/#422-taskartifactupdateevent):
         #   "append — If true, the content of this artifact should be
         #    appended to a previously sent artifact."
         # append=True with no prior artifact of this ID is an agent
--- a/tests/server/tasks/test_task_manager.py
+++ b/tests/server/tasks/test_task_manager.py
@@ -439,5 +439,8 @@ def test_append_artifact_to_task():
         task_id='123',
         context_id='123',
     )
-    with pytest.raises(InvalidAgentResponseError, match='append=True for nonexistent artifact_id'):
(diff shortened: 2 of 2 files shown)


src/a2a/server/tasks/task_manager.py
```diff
@@ -75,11 +75,12 @@ def append_artifact_to_task(task: Task, event: TaskArtifactUpdateEvent) -> None:
...
     else:
         # We received a chunk to append, but we don't have an existing artifact.
-        # we will ignore this chunk
-        logger.warning(
-            'Received append=True for nonexistent artifact index %s in task %s. Ignoring chunk.',
-            artifact_id,
-            task.id,
+        # Raising an error ensures the agent knows the append failed, rather
+        # than silently dropping the chunk (#1038).
+        raise ValueError(
```
ishymko: Let's make it `InvalidAgentResponseError`, as it's basically agent implementation not following [the spec](https://a2a-protocol.org/latest/specification/#422-taskartifactupdateevent):

> `append` - If true, the content of this artifact should be appended to **a previously sent artifact**

src/a2a/server/tasks/task_manager.py
```diff
@@ -75,11 +75,12 @@ def append_artifact_to_task(task: Task, event: TaskArtifactUpdateEvent) -> None:
         )
     else:
         # We received a chunk to append, but we don't have an existing artifact.
-        # we will ignore this chunk
-        logger.warning(
-            'Received append=True for nonexistent artifact index %s in task %s. Ignoring chunk.',
-            artifact_id,
-            task.id,
+        # Raising an error ensures the agent knows the append failed, rather
+        # than silently dropping the chunk (#1038).
```
ishymko: Let's quote the spec here instead, current comment duplicates the code below.

ishymko (commented):
Thanks for the PR!

note: [linter](https://github.com/a2aproject/a2a-python/actions/runs/26357349649?pr=1077) is failing

ishymko (approved):

PR #1082, opened 2026-05-27
fix: distinguish graceful EventQueue shutdown

AXEG0:
## Summary
- add QueueShutDownNormal for graceful EventQueueSink shutdowns
- map graceful sink dequeue shutdowns to the narrower subclass
- keep immediate shutdown and closed enqueue/tap paths on the generic QueueShutDown signal

## Motivation
The v2 queue currently uses the same QueueShutDown class for graceful stream teardown and abnormal lifecycle paths such as immediate close or attempts to use a closed queue. Telemetry wrappers and consumers cannot distinguish those cases after the generic exception is raised.

## Validation
- UV_CACHE_DIR=/var/tmp/argylle/tmp/task-1779877269025068489-2f550c95/uv-cache UV_PYTHON_INSTALL_DIR=/var/tmp/argylle/tmp/task-1779877269025068489-2f550c95/uv-python uv run pytest tests/server/events/test_event_queue_v2.py -q

### DIFF
--- a/src/a2a/server/events/event_queue_v2.py
+++ b/src/a2a/server/events/event_queue_v2.py
@@ -20,6 +20,10 @@ from a2a.utils.telemetry import SpanKind, trace_class
 logger = logging.getLogger(__name__)
 
 
+class QueueShutDownNormal(QueueShutDown):
+    """Raised when a sink is closed by a graceful queue shutdown."""
+
+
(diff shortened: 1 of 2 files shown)


ishymko:
Thank you @AXEG0, this was done in #1075.

PR #1024, opened 2026-04-27
feat(server): restore FastAPI /docs visibility for A2A routes

martimfasantos:
## Summary

- Introduces `src/a2a/server/routes/helpers/` — a new subpackage of transport-specific route helpers mirroring the layout of `client/transports/`
- `add_a2a_routes_to_fastapi()` re-registers A2A Starlette routes as `APIRoute` instances so FastAPI's OpenAPI generator picks them up and they appear in `/docs` and `/openapi.json`
- Each endpoint group is tagged (`A2A: Agent Card`, `A2A: JSON-RPC`, `A2A: REST`) and annotated with proto-derived request-body schemas so Swagger UI's **Try it out** panel renders a typed, editable payload
- The required `A2A-Version: 1.0` header is declared on every dispatcher route so Swagger pre-fills it and callers aren't blocked by the `@validate_version` guard

## Why

In v0.3, `A2AFastAPIApplication` was a first-class FastAPI subclass, so all endpoints were automatically visible in `/docs` (see [#280](https://github.com/a2aproject/a2a-python/pull/280)).

The v1.0 rewrite dropped that wrapper in favour of bare Starlette route factories (`create_agent_card_routes`, `create_jsonrpc_routes`, `create_rest_routes`). These return `starlette.routing.Route` objects; FastAPI's OpenAPI generator only enumerates `fastapi.routing.APIRoute` instances, so every A2A endpoint silently vanished from `/docs` and `/openapi.json`.

This PR restores that capability without requiring callers to switch frameworks or restructure their apps — a single helper function wraps and enriches the routes in place.

Additional improvements over the v0.3 approach:
- **Rich request-body schemas** generated from proto descriptors (`google.protobuf.descriptor`) rather than Pydantic models, matching the proto-first v1.0 types
- **Proto `oneof` → JSON Schema `oneOf`** so `Part.content` variants (`text`, `raw`, `url`, `data`) are correctly mutually exclusive in Swagger — prevents `-32602 Invalid params` errors from the auto-generated sample payload
- **JSON-RPC `params` as `oneOf`** over all 11 method types — every method is individually inspectable from a single endpoint row
- **`A2A-Version` header** pre-filled to `1.0` on all dispatcher routes — removes the `-32009` version error that would otherwise block every "Try it out" call
- **Scalable layout** — `helpers/` mirrors `client/transports/` with one file per integration target; adding a new framework is a single new file that imports from `_proto_schema.py` and `jsonrpc.py`

## Changes

```
[...]

PR #1092, opened 2026-06-01
fix(server): improve Swagger /docs example annotations for A2A routes

martimfasantos:
## Summary

- Uses `add_a2a_routes_to_fastapi()` in the sample instead of `app.routes.extend()` — the feature introduced in #1024 was not reflected in the sample
- Adds principled Swagger example annotations to the proto → JSON Schema generator (`_proto_schema.py`), driven by `google.api.field_behavior` proto annotations and proto type information rather than hardcoded field-name lists

### Schema example improvements

| Field type | Before | After |
|---|---|---|
| Enum | no example → Swagger guesses (e.g. `"user"`) causing `-32602` | first non-`UNSPECIFIED` value (e.g. `ROLE_USER`) |
| String, optional | no example → Swagger generates `"string"` causing task-lookup failures on `task_id` | `""` (proto3 empty = field unset) |
| String, `REQUIRED` | no example → `"string"` causing validation failures on `message_id` | `field.name` (non-empty, self-describing) |
| Bool | no example → Swagger picks `true` causing `return_immediately: true` → `SUBMITTED` | `false` |
| Message, non-`REQUIRED` | expanded inline → optional config sub-fields fail required-field validation | `oneOf: [$ref, null]` with `example: null` |
| Message, `REQUIRED` | — | unchanged `$ref` (not nullable) |
| Repeated (array) | no array example → Swagger generates one item per `oneOf` branch | `[item_example]` (single pre-filled entry) |
| oneOf schema | no example → all branches expanded in arrays | example using first variant only |

Optionality is derived from `google.api.field_behavior = REQUIRED` annotations already present on the A2A proto fields — no hardcoded field-name exceptions.

## Checklist

- [x] Follows [`CONTRIBUTING` Guide](https://github.com/a2aproject/a2a-python/blob/main/CONTRIBUTING.md)
- [x] PR title follows Conventional Commits
- [x] Tests pass — 95 passing (8 new tests covering the new example behaviour)
- [x] `ruff format` and `ruff check` clean

Continues #1024 🦕

### DIFF
--- a/src/a2a/server/routes/_proto_schema.py
+++ b/src/a2a/server/routes/_proto_schema.py
@@ -2,6 +2,7 @@
 
 from typing import Any
 
+from google.api import field_behavior_pb2 as _fb
 from google.protobuf.descriptor import Descriptor, FieldDescriptor
 from google.protobuf.message import Message
 
@@ -33,6 +34,15 @@ _PROTO_SCALAR_SCHEMAS: dict[int, dict[str, Any]] = {
     FieldDescriptor.TYPE_SINT64: {'type': 'string'},
 }
 
+
+def _is_required(field: FieldDescriptor) -> bool:
+    """Returns True if the field carries google.api.field_behavior = REQUIRED."""
+    try:
+        return _fb.REQUIRED in field.GetOptions().Extensions[_fb.field_behavior]
+    except KeyError:
+        return False
+
+
 _WELL_KNOWN_SCHEMAS: dict[str, dict[str, Any]] = {
     'google.protobuf.Timestamp': {'type': 'string', 'format': 'date-time'},
     'google.protobuf.Duration': {'type': 'string'},
@@ -57,16 +67,44 @@ def field_schema(
 
     if field.type == FieldDescriptor.TYPE_MESSAGE:
         item = message_schema(field.message_type, components)
+        # Well-known types return an inline schema (no $ref); don't wrap them as
+        # nullable — they're already inlined as their JSON-Schema equivalent.
+        if not _is_required(field) and '$ref' in item:
+            return {'oneOf': [item, {'type': 'null'}], 'example': None}
     elif field.type == FieldDescriptor.TYPE_ENUM:
-        item = {
-            'type': 'string',
-            'enum': [v.name for v in field.enum_type.values],
-        }
+        values = [v.name for v in field.enum_type.values]
+        example = next(
+            (
+                v
+                for v in values
+                if 'UNSPECIFIED' not in v and 'UNKNOWN' not in v
+            ),
+            values[0] if values else None,
+        )
+        item: dict[str, Any] = {'type': 'string', 'enum': values}
+        if example:
+            item['example'] = example
     else:
         item = dict(_PROTO_SCALAR_SCHEMAS.get(field.type, {'type': 'string'}))
+        if field.type == FieldDescriptor.TYPE_STRING:
+            # REQUIRED fields must be non-empty; use the field name as a
+            # recognisable placeholder. All other strings default to "".
+            item['example'] = field.name if _is_required(field) else ''
+        elif field.type == FieldDescriptor.TYPE_BOOL:
+            item['example'] = False
 
     if field.is_repeated:
-        return {'type': 'array', 'items': item}
+        array_schema: dict[str, Any] = {'type': 'array', 'items': item}
+        # Propagate the item example to the array so Swagger pre-fills one entry
+        # instead of generating one entry per oneOf branch.
+        item_example = (
+            components.get(item['$ref'].split('/')[-1], {}).get('example')
+            if '$ref' in item
+            else item.get('example')
+        )
+        if item_example is not None:
+            array_schema['example'] = [item_example]
+        return array_schema
     return item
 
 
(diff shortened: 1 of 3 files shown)


src/a2a/server/routes/_proto_schema.py
```diff
@@ -57,16 +67,44 @@ def field_schema(
 
     if field.type == FieldDescriptor.TYPE_MESSAGE:
         item = message_schema(field.message_type, components)
+        # Well-known types return an inline schema (no $ref); don't wrap them as
+        # nullable — they're already inlined as their JSON-Schema equivalent.
+        if not _is_required(field) and '$ref' in item:
+            return {'oneOf': [item, {'type': 'null'}], 'example': None}
```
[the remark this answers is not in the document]
martimfasantos: Fixed in f5fa47f. Guarded the nullable wrapping with `not field.is_repeated` so repeated non-REQUIRED message fields fall through to the array-wrapping block instead of returning early as a nullable oneOf.

One note: `field.label != FieldDescriptor.LABEL_REPEATED` triggers a deprecation warning in modern protobuf Python — `label()` is deprecated in favour of `is_repeated` — so I used `not field.is_repeated` instead, which is the currently idiomatic form and avoids the warning.

src/a2a/server/routes/_proto_schema.py
```diff
@@ -114,5 +152,14 @@ def message_schema(
...
-    components[name] = parts[0] if len(parts) == 1 else {'allOf': parts}
+    schema: dict[str, Any] = parts[0] if len(parts) == 1 else {'allOf': parts}
+    # Provide a single concrete example using the first oneof variant so Swagger
+    # doesn't expand every branch into separate array items.
+    first_oneof_field = real_oneofs[0].fields[0]
+    schema['example'] = {
+        first_oneof_field.name: first_oneof_field.name
+        if _is_required(first_oneof_field)
+        else ''
+    }
```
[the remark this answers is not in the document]
martimfasantos: Fixed in f5fa47f. The example value is now derived dynamically: call `field_schema` on the first oneof field, use its `example` if present, resolve the example through a `$ref` from components, or fall back to a type-appropriate default (0 for integers, False for booleans, {} for objects, etc.).

src/a2a/server/routes/_proto_schema.py
```diff
@@ -2,6 +2,7 @@
 
 from typing import Any
 
+from google.api import field_behavior_pb2 as _fb
```
ishymko: nit: I'd just do `as fb`:

```suggestion
from google.api import field_behavior_pb2 as fb
```

The entire `_proto_schema.py` is already marked as internal and also it's not an established practice in the repo.

tests/server/routes/test_proto_schema.py
```diff
@@ -120,6 +120,185 @@ def test_field_schema_enum():
...
+def test_field_schema_required_message_is_not_nullable():
+    from a2a.types.a2a_pb2 import SendMessageRequest
+
+    field = SendMessageRequest.DESCRIPTOR.fields_by_name['message']
+    schema = field_schema(field, {})
+    assert '$ref' in schema
+    assert 'oneOf' not in schema
+
+
+def test_field_schema_repeated_optional_message_is_array_not_nullable():
```
ishymko, line 132: Can we use `Task.history` here instead of mocking? Based on the test description it should suit it: repeated and non-required.

src/a2a/server/routes/_proto_schema.py
```diff
@@ -33,6 +34,15 @@
     FieldDescriptor.TYPE_SINT64: {'type': 'string'},
 }
 
+
+def _is_required(field: FieldDescriptor) -> bool:
+    """Returns True if the field carries google.api.field_behavior = REQUIRED."""
+    try:
+        return _fb.REQUIRED in field.GetOptions().Extensions[_fb.field_behavior]  # type: ignore[index]  # ty: ignore[invalid-argument-type]
+    except KeyError:
+        return False
```
ishymko: I feel like this is redundant as `field_behavior` is repeated so `Extensions[field_behavior]` should just give an empty list when it's not there. However dropping this fails tests, which I believe are not real failures as they use mocks.

Let's try to avoid mocks for such tests, if we don't have a certain structure across current protobufs I'd just skip testing this. We should just make sure we invoke building the entire schema in tests so that in case of changes we don't support it fails right there.

src/a2a/server/routes/_proto_schema.py
martimfasantos: Done in dc7f066 — imported as fb.

tests/server/routes/test_proto_schema.py
martimfasantos: Done in dc7f066 — the test now uses Task.history (a real repeated, non-required message field) instead of a mock.

src/a2a/server/routes/_proto_schema.py
martimfasantos: Done in dc7f066 — confirmed Extensions[field_behavior] returns an empty list (never KeyError) on real descriptors, so I dropped the try/except; the mock-only synthetic tests are gone and a full-schema-build test now exercises every REST body type so unsupported proto changes fail right there.

ishymko (approved):
Thank you for improving this!

PR #1053, opened 2026-05-06
refactor: Enqueue exceptions in ActiveTask

bartek-w:
# Description
Remove ActiveTask._exception. Make exception handling fully synchronized in the queues.

Fixes #1032 🦕

PR #1105, opened 2026-06-14
feat(server): add aclose() to drain ActiveTask background tasks (#1101)

### LINKED ISSUE #1101: [Bug]: DefaultRequestHandlerV2 ActiveTask producer can remain pending during EventQueue subscriber cleanup
### What happened?

We are using `a2a-sdk==1.1.0` on Python 3.12 with the standard `DefaultRequestHandler`, which resolves to `DefaultRequestHandlerV2`.

Under a concurrent A2A request scenario, one request was rejected/finished at the application layer and persisted as a terminal task state, but the SDK-created background producer task did not fully complete. Later, when the event loop / worker was shutting down, asyncio reported:

`Task was destroyed but it is pending!`

The pending task was the SDK `ActiveTask._run_producer()` task named `producer:<task_id>`. The coroutine was stuck in the producer `finally` block at:

```python
await self._event_queue_subscribers.close(immediate=False)
```

From reading the SDK code, EventQueueSource.close(immediate=False) waits for already queued events and child sinks to drain. If a subscriber/tapped sink is no longer being consumed, this can leave the producer cleanup pending indefinitely.

Expected behavior: DefaultRequestHandlerV2 / ActiveTask should not leave SDK-owned producer tasks pending after the request has already reached a terminal application state. Cleanup of subscriber queues should either complete reliably or be bounded/forced in a way that does not leave a background task alive until event loop destruction.

This also causes confusing observability because the asyncio warning is emitted later from the event loop, not from the original request context.

### Relevant log output

```shell
level: ERROR
logger_name: asyncio
message: Task was destroyed but it is pending!
task: <Task pending name='producer:task-REDACTED'
  coro=<ActiveTask._run_producer() running at
  /path/to/.venv/lib/python3.12/site-packages/a2a/server/agent_execution/active_task.py:566>
  wait_for=<_GatheringFuture pending cb=[Task.task_wakeup()]>>


Relevant SDK location:


# a2a/server/agent_execution/active_task.py

finally:
    self._request_queue.shutdown(immediate=True)
    await self._event_queue_agent.close(immediate=False)
    await self._event_queue_subscribers.close(immediate=False)
    logger.debug('Producer[%s]: Completed', self._task_id)


Environment:


Python: 3.12
a2a-sdk: 1.1.0
Request handler: a2a.server.request_handlers.default_request_handler_v2.DefaultRequestHandlerV2
Transport: HTTP / streaming A2A endpoint
```

### Code of Conduct

astrogilda:
## Summary

Adds a public `aclose()` teardown to `ActiveTask`, `ActiveTaskRegistry`, and
`DefaultRequestHandlerV2` that force-drains the SDK-owned producer, consumer,
and dispatcher `asyncio.Task`s so none are left pending at event-loop shutdown.

- `ActiveTask.aclose()` force-closes both event queues and cancels the producer
  and consumer tasks, then awaits them. It sets `_is_finished` under `_lock`, so
  it is mutually exclusive with `start()` (which refuses to spawn once
  `_is_finished` is set).
- `ActiveTaskRegistry.aclose()` marks the registry closed so `get_or_create`
  refuses new work, drains every active task, then awaits the in-flight
  `_remove_task` cleanup tasks. The lock is released before awaiting because
  `_remove_task` re-acquires it.
- `DefaultRequestHandlerV2.aclose()` delegates to the registry drain, for wiring
  into an ASGI `lifespan` / `on_shutdown` hook.

## Why

Fixes #1101. At shutdown the `ActiveTask` producer can stay pending and surface
as `Task was destroyed but it is pending!`. The producer's `finally` calls
`_event_queue_subscribers.close(immediate=False)`, which awaits `join()` on
every subscriber sink; an abandoned subscriber leaves an undrained sink, so the
`join()` never returns and the producer hangs. There is no public way to drain
these background tasks today. `aclose()` closes the queues with `immediate=True`,
which releases the wedged producer, and reaps the tasks.

The teardown always forces rather than exposing a graceful `immediate=False`
option, because that path inherits the documented `close(immediate=False)`
deadlock and a shutdown hook must be bounded.

## Test plan

- [x] `uv run pytest tests/server/agent_execution/ tests/server/request_handlers/ tests/server/events/` — pass
- [x] `uv run pytest --cov=a2a --cov-fail-under=88` — pass
- [x] `./scripts/lint.sh` — ruff, ruff-format, and ty clean

New tests cover the registry drain, idempotency, empty registry, new-work
rejected after close, and an errored task being logged rather than propagated;
`ActiveTask` reaping a running producer and force-closing past an undrained
subscriber (the #1101 repro); and the handler drain.

Fixes #1101 🦕

### DIFF
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -768,6 +768,11 @@ class ActiveTask:
         background tasks are cancelled, so teardown is bounded even when a
         subscriber sink was never drained. It is safe to call multiple times.
         """
+        # Shut down the request queue first, mirroring the producer's
+        # ``finally``. If `start()` was never called, the producer/consumer
+        # ``finally`` blocks never run, and without this a caller parked in
+        # `enqueue_request()` would wait forever.
+        self._request_queue.shutdown(immediate=True)
         await self._event_queue_agent.close(immediate=True)
         await self._event_queue_subscribers.close(immediate=True)
         # Set `_is_finished` and collect the background tasks under `_lock` so
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -755,6 +755,43 @@ class ActiveTask:
             raise RuntimeError('Task should have been created')
         return task
 
+    async def aclose(self) -> None:
+        """Force-closes the task's queues and drains its background tasks.
+
+        Provides a bounded, public teardown for the producer and consumer
+        ``asyncio.Task``s spawned in ``start()``. Without it, a producer
+        wedged in its ``finally`` closing an abandoned subscriber sink can
+        survive until event-loop shutdown and surface as
+        ``Task was destroyed but it is pending!``.
+
+        Always forces: the queues are closed with ``immediate=True`` and the
+        background tasks are cancelled, so teardown is bounded even when a
+        subscriber sink was never drained. It is safe to call multiple times.
+        """
+        await self._event_queue_agent.close(immediate=True)
+        await self._event_queue_subscribers.close(immediate=True)
+        # Set `_is_finished` and collect the background tasks under `_lock` so
+        # this is mutually exclusive with `start()`, which refuses to spawn
+        # once `_is_finished` is set. The lock is released before awaiting the
+        # tasks, because their teardown re-acquires it.
+        async with self._lock:
+            self._is_finished.set()
+            background_tasks = [
+                task
+                for task in (self._producer_task, self._consumer_task)
+                if task is not None
+            ]
+            for task in background_tasks:
+                task.cancel()
+        if background_tasks:
+            results = await asyncio.gather(
+                *background_tasks, return_exceptions=True
+            )
+            for result in results:
+                # CancelledError is a BaseException, so it is excluded here.
+                if isinstance(result, Exception):
+                    logger.error('Error during aclose', exc_info=result)
+
     async def _maybe_cleanup(self) -> None:
         """Triggers cleanup if task is finished and has no subscribers.
 
--- a/src/a2a/server/agent_execution/active_task_registry.py
+++ b/src/a2a/server/agent_execution/active_task_registry.py
@@ -35,6 +35,7 @@ class ActiveTaskRegistry:
         self._active_tasks: dict[str, ActiveTask] = {}
         self._lock = asyncio.Lock()
         self._cleanup_tasks: set[asyncio.Task[None]] = set()
+        self._closed = False
 
     async def get_or_create(
         self,
@@ -46,6 +47,8 @@ class ActiveTaskRegistry:
     ) -> ActiveTask:
         """Retrieves an existing ActiveTask or creates a new one."""
         async with self._lock:
+            if self._closed:
+                raise RuntimeError('ActiveTaskRegistry is closed')
             if task_id in self._active_tasks:
                 return self._active_tasks[task_id]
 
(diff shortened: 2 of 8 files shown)


src/a2a/server/agent_execution/active_task.py
```diff
@@ -729,6 +729,43 @@ async def cancel(self, call_context: ServerCallContext) -> Task:
...
+        wedged in its ``finally`` closing an abandoned subscriber sink can
+        survive until event-loop shutdown and surface as
+        ``Task was destroyed but it is pending!``.
+
+        Always forces: the queues are closed with ``immediate=True`` and the
+        background tasks are cancelled, so teardown is bounded even when a
+        subscriber sink was never drained. It is safe to call multiple times.
+        """
+        await self._event_queue_agent.close(immediate=True)
+        await self._event_queue_subscribers.close(immediate=True)
```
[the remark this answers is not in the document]
astrogilda: Good catch. If start() is never called, the producer/consumer finally blocks (the only other places this queue gets shut down) never run, so a caller parked in enqueue_request() would just sit there forever. Added self._request_queue.shutdown(immediate=True) as the first thing aclose() does, mirroring the producer's finally ordering, plus a test for the never-started case.

astrogilda:
Addressed the review suggestion: aclose() now shuts down _request_queue (immediate=True) first, covering the never-started case, with a test. CI is green on the new push. Ready for review whenever you have a moment @rohityan.

src/a2a/server/agent_execution/active_task.py
```diff
@@ -729,6 +729,48 @@ async def cancel(self, call_context: ServerCallContext) -> Task:
             raise RuntimeError('Task should have been created')
         return task
 
+    async def aclose(self) -> None:
```
mykytanetipa, line 758: minor: The ActiveTask.aclose() drains producer/consumer but not the two sources _dispatcher_tasks (EventQueueSource.close(immediate=True) cancels each dispatcher without awaiting it). 

For _event_queue_subscribers (create_default_sink=False) when aclose() cleans up a task that has no subscribers, there are no sinks to close, so the gather has nothing to await and returns instantly without ever letting the cancelled dispatcher run -> the dispatcher is still pending when aclose() returns -> Task was destroyed but it is pending.

Consider handling this in aclose() by awaiting the sources dispatcher tasks alongside the producer/consumer.

astrogilda:
Rebased onto current main to clear the "behind" state after #1053 and the recent refactors. No content changes, and the full suite is green on the rebased branch.

Quick recap of where things stand: the approach was agreed in #1101 (thanks @rohityan), the one review finding (never-started request queue) is addressed with a test, and CI has been green since the July push. Ready for review whenever you have a moment.

mykytanetipa (changes requested):

src/a2a/server/agent_execution/active_task.py
astrogilda: You're right, and the reproduction is exactly as you describe. close(immediate=True) cancelled _dispatcher_task without awaiting it, and in the case with no sinks (create_default_sink=False, zero subscribers) the gather had nothing to await. aclose() then returned before the cancelled dispatcher ever got to run, hence the pending-task warning at loop shutdown. Fixed in 0728696, pushed on the 17th; after today's rebase onto main it's carried as cd52b14, the current head.

Two choices in the fix worth explaining. First, the drain lives in EventQueueSource.close() itself, after the if/else, not in ActiveTask.aclose(). Awaiting from aclose() would have patched one call site and left every other close() caller with the same latent bug. Putting it at the root means close() now guarantees the dispatcher is fully retired on return, for all callers, in both the immediate and graceful branches.

Second, the await is wrapped in contextlib.suppress(asyncio.CancelledError, Exception). CancelledError is the expected result of the cancel. Exception is included because close() is teardown, reachable from __aexit__, and shouldn't raise mid-cleanup. Nothing is silently lost, since _dispatch_loop already logs its own exceptions before exiting. If you'd rather have close() re-raise a non-cancellation dispatcher failure instead of relying on the loop's own logging, happy to make that change.

There's also a regression test for your exact scenario: test_close_immediate_awaits_dispatcher_without_sinks builds a source with no sinks, closes it immediately, and asserts the dispatcher is done when close() returns. It fails on the pre-fix code.

astrogilda:
@mykytanetipa Your finding is addressed in 0728696: the drain now happens at the root in EventQueueSource.close(), so every caller gets the guarantee, and there's a regression test for the subscriber-less immediate-close case. Pushed on the 17th, the day after your changes-requested review; details in the thread reply above.

The branch has also been rebased onto current main (the 1.1.1 release commit) with no content changes to the three PR commits, and the touched-area suite and ruff are clean.

Requesting a re-review, since the change this review asked for was pushed on the 17th and is in the current head.

mykytanetipa:
@astrogilda thank you for the quick fix, I see that tests/server/agent_execution/test_active_task.py is conflicting with main. Can you please resolve it?

astrogilda:
@mykytanetipa Done. Rebased onto current main and fixed the overlap in that test file. The teardown-warning test arunmm8335 landed in #1122 and our three drain tests cover different paths, so I kept all four. Green here: agent_execution plus request_handlers, 131 passed, ruff clean.

@rohityan if you get a minute: this is the third rebase forced by unrelated edits to the same test file (#1053, then #1122). We settled the design in #1101 and the one review note is handled, so a merge would spare us the next round. Happy to turn around anything you flag.

Edited 2026-08-20 to fix an attribution. The teardown-warning test in #1122 is arunmm8335's work, not mykytanetipa's; this said "your test" to the wrong person.

mykytanetipa (approved):

PR #1110, opened 2026-06-21
fix: prevent orphan SUBMITTED rows when producer fails early

### LINKED ISSUE #1067: [Bug]: ActiveTaskRegistry.get_or_create produces orphan SUBMITTED rows on early _run_producer exception

Linux2010:
## Summary

Fixes #1067: ActiveTaskRegistry produces orphan SUBMITTED rows when AgentExecutor.execute() raises before emitting any events.

## Problem

In long-running production deployments using `DefaultRequestHandlerV2` + `DatabaseTaskStore`, tasks could be persisted as orphan rows:
- `status.state == TaskState.TASK_STATE_SUBMITTED`
- `history == []` (empty)
- `metadata == None`
- `last_updated == None`

This occurred due to a race between `ActiveTaskRegistry.get_or_create()` and the `_run_producer` exception path.

## Solution

**Pass initial_message to TaskManager**: Add `initial_message` parameter to `ActiveTaskRegistry.get_or_create()` and forward `params.message` from `DefaultRequestHandlerV2._setup_active_task`.

This ensures that even if the producer fails before emitting any events, the persisted row has a populated `history` with the original message.

## Changes

- `src/a2a/server/agent_execution/active_task_registry.py`: Add `initial_message` parameter
- `src/a2a/server/request_handlers/default_request_handler_v2.py`: Pass `params.message` to `get_or_create()`
- `tests/server/agent_execution/test_active_task.py`: Add regression test for context_id

## Testing

- All existing tests pass
- New test verifies context_id is correctly set

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler_v2.py
+++ b/src/a2a/server/request_handlers/default_request_handler_v2.py
@@ -231,7 +231,7 @@ class DefaultRequestHandlerV2(RequestHandler):
             context_id=context_id,
             call_context=call_context,
             create_task_if_missing=True,
-            initial_message=request_context.message,
+            initial_message=params.message,
         )
 
         return active_task, request_context


Linux2010:
已根据 CR 反馈修复：在 TaskStatusUpdateEvent 中添加了 context_id 字段。请重新审核。

Linux2010:
感谢 CR 反馈！

关于 `context_id` 的评论是针对之前已回滚的代码。

## 回滚原因

我在 producer exception handler 中添加的显式 FAILED 状态转换 (`save_task_event`) 破坏了现有测试的异常传播行为：

1. **原行为**: producer 异常 → enqueue 异常 → consumer re-raise → subscribe 捕获异常
2. **新行为**: producer 异常 → `save_task_event` → 任务状态变为 FAILED → consumer 处理异常时发现任务已完成 → 抛出 `InvalidParamsError`

这导致 3 个测试失败：
- `test_active_task_producer_failure`
- `test_active_task_subscribe_exception_handling`
- `test_active_task_start_producer_immediate_error`

## 当前解决方案

只保留了传递 `initial_message` 到 TaskManager 的修改，确保即使 producer 失败，history 也有内容。

这解决了 issue #1067 的核心问题：orphan SUBMITTED rows 没有 history。

如果需要显式 FAILED 状态转换，可能需要在不破坏异常传播的前提下实现，例如：
- 在 consumer 的异常处理路径中处理
- 或者修改测试以适应新的行为

请指导下一步方向。

Linux2010:
Closing this PR as the issue #1067 has already been resolved on main by other commits:

1. **Producer exception handler** (main) already calls `ensure_task_id()` + `save_task_event(task)` with FAILED state transition, preventing orphan SUBMITTED rows
2. **`initial_message` parameter** already exists in `get_or_create()` and is passed to `TaskManager`
3. **Regression tests** already exist on main:
   - `test_on_message_send_early_producer_exception_marks_task_failed_and_preserves_originating_message`
   - `test_on_message_send_early_producer_exception_preserves_originating_message`

The remaining change in this PR (`initial_message=request_context.message` → `initial_message=params.message`) is functionally equivalent because `RequestContext.message` returns `self._params.message` and protobuf objects are passed by reference — both point to the same object.

No behavior change, no longer needed.

PR #1114, opened 2026-06-26
feat: add composable URL validation utilities

### LINKED ISSUE #1023: [Task]: Prepare URL validation infrastructure for agent card and webhook SSRF protection
### Context

We have two places which can benefit from SSRF protection:
1. Client: fetching agent cards and using URLs to make A2A calls (brought in #895, filed a separate issue #975).
2. Push notification webhooks URLs (#786): there is a spec statement about SSRF protection there.

### Description

Both cases above can share domain agnostic machinery for URL validation.

The following requirements apply:
1. Composable: local addresses are perfectly fine for some deployments.
2. Validation should happen before the actual invocation and invocation should use "pinned" IP to protect from DNS rebinding. 
3. Built-in rules: `RequireScheme` (to restrict HTTP(S)-only or HTTPS-only), `BlockPrivateNetworks` (with allowlist for deployments where using private networks is a legitimate use-case).

This issue scopes to **`UrlValidator`** only and should be a foundation for #975 and #786 where appropriate domain validators are going to be implemented.

```mermaid
flowchart TB
    subgraph Wrappers["Domain wrappers (own which fields to validate + domain error type)"]
        direction LR
        PN["PushNotificationUrlValidator"]
        AC["AgentCardUrlValidator"]
    end
    V["UrlValidator<br/>parse → resolve → run rules in order<br/>(rule raises = reject, returns = continue)"]
    subgraph Rules["Composable rules"]
        direction LR
        RS["RequireScheme"]
        BPN["BlockPrivateNetworks<br/>(allow_hosts, allow_cidrs)"]
        CR["…custom…"]
    end
    PN --> V
    AC --> V
    V --> Rules
    class PN,AC wrap
    class V core
    class RS,BPN,HD,CR rule
```

Sketch:

```python
class InvalidUrlError(ValueError):
    """Raised by UrlValidator.validate when a URL is rejected."""

class UrlValidationRule(ABC):
    @abstractmethod
    async def check(self, url: ResolvedUrl) -> None:
        """Raise InvalidUrlError to reject url."""

class ResolvedUrl:
    raw: str
    parsed: SplitResult
    addresses: tuple[IpAddress, ...]


class UrlValidator:
    def __init__(
        self,
        rules: Sequence[UrlValidationRule] = (),
        resolve: bool = True,
    ) -> None:    
    . . .

    async def validate(self, url: str) -> ResolvedUrl:
        # Conditionally resolves IPs via getaddrinfo
        resolved = await self._build(url)

        # Just runs all rules one by one, they throw to stop
        for rule in self._rules:
            await rule.check(resolved)

        # Returns URL so that pinned address can be used
        return resolved
```

Linux2010:
## Summary

Adds reusable URL validation infrastructure for #1023:

- `UrlValidator` parses URLs, optionally resolves hosts, and returns a `ResolvedUrl` containing pinned IP addresses.
- `UrlValidationRule` provides composable async validation hooks.
- Built-in rules:
  - `RequireScheme`
  - `BlockPrivateNetworks` with `allow_hosts` and `allow_cidrs`
- Exports the public utilities from `a2a.utils`.

This PR intentionally stops at the issue scope: it does not wire validation into agent-card fetching or push notification webhooks yet. Those domain-specific integrations can build on this foundation in follow-up work.

## Testing

- `uv run pytest tests/utils/test_url_validator.py -q`
- `./scripts/lint.sh`
- `uv run pytest`
- `uv run pytest --cov=src --cov-report=term-missing`

Closes #1023

### DIFF
--- a/src/a2a/utils/__init__.py
+++ b/src/a2a/utils/__init__.py
@@ -7,12 +7,26 @@ from a2a.utils.constants import (
     TransportProtocol,
 )
 from a2a.utils.proto_utils import to_stream_response
+from a2a.utils.url_validator import (
+    BlockPrivateNetworks,
+    InvalidUrlError,
+    RequireScheme,
+    ResolvedUrl,
+    UrlValidationRule,
+    UrlValidator,
+)
 
 
 __all__ = [
     'AGENT_CARD_WELL_KNOWN_PATH',
     'DEFAULT_RPC_URL',
+    'BlockPrivateNetworks',
+    'InvalidUrlError',
+    'RequireScheme',
+    'ResolvedUrl',
     'TransportProtocol',
+    'UrlValidationRule',
+    'UrlValidator',
     'proto_utils',
     'to_stream_response',
 ]
--- a/src/a2a/utils/url_validator.py
+++ b/src/a2a/utils/url_validator.py
@@ -1,0 +1,179 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


Linux2010:
Closing as duplicate of #1117 which has the improved implementation with all review feedback addressed.

PR #1130, opened 2026-07-10
refactor(grpc): remove grpcio-status dependency

Iwaniukooo11:
### Summary
Removed the third-party dependency `grpcio-status` to reduce the SDK's attack surface. Replaced it with native helper functions in the client and server transport layers.

---

### Changes
1.  **Dependencies:** Removed `grpcio-status` from `pyproject.toml` and updated `uv.lock`.
2.  **Client (`grpc.py`):** Replaced `rpc_status.from_call` with a local `_from_call` helper that parses status payloads and validates details.
3.  **Server (`grpc_handler.py`):** Replaced `rpc_status.to_status` with a local `_to_status` helper that maps status codes and returns a typed `GrpcStatus` namedtuple.

---

### Verification
Ran tests and linters locally with `grpcio-status` uninstalled:
```bash
./scripts/lint.sh
uv run pytest tests/client/transports/test_grpc_client.py tests/server/request_handlers/test_grpc_handler.py
```
**Status:** 73 tests passed, linter green.


src/a2a/client/transports/grpc.py
```diff
@@ -55,13 +54,35 @@
 logger = logging.getLogger(__name__)
 
 
-def _map_grpc_error(e: grpc.aio.AioRpcError) -> NoReturn:
+def _from_call(call: grpc.Call) -> status_pb2.Status | None:
```
ishymko: nit: `_status_from_call` or smth like that? I believe the name is reasonable in a module specific to status but here the scope is wider.

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -400,8 +421,8 @@ async def abort_context(
                 )
                 status.details.append(bad_request_detail)
 
-            # Use grpc_status to safely generate standard trailing metadata
-            rich_status = rpc_status.to_status(status)
+            # Use local helper to safely generate standard trailing metadata
```
ishymko: nit: I think this comment can be removed completely.

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -37,6 +35,29 @@
 from a2a.utils.proto_utils import validation_errors_to_bad_request
 
 
+class GrpcStatus(NamedTuple):
```
ishymko: Consider creating an internal module with both functions under `utils` and adding some "roundtrip" test, so that code used to set this status produces something that `from_call` can recognize. This should also help to reuse `grpc-status-details-bin` constant which is not a magic string in both places.

ishymko (approved):

PR #1132, opened 2026-07-13
refactor(http): eliminate third-party httpx-sse dependency

Iwaniukooo11:
Summary
Removed the third-party dependency `httpx-sse` to reduce the SDK's attack surface. Replaced it with a native, lightweight Server-Sent Events (SSE) streaming parser inside the HTTP transport helpers.

Changes
- **Dependencies:** Removed `httpx-sse` from `pyproject.toml` and updated `uv.lock`.
- **Client (http_helpers.py):** Replaced `EventSource.aiter_sse()` with a local `parse_sse_stream` generator that parses streaming line blocks and yields typed event payloads.
- **Tests:** Refactored `test_rest_client.py` and `test_jsonrpc_client.py` to mock standard `Response` line streaming instead of the third-party `EventSource` object.

Verification
Ran tests and linters locally with `httpx-sse` uninstalled:

```bash
./scripts/lint.sh
uv run pytest tests/client/transports/test_rest_client.py tests/client/transports/test_jsonrpc_client.py
```

Status: 72 tests passed, linter green.


src/a2a/client/transports/http_helpers.py
```diff
@@ -69,6 +63,46 @@ async def send_http_request(
         return response.json()
 
 
+async def parse_sse_stream(
```
ishymko, line 66: Not sure if Gemini is correct here but either way I'd recommend improving test coverage for this function as some conditions were not hit:

Image

(you can download this report from the code coverage bot comment).

src/a2a/client/transports/http_helpers.py
```diff
@@ -94,29 +128,29 @@ async def send_http_stream_request(
...
                     continue
-                if sse.event == 'error':
-                    sse_error_handler(sse.data)
-                yield sse.data
+                if event_name == 'error':
+                    sse_error_handler(data)
+                yield data
 
 
 class _SSEEventSource:
```
ishymko, line 156: `_SSEEventSource` was added in #912 as a workaround for this library. Could you please check if it's still required? Also its comment likely requires update as it mentions `aconnect_sse` function from the package we remove now.

Iwaniukooo11: Regarding the docstring - You are right that it was outdated, thanks for checking that

Regarding the `_SSEEventSource` - Dumping `httpx_sse` doesn't resolve the nested async generators issue (as the core behaviour of `send_http_stream_request` didn't change), thus  current class still handles early break scenario in a proper way, acting as a `@contextmanager` with properly defined `__aexit__` and `__aenter__` methods

ishymko (approved):

PR #1140, opened 2026-07-20
fix(client): apply all schemes in multi-scheme security requirement (AND semantics)

### LINKED ISSUE #1138: [Bug]: AuthInterceptor applies only the first scheme of a multi-scheme security requirement (AND semantics broken)
## Summary

`AuthInterceptor.before()` returns as soon as it successfully applies **one** credential. When a single `SecurityRequirement` lists **multiple** schemes — which, per the OpenAPI security requirement semantics the A2A spec follows, means **all** of them must be satisfied together (AND) — only one credential ever gets attached to the outgoing request.

In my setup, the agent card declares one requirement containing two api-key-in-header schemes (a gateway auth key + a user id header). The gateway requires both headers, but the interceptor only sends one, so every request is rejected.

Because proto3 map iteration order is not deterministic, *which* of the schemes gets applied can even differ between runs (see log output below), making the failure intermittent-looking and hard to debug.

## Expected behavior

Per OpenAPI Security Requirement semantics (A2A's security model is based on the OpenAPI Security Scheme / Security Requirement objects):

- the outer `security_requirements` list is **OR** — satisfying any one requirement is sufficient;
- multiple schemes **inside a single requirement** are **AND** — all of them must be applied together.

So for the card below, both `AUTH_KEY` and `USER_ID` headers should be attached.

## Reproduction

```python
# pip install "a2a-sdk==1.1.1"  (also reproduces on main @ 86c6b0d)
import asyncio

from a2a.client.auth.credentials import CredentialService
from a2a.client.auth.interceptor import AuthInterceptor
from a2a.client.interceptors import BeforeArgs
from a2a.types.a2a_pb2 import (
    AgentCard,
    APIKeySecurityScheme,
    SecurityRequirement,
    SecurityScheme,
    StringList,
)

CREDS = {"AUTH_KEY": "gw-secret-key", "USER_ID": "user-123"}


class DictCredentialService(CredentialService):
    async def get_credentials(self, security_scheme_name, context):
        return CREDS.get(security_scheme_name)


card = AgentCard(
    security_schemes={
        "AUTH_KEY": SecurityScheme(
            api_key_security_scheme=APIKeySecurityScheme(
                location="header", name="AUTH_KEY"
            )
        ),
        "USER_ID": SecurityScheme(
            api_key_security_scheme=APIKeySecurityScheme(
                location="header", name="USER_ID"
            )
        ),
    },
    # ONE requirement containing BOTH schemes -> AND: both headers required
    security_requirements=[
        SecurityRequirement(
            schemes={"AUTH_KEY": StringList(), "USER_ID": StringList()}
        )
    ],
)


async def main():
    args = BeforeArgs(input=None, method="message/send", agent_card=card)
    await AuthInterceptor(DictCredentialService()).before(args)
    # Expected: {'AUTH_KEY': 'gw-secret-key', 'USER_ID': 'user-123'}
    # Actual:   only one of the two headers is present
    print(args.context.service_parameters)


asyncio.run(main())
```

Note that credentials for **both** schemes are available from the `CredentialService` — the missing header is purely due to the interceptor's control flow.

## Root cause

In `src/a2a/client/auth/interceptor.py`, every successful branch (`bearer`, `oauth2`/`oidc`, api-key-in-header) ends with a `return` **inside the inner per-scheme loop**, so processing stops after the first applied credential instead of finishing the requirement.

## Proposed fix

Move the `return` so it happens **after the inner (per-scheme) loop completes**, i.e. once one requirement has been *fully* satisfied. Simply removing the returns and finishing both loops would over-correct: it would apply credentials from *alternative* (OR) requirements too, and since the bearer/oauth2/oidc branches all write to the same `'Authorization'` key, a later requirement would silently overwrite an earlier one (last-write-wins).

A two-pass approach per requirement keeps AND strict (no partial application when a credential is missing) while preserving OR across requirements:

```python
for requirement in agent_card.security_requirements:
    # pass 1: can EVERY scheme in this requirement be satisfied?
    collected: dict[str, str] = {}
    satisfiable = True
    for scheme_name in requirement.schemes:
        credential = await self._credential_service.get_credentials(
            scheme_name, args.context
        )
        if not credential or scheme_name not in agent_card.security_schemes:
            satisfiable = False
            break
        collected[scheme_name] = credential

    if not satisfiable:
        continue  # OR: try the next requirement

    # ... ensure args.context / service_parameters exist ...

    # pass 2: apply ALL schemes of this requirement
    for scheme_name, credential in collected.items():
        scheme = agent_card.security_schemes[scheme_name]
        # existing per-scheme-type application logic, without early returns

    return  # one requirement fully satisfied -> done
```

One open design question: should a scheme type the interceptor cannot apply (e.g. API key in `query`/`cookie`, currently skipped) make the whole requirement unsatisfiable (strict reading of AND), or be skipped as today? Happy to align with whatever the maintainers prefer.

I'm glad to submit a PR (including unit tests for the multi-scheme AND case and the OR fallback) if this direction sounds right.

## Current workaround

We're temporarily carrying a local patch that simply removes the early returns and returns after the loop completes. That happens to be safe for our card because it declares a single requirement, but as explained above it is not a general fix (it would break OR semantics for multi-requirement cards) — hence this issue.

## Environment

- `a2a-sdk` 1.1.1 (latest release) — reproduced by execution
- `main` @ `86c6b0d` — reproduced by execution
- Python 3.12, Linux

Related: #445 (broader client auth ergonomics — this issue is only about the correctness of the existing `AuthInterceptor` control flow).

### Relevant log output

```shell
$ python repro.py
{'AUTH_KEY': 'gw-secret-key'}          # USER_ID header missing -> gateway returns 401/403

# proto3 map iteration order is not stable across processes, so WHICH header
# gets applied is nondeterministic. Reading the same card in 3 fresh processes:
$ for i in 1 2 3; do python -c "..."; done
['AUTH_KEY', 'USER_ID']
['AUTH_KEY', 'USER_ID']
['USER_ID', 'AUTH_KEY']                # order flipped -> USER_ID would be the one applied
```

### Code of Conduct

arunmm8335:
Fixes #1138

### DIFF
--- a/src/a2a/client/auth/interceptor.py
+++ b/src/a2a/client/auth/interceptor.py
@@ -1,12 +1,20 @@
-import logging  # noqa: I001
+from __future__ import annotations
 
-from a2a.client.auth.credentials import CredentialService
+import logging
+
+from typing import TYPE_CHECKING
+
 from a2a.client.client import ClientCallContext
 from a2a.client.interceptors import (
     AfterArgs,
     BeforeArgs,
     ClientCallInterceptor,
 )
+
+
+if TYPE_CHECKING:
+    from a2a.client.auth.credentials import CredentialService
+    from a2a.types.a2a_pb2 import SecurityScheme
 
 logger = logging.getLogger(__name__)
 
@@ -21,7 +29,17 @@ class AuthInterceptor(ClientCallInterceptor):
         self._credential_service = credential_service
 
     async def before(self, args: BeforeArgs) -> None:
-        """Applies authentication headers to the request if credentials are available."""
+        """Applies authentication headers to the request if credentials are available.
+
+        Follows OpenAPI Security Requirement semantics:
+        - The outer ``security_requirements`` list uses **OR** semantics:
+          satisfying any single requirement is sufficient.
+        - Multiple schemes **within** a single requirement use **AND**
+          semantics: all of them must be satisfied together.
+
+        A two-pass approach is used per requirement to avoid partial
+        application when one scheme's credential is unavailable.
+        """
         agent_card = args.agent_card
 
         # Proto3 repeated fields (security) and maps (security_schemes) do not track presence.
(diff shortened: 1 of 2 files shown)


src/a2a/client/auth/interceptor.py
```diff
@@ -21,7 +29,17 @@ def __init__(self, credential_service: CredentialService):
...
+        """Applies authentication headers to the request if credentials are available.
+
+        Follows OpenAPI Security Requirement semantics:
+        - The outer ``security_requirements`` list uses **OR** semantics:
+          satisfying any single requirement is sufficient.
+        - Multiple schemes **within** a single requirement use **AND**
+          semantics: all of them must be satisfied together.
+
+        A two-pass approach is used per requirement to avoid partial
+        application when one scheme's credential is unavailable.
```
sokoliva, line 41: Let's remove implementation details and just keep high level description to not clutter the code with comments.

arunmm8335: Yes sure will do that could you please review it and assign this issue to me

src/a2a/client/auth/interceptor.py
```diff
@@ -34,63 +52,92 @@ async def before(self, args: BeforeArgs) -> None:
             return
 
         for requirement in agent_card.security_requirements:
+            # Pass 1: collect credentials for every scheme in this
+            # requirement.  If any scheme cannot be satisfied the whole
+            # requirement is skipped (AND semantics).
```
sokoliva, line 57: This detailed description is not needed, it adds unnecessary clutter, the behaviour from the description in visible in the next few lines of code.

src/a2a/client/auth/interceptor.py
```diff
@@ -34,63 +52,92 @@ async def before(self, args: BeforeArgs) -> None:
...
+                    # Treat as unsatisfiable so the requirement is skipped.
+                    satisfiable = False
+                    break
+
+                collected[scheme_name] = header
+
+            if not satisfiable:
+                continue  # OR: try the next requirement
+
+            # Pass 2: apply all collected credentials at once.
```
sokoliva, line 86: Remove the comment.

tests/client/test_auth_interceptor.py
```diff
@@ -332,3 +332,221 @@ async def test_auth_interceptor_skips_when_scheme_not_in_security_schemes(
...
+        security_schemes=security_schemes,
+        security_requirements=security_requirements,
+    )
+
+
+@pytest.mark.asyncio
+async def test_multi_scheme_and_semantics() -> None:
+    """A single requirement with two API-key-in-header schemes must apply both headers.
+
+    This is the exact reproduction from issue #1138.
```
sokoliva, line 379: Remove issue reference

tests/client/test_auth_interceptor.py
```diff
@@ -332,3 +332,221 @@ async def test_auth_interceptor_skips_when_scheme_not_in_security_schemes(
 
     await auth_interceptor.before(args)
     assert context.service_parameters is None
+
+
+# ---------------------------------------------------------------------------
+# Tests for multi-scheme AND / multi-requirement OR semantics (#1138)
+# ---------------------------------------------------------------------------
```
sokoliva, line 339: Remove this.

tests/client/test_auth_interceptor.py
```diff
@@ -332,3 +332,221 @@ async def test_auth_interceptor_skips_when_scheme_not_in_security_schemes(
...
+        name='testbot',
+        description='test',
+        version='1.0',
+        default_input_modes=[],
+        default_output_modes=[],
+        skills=[],
+        capabilities=AgentCapabilities(),
+        security_schemes=security_schemes,
+        security_requirements=security_requirements,
+    )
```
sokoliva, line 372: Let's follow the test design and put helper functions and classes at the top of the file.

sokoliva (changes requested):

src/a2a/client/auth/interceptor.py
sokoliva: Thank you! I submitted the review.

chopmob-cloud:
Verified the security-correctness of this against current main. On main, AuthInterceptor.before iterates `for requirement in agent_card.security_requirements:` (36) then `for scheme_name in requirement.schemes:` (37), and each scheme branch returns as soon as one credential is attached (63 Bearer, 76 OAuth2/OIDC, 91 API key). So for a requirement that lists more than one scheme, only the first scheme's header is applied and the request goes out under one factor.

That is the part worth stating precisely: for an AND requirement such as [bearer, apiKeyHeader], the old single-return path emits Authorization and drops the API key, producing a request that is authenticated for one factor and silently missing the other. This PR's collect-then-apply design fixes exactly that: pass one gathers a header for every scheme in the requirement and marks the requirement unsatisfiable on any miss, pass two applies them together, and it returns after the first fully satisfiable requirement. That gives AND within a requirement and OR across requirements, and it guarantees a requirement is applied whole or skipped whole. No partial application. That is the right property.

One behavior change to call out in the notes so integrators are not surprised: unsupported API-key placements (query and cookie) now make the whole requirement unsatisfiable and cause a fallthrough to the next requirement, where main silently skipped only that one scheme and kept the rest. That is more correct for AND (no half-authenticated request), but it is a real change from prior behavior and worth a changelog line.

CI is green across 3.10 to 3.14, the new tests cover multi-scheme AND, OR fallback and partial-credential skip, and the only outstanding review comments look stylistic. The security semantics are sound.

PR #1141, opened 2026-07-21
fix(utils): protobuf-compatible repeated field detection

### LINKED ISSUE #1011: [Task]: Replace deprecated `FieldDescriptor.label` with `is_repeated` in `proto_utils`

piyushbag:
## Summary

- Add `_is_field_repeated()` in `proto_utils` that prefers `field.is_repeated` and falls back to `field.label == FieldDescriptor.LABEL_REPEATED` when the newer API is unavailable.
- Replace the three repeated-field checks in REST param parsing and proto required-field validation with the helper.
- Add unit tests for scalar/repeated fields and for the label fallback path.

## Problem

`proto_utils` used the deprecated `FieldDescriptor.label` API. Protobuf 7 removes `label`, which breaks validation on newer runtimes even though the SDK dependency range can resolve protobuf 7.x in some environments.

## Test plan

- [x] `uv run pytest tests/utils/test_proto_utils.py -q`
- [x] Verified `_is_field_repeated` against protobuf 6.x locally (is_repeated + label both present)

Fixes #1011

### DIFF
--- a/src/a2a/utils/proto_utils.py
+++ b/src/a2a/utils/proto_utils.py
@@ -174,10 +174,7 @@ def parse_params(params: QueryParams, message: ProtobufMessage) -> None:
         field = fields[k]
         v_list = params.getlist(k)
 
-        # TODO(https://github.com/a2aproject/a2a-python/issues/1011): Replace
-        # deprecated `field.label` with `field.is_repeated` once the minimum
-        # protobuf version requirement is bumped.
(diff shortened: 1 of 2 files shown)


sokoliva:
Thank you for your contribution. The Issue [#1011 ](https://github.com/a2aproject/a2a-python/issues/1011) is labeled as "maintainers only" and has a set plan of implementation for the future, as seen in the description of the PR.

PR #1153, opened 2026-07-27
fix(server): warn when queue_manager is ignored in DefaultRequestHandlerV2

### LINKED ISSUE #1135: [Bug]: `queue_manager` passed to DefaultRequestHandler is silently ignored since v2
### What happened?

We run an A2A server on multiple replicas behind a load balancer. To make streaming work across replicas, we pass a custom `QueueManager` (Redis-backed) to `DefaultRequestHandler` that way a client can reconnect to a task's stream even if it lands on a different replica.

After upgrading the SDK, streaming across replicas stopped working, with no error or warning. The cause: since #933, `DefaultRequestHandler` points to `DefaultRequestHandlerV2`, which still accepts `queue_manager` in its constructor but never uses it:

```python
queue_manager: Any | None = None,  # Kept for backward compat in signature
```

Our queue manager's methods are simply never called anymore. Everything compiles and runs, and single-replica setups look fine, so this is very easy to miss.

**What I expected**

Either the parameter keeps working, or passing one raises / warns that it's no longer supported. Accepting it and dropping it silently is the worst case the code says the old wiring works when it doesn't.

**How to reproduce**

Pass any `QueueManager` subclass that records calls to `DefaultRequestHandler`, run a `message/stream` request, and check the recorder:

```python
class SpyQueueManager(InMemoryQueueManager):
    async def create_or_tap(self, task_id):
        print("called!")  # never prints on 1.0.x / 1.1.0
        return await super().create_or_tap(task_id)
```

On `LegacyRequestHandler` the same spy is called as before. That's our current workaround: use `LegacyRequestHandler` and pin the SDK version.

Is there an intended replacement for this use case (sharing event queues across replicas) in the v2 design? Happy to provide more detail.

### Code of Conduct

NishchayMahor:
## Summary

Fixes #1135.

`DefaultRequestHandlerV2.__init__` accepts a `queue_manager` argument (kept for signature compatibility with `DefaultRequestHandler`) but never stores or uses it. As @rohityan confirmed on the issue, v2 delegates event streaming to an in-memory `ActiveTaskRegistry`, so any custom or distributed `QueueManager` — e.g. a Redis-backed one used for multi-replica stream reconnection — is silently dropped. After upgrading, multi-replica streaming breaks with no error or warning.

## Fix

Emit a `logger.warning` at construction when a non-`None` `queue_manager` is passed, explaining that it is ignored in v2 and pointing at the documented workarounds (switch to `LegacyRequestHandler`, or route `/tasks/{id}:subscribe` to the replica holding the `ActiveTask`). This turns a silent misconfiguration into a visible one without changing v2's architecture. Also corrected the now-inaccurate inline comment on the parameter.

## Evidence

- The parameter is dropped: `default_request_handler_v2.py` `__init__` binds `agent_executor`/`task_store`/etc. but has no `self._queue_manager = queue_manager` (contrast `default_request_handler.py:123`, which does store it and uses it at lines 209/313/503/611).
- v2 uses the registry instead: `self._active_task_registry = ActiveTaskRegistry(...)`.

## Testing

Added two tests to `tests/server/request_handlers/test_default_request_handler_v2.py`:
- `test_init_warns_when_queue_manager_passed` — asserts a WARNING mentioning `queue_manager` is logged (fails on `main`, passes with the fix).
- `test_init_no_warning_without_queue_manager` — asserts no such warning in the default case.

```
uv run pytest tests/server/request_handlers/            # 195 passed
```
`ruff check`, `ruff format --check`, and `ty check` all clean on the changed files.

This change was developed with AI assistance; I verified the root cause and behavior against the code paths cited above, reviewed every line, and ran the tests.


chopmob-cloud:
Confirmed the root cause on current main: DefaultRequestHandlerV2 still accepts queue_manager in its signature (84) but never stores or uses it, since V2 drives streaming through ActiveTaskRegistry (108). A caller passing a Redis-backed QueueManager gets it dropped with no signal, which is the exact failure in the issue. Warning on a non-None value is a reasonable stopgap given that actually honoring a distributed manager in V2 is an architectural change, and it does satisfy the issue's stated bar (either the parameter works, or passing one warns).

One suggestion to make the warning land where it matters. logger.warning surfaces in server logs, but the point of the issue is that this is easy to miss, especially across replicas. warnings.warn(..., DeprecationWarning, stacklevel=2) would put the signal at the caller's construction site, make it catchable and filterable in test suites, and read as a deprecation of the parameter rather than a runtime log line. Emitting both (the DeprecationWarning plus the existing log) would cover interactive and headless callers. Since the parameter is being kept only for backward compatibility, stating that deprecation in the docstring would also carry it into the generated API docs.

Tests and CI are green. This is a clean interim fix; the warnings.warn form would match the issue's raises-or-warns framing a little more faithfully.

NishchayMahor:
Thanks for actually going and checking this against main rather than taking the description at face value — the line references are right, and confirming that `queue_manager` appears only in the signature while `__init__` builds the `ActiveTaskRegistry` is the part that matters.

On `warnings.warn(..., DeprecationWarning, stacklevel=2)`: agreed, and more so than I expected, because the codebase already does exactly this. `EventQueue.__new__` in `server/events/event_queue.py` uses the same three-argument form for the same shape of problem — a parameter/entry point kept only for backward compatibility. So I've matched that rather than inventing a house style, and I've kept the `logger.warning` too, for the reason you gave: `DeprecationWarning` is ignored by default outside `__main__`, so a headless server would otherwise see nothing at all. `stacklevel=2` is asserted in the test via `record[0].filename`, so it will fail if the attribution ever regresses.

The docstring suggestion I've taken but moved. `DefaultRequestHandlerV2.__init__` has no docstring at all — unlike `DefaultRequestHandler`, which carries a full `Args:` block — and adding an `Args:` section documenting one parameter out of nine seemed worse than the alternative, so the note is on the class docstring, which still lands in the generated docs.

One thing that came out of this that's worth recording: the repo's own integration harness hits this. `tests/integration/test_scenarios.py`, `create_handler` does

```python
queue_manager = queue_manager or InMemoryQueueManager()
```

and then passes it positionally to `DefaultRequestHandlerV2`, where it's dropped. So this isn't only a hypothetical about someone wiring up Redis — there's an in-tree caller already relying on a parameter that does nothing.

Numbers on the updated branch: 61 passed in `test_default_request_handler_v2.py`, 1775 passed / 90 skipped / 3 xfailed / 1 xpassed across `tests/`, ruff clean. With the source change stashed, the new test fails with `DID NOT WARN`. `filterwarnings` in `pyproject.toml` has no `error` entry, so the new warning doesn't turn any existing run red — the integration scenarios still pass, they just emit it now.

What I haven't done is make V2 honor a real `QueueManager`. That's the architectural change you alluded to, and it needs a maintainer's call on whether V2 is ever meant to support distributed streaming or whether the parameter should eventually come off the signature entirely. This stays a warn-only stopgap.

mykytanetipa (approved):
LGTM

Confirm that queue_manager should remain as a no-op param on DefaultRequestHandlerV2 for handler schema compatibility, the logging implementation is correct and satisfies #1135.

Actual multi-replica streaming support for DefaultRequestHandlerV2 is a separate architectural effort for the maintainer team.

PR #1158, opened 2026-08-02
fix: replace deprecated FieldDescriptor.label with is_repeated in proto_utils

### LINKED ISSUE #1011: [Task]: Replace deprecated `FieldDescriptor.label` with `is_repeated` in `proto_utils`

Linux2010:
## Summary

Closes #1011

This PR introduces a compatibility helper that prefers is_repeated when available and falls back to label otherwise, fixing the deprecation without bumping the minimum protobuf version (no breaking change for downstream consumers)

## Changes

- src/a2a/utils/proto_utils.py:
  - Add _field_is_repeated(field) helper: uses field.is_repeated when present (protobuf ≥ 6.31.0), otherwise falls back to field.label == FieldDescriptor.LABEL_REPEATED.
  - Route all three sites (parse_params, _check_required_field_violation, _recurse_validation) through the helper.
  - Remove the three TODO(#1011) comments.
- tests/utils/test_proto_utils.py: add tests covering both helper branches (new is_repeated API and the legacy label fallback), plus repeated-nested-message validation (Task.history).

## Verification

- `is_repeated` is available in the current minimum `protobuf>=5.29.5`
- All existing proto_utils tests pass (14 passed)
- Lint passes (`./scripts/lint.sh`)

### DIFF
--- a/src/a2a/utils/proto_utils.py
+++ b/src/a2a/utils/proto_utils.py
@@ -174,10 +174,7 @@ def parse_params(params: QueryParams, message: ProtobufMessage) -> None:
         field = fields[k]
         v_list = params.getlist(k)
 
-        # TODO(https://github.com/a2aproject/a2a-python/issues/1011): Replace
-        # deprecated `field.label` with `field.is_repeated` once the minimum
-        # protobuf version requirement is bumped.
-        if field.label == FieldDescriptor.LABEL_REPEATED:
+        if field.is_repeated:
             accumulated: list[Any] = []
             for v in v_list:
                 if not v:
@@ -211,10 +208,7 @@ def _check_required_field_violation(
 ) -> ValidationDetail | None:
     """Check if a required field is missing or invalid."""
     val = getattr(msg, field.name)
-    # TODO(https://github.com/a2aproject/a2a-python/issues/1011): Replace
-    # deprecated `field.label` with `field.is_repeated` once the minimum
-    # protobuf version requirement is bumped.
-    if field.label == FieldDescriptor.LABEL_REPEATED:
+    if field.is_repeated:
         if not val:
             return ValidationDetail(
                 field=field.name,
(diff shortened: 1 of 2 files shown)


mykytanetipa (changes requested):
Blocking: this breaks users on the supported protobuf floor, field.is_repeated only exists in protobuf ≥ 6.31.0, but current project.toml pins protobuf>=5.29.5,<7. For any user on 5.29.5 ≤ protobuf ≤ 6.30.2, the 3 changed sites raise AttributeError at runtime. CI is green only because it resolves to the 6.33.6 protobuf version, masking the floor violation.

Proof (minimum version):
```
cd /tmp && rm -rf pbcheck && mkdir pbcheck && cd pbcheck
python3 -m venv .venv
./.venv/bin/pip install "protobuf==5.29.5"
./.venv/bin/python -c "from google.protobuf.descriptor import FieldDescriptor; print(hasattr(FieldDescriptor,'is_repeated'))"

Looking in indexes: http://airlock-proxy.uplink.goog:999/python/artifact-foundry-prod/ah-3p-staging-python/simple/
Collecting protobuf==5.29.5
  Downloading protobuf-5.29.5-cp38-abi3-manylinux2014_x86_64.whl (319 kB)
Installing collected packages: protobuf
Successfully installed protobuf-5.29.5
False
```
(also False through 6.30.2; True from 6.31.0)

Suggest approach that uses a temporal fallback helper at the 3 sites, but requires neither < 7 pin nor breaking change version bump at this moment:
```
def _field_is_repeated(field: FieldDescriptor) -> bool:
    """protobuf>=6.31.0 exposes is_repeated; older (our >=5.29.5 floor) does not."""
    is_rep = getattr(field, 'is_repeated', None)
    if is_rep is not None:
        return is_rep
    return field.label == FieldDescriptor.LABEL_REPEATED
```

mykytanetipa (changes requested):
Please update description to match the _field_is_repeated approach, right now it references the older version

LGTM after description update

mykytanetipa (approved):

PR #1164, opened 2026-08-05
fix(server): validate push-notification URLs before dispatch (SSRF hardening)

SashaMIT:
## Summary
In plain terms: when a client tells an A2A agent "send my task updates to this webhook", the agent server POSTs to **whatever URL the client supplied — no checks at all**. That means any client can make the agent server send requests to internal network addresses: cloud metadata endpoints (`169.254.169.254`), `localhost` admin panels, or unauthenticated internal services. This is the classic server-side request forgery (SSRF) pattern, and it fires on every task event.

Concretely, `BasePushNotificationSender._dispatch_notification` used `push_info.url` exactly as supplied:
- no scheme restriction (`ftp://`, future handler schemes),
- no destination restriction (loopback / link-local / RFC1918 / reserved),
- and configs can be registered through multiple paths (`tasks/pushNotificationConfig/create`, inline on `message/send`), so write-time validation alone wouldn't cover all of them.

## Fix
`BasePushNotificationSender` now validates each URL **at dispatch time** (one choke point covering every registration path):
- scheme must be `http`/`https`,
- host must resolve (unresolvable fails closed — the POST would fail anyway),
- every resolved address must be public unicast; loopback, link-local, private, reserved, multicast, and unspecified addresses are rejected.

Operators whose legitimate webhooks live on private networks opt out explicitly: `BasePushNotificationSender(..., allow_private_push_urls=True)`.

**Residual risk, stated honestly:** DNS rebinding between validation and the POST itself remains possible for attacker-controlled domains (validation and the actual connection resolve the name separately). Static internal targets — the realistic SSRF cases here — are fully blocked. Noted in the constructor docstring.

## Test plan
- [x] 7 new unit tests: metadata IP blocked, loopback blocked, private range blocked, non-HTTP scheme blocked, unresolvable host fails closed, public host allowed, opt-out allows private
- [x] `tests/server/tasks/` 185 pass (existing suites made DNS-hermetic)
- [x] Push-notification e2e suite passes (test app opts out — its webhooks are real local servers)
- [x] `ruff check` + `ruff format` clean on touched files

Made with Cursor

Made with [Cursor](https://cursor.com)


kuangmi-bit (commented):
Reviewed the SSRF hardening — solid implementation. A few notes from having done the same analysis on a sibling A2A-ecosystem project (a registry service) recently:

**What's done well:**
- Validation happens after DNS resolution (`getaddrinfo` → check every returned address) — this catches IP literals in integer/hex forms and IPv4-mapped IPv6 (`::ffff:127.0.0.1`), not just dotted-quad strings
- Fail-closed on unresolvable hosts (the POST would fail anyway; treating it as a pass would be a bypass)
- The `for info in infos: if blocked -> reject` loop requires ALL resolved addresses to be public, not just any — correct
- `allow_private_push_urls` escape hatch keeps legitimate private-network webhooks working without weakening the default
- Test coverage is thorough (metadata endpoint, loopback, private range, scheme, fail-closed, public allow, opt-out)

**Two residual risks worth documenting (not blocking):**

1. **Redirect targets are not re-validated.** `httpx.AsyncClient` defaults to `follow_redirects=False`, but a caller can enable it — in that case the initial URL passes validation and a redirect to an internal address (e.g. `https://public.example` → `http://169.254.169.254/`) is dispatched without re-checking. Worth a doc note on the `httpx_client` parameter: "validation covers the initial URL only; keep `follow_redirects=False` (the default) or the client is exposed to redirect-based SSRF."

2. **DNS rebinding TOCTOU window.** Validation and connection are two separate resolutions; a hostile DNS server can return a public IP for the validation lookup and a private IP for the connection lookup. Hard to close fully at this layer (would require pinning the validated IP in the transport), but worth documenting as a known limitation so operators can mitigate with network controls.

Minor: consider a short docstring note in `push_url_validation_error` that IPv4-mapped IPv6 is covered (the `is_private`/`is_loopback` checks on mapped addresses already handle it, but the comment would save future readers a double-take).

SashaMIT:
Thanks @kuangmi-bit, both points addressed in 9747ab2:

1. **Redirects**: the constructor now rejects an `httpx.AsyncClient` with `follow_redirects=True` outright, with a `ValueError` explaining why. That turns the redirect-SSRF misconfiguration into a fail-fast at startup instead of a doc note people may miss. The residual-risk note is also in the `httpx_client` docstring.
2. **DNS rebinding TOCTOU**: documented as a known limitation in `push_url_validation_error` (validation and connection resolve separately; pinning the validated IP belongs in the transport layer), framed as defense-in-depth with network-level egress controls as the operator mitigation.

Also added the IPv4-mapped IPv6 note to the docstring as suggested, and a test asserting the constructor guard raises. Full suite green (1354 passed).

kuangmi-bit (approved):
Both points from my earlier review are addressed in 9747ab2 (fail-fast on follow_redirects=True, plus the second item). The SSRF hardening looks solid — approving. (cc: the duplicate #1169 covers the same ground; consider closing one.)

ez-lbz:
Hi @SashaMIT,

Re: the overlap discussion on [#1169](https://github.com/a2aproject/a2a-python/pull/1169) — a few points from our side:

1. [#1169](https://github.com/a2aproject/a2a-python/pull/1169) is part of a batch of findings from a test framework we are building to exercise the A2A protocol across the SDKs; the whole batch was filed on the same day.
2. Between two PRs covering the same fix, the deciding factor should be the check results, not the filing order. As of now, [#1164](https://github.com/a2aproject/a2a-python/pull/1164) has not passed all of its required checks — and per project rules, a PR that has not passed all checks cannot be accepted.
3. On that basis, review should treat [#1169](https://github.com/a2aproject/a2a-python/pull/1169) as the primary candidate; if [#1164](https://github.com/a2aproject/a2a-python/pull/1164) is to be considered at all, its failing checks need to be resolved first.

Happy to defer to the maintainers either way.

sokoliva (commented):
Thank you for this PR. Could you please resolve Lint and Check Spelling issues?

SashaMIT:
Thanks. Lint and Check Spelling should be clear now (type-check on the resolved address, and the unrecognized words are gone from the docstring).

ez-lbz:
Following up on my comment above: confirmed — `Lint Code Base` and `Check Spelling` now pass on the current head, so that objection is resolved.

The one remaining red, `ITK`, looks transient rather than code-related: the run logs show `Error response from daemon: No such container: itk-service` (the star-topology peers — go_v03/go_v10/python_v03/python_v10 — never came up), and ITK failed for two unrelated PRs at the same minute (17:55) while runs for other branches before and after that window succeed. A rerun should clear it.

With both #1164 and #1169 green, the overlap question is squarely the maintainers' call — happy to defer either way, as said.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -20,6 +23,68 @@
...
+        return 'no hostname'
+    port = parsed.port or (443 if parsed.scheme == 'https' else 80)
+    try:
+        infos = socket.getaddrinfo(host, port, type=socket.SOCK_STREAM)
+    except OSError:
+        return f"host '{host}' could not be resolved"
+    for info in infos:
+        if _ip_is_blocked(str(info[4][0])):
+            return f"host '{host}' resolves to a non-public address"
+    return None
```
sokoliva: This was merged in PR #1173, but let's actually move it to a new file /src/a2a/utils/push_url_validator.py.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -28,6 +93,8 @@ def __init__(
         httpx_client: httpx.AsyncClient,
         config_store: PushNotificationConfigStore,
         context: ServerCallContext | None = None,
+        *,
+        allow_private_push_urls: bool = False,
```
sokoliva: Let's do 
```
push_url_validator: Callable[[str], Awaitable[str | None]]
        | None = None
```
like we did in #1173 and encourage users to use push_url_validation_error in docstring.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -52,8 +134,17 @@ def __init__(
...
                 'context argument from the constructor call.'
             )
+        if httpx_client.follow_redirects:
+            raise ValueError(
+                'BasePushNotificationSender validates the initial push URL '
+                'only; a client with follow_redirects=True would dispatch '
+                'redirect targets without re-validation (redirect-based '
+                'SSRF). Construct the client with follow_redirects=False '
+                '(the default).'
+            )
```
sokoliva: Let's remove this. Since this is a breaking change let's not do it today We will consider this for a next major version update.

sokoliva (changes requested):
Thank you for this PR! I left a few comments.

SashaMIT:
Thanks Iva. I moved the validator to src/a2a/utils/push_url_validator.py, defaulted the sender hook to None like #1173 (the docstring points at push_url_validation_error), and dropped the follow_redirects constructor raise. This branch is now on current main so it sits on top of #1173.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -18,71 +17,15 @@
...
-    except OSError:
-        return f"host '{host}' could not be resolved"
-    for info in infos:
-        if _ip_is_blocked(str(info[4][0])):
-            return f"host '{host}' resolves to a non-public address"
-    return None
+__all__ = [
+    'BasePushNotificationSender',
+    'push_url_validation_error',
+]
```
sokoliva: Please remove. Since we moved `push_url_validation_error` to utils, let's also import from 
`from a2a.utils.push_url_validator`

Alternatively, You can add `push_url_validation_error` to `a2a/utils/__init__.py`.

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,63 @@
...
+    try:
+        parsed = urllib.parse.urlparse(url)
+    except ValueError:
+        return 'unparseable URL'
+    if parsed.scheme not in ('http', 'https'):
+        return f"scheme '{parsed.scheme}' is not http/https"
+    host = parsed.hostname
+    if not host:
+        return 'no hostname'
+    port = parsed.port or (443 if parsed.scheme == 'https' else 80)
```
sokoliva: It seems port also raises ValueError: [port](https://github.com/python/cpython/blob/4e8bce4d547eab49410ae9fb0fcf7b67c2349056/Lib/urllib/parse.py#L176)

Let's also put it in a try/except block to catch those errors, something like: 
```
try:
     parsed = urllib.parse.urlparse(url)
     explicit_port = parsed.port
except ValueError:
       return 'unparseable URL'
(.... other code ...)
port = explicit_port or (443 if parsed.scheme == 'https' else 80)
```

sokoliva (changes requested):

sokoliva:
Left a couple more NITs, please fix and then re-request review. :)

SashaMIT:
Thanks Iva. I dropped the sender re-export so `push_url_validation_error` is imported from `a2a.utils.push_url_validator` only, and wrapped `urlparse` plus `parsed.port` so a bad port returns unparseable instead of raising. Re-requesting review.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -22,55 +21,9 @@
...
-        infos = await loop.getaddrinfo(host, port, type=socket.SOCK_STREAM)
-    except OSError:
-        return f"host '{host}' could not be resolved"
-    for info in infos:
-        if _ip_is_blocked(str(info[4][0])):
-            return f"host '{host}' resolves to a non-public address"
-    return None
+__all__ = [
+    'BasePushNotificationSender',
+]
```
sokoliva: Please remove this too. We export it [here](https://github.com/a2aproject/a2a-python/blob/4b7b24293c55518e3f8b815b04fadf77ed488505/src/a2a/server/tasks/__init__.py#L72) so there is no need to export it also in this file.

sokoliva (changes requested):

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,64 @@
...
+        addr.is_private
+        or addr.is_loopback
+        or addr.is_link_local
+        or addr.is_multicast
+        or addr.is_reserved
+        or addr.is_unspecified
+    )
+
+
+async def push_url_validation_error(url: str) -> str | None:
```
mykytanetipa: can we rename this to smth like validate_push_notification_url? the "error" prefix kinda distorts the purpose of the function which is validation.

mykytanetipa: since the purpose is validation I would change the return type to bool instead of "str | None", it's uncommon to use None for "safe" outcome

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    ``push_url_validator`` on ``DefaultRequestHandler`` /
+    ``DefaultRequestHandlerV2`` / ``BasePushNotificationSender``.
+    The default on those constructors is ``None`` (no library
+    screening).
+    """
+    try:
+        parsed = urllib.parse.urlparse(url)
+        explicit_port = parsed.port
+    except ValueError:
+        return 'unparseable URL'
```
mykytanetipa: please avoid propagation of errors as raw strings, if the intent is logging only please do it inplace with logger.warning

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    if parsed.scheme not in ('http', 'https'):
+        return f"scheme '{parsed.scheme}' is not http/https"
```
mykytanetipa: use logger.warning

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    host = parsed.hostname
+    if not host:
+        return 'no hostname'
```
mykytanetipa: use logger.warning

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    port = explicit_port or (443 if parsed.scheme == 'https' else 80)
+    try:
+        loop = asyncio.get_running_loop()
+        infos = await loop.getaddrinfo(host, port, type=socket.SOCK_STREAM)
+    except OSError:
+        return f"host '{host}' could not be resolved"
```
mykytanetipa: use logger.warning

src/a2a/utils/push_url_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    for info in infos:
+        if _ip_is_blocked(str(info[4][0])):
+            return f"host '{host}' resolves to a non-public address"
```
mykytanetipa: use logger.warning

mykytanetipa (commented):
thank you for the contribution, couple comments from my side

tests/server/tasks/test_inmemory_push_notifications.py
```diff
@@ -531,7 +531,6 @@ def setUp(self) -> None:
         self.mock_httpx_client.post.return_value = mock_response
 
         self.config_store = InMemoryPushNotificationConfigStore()
-
```
mykytanetipa, line 534: minor: unrelated edit

SashaMIT:
Thanks Iva, and thanks Mykyta.

I dropped the sender `__all__` so `BasePushNotificationSender` is exported only from `a2a.server.tasks`. The helper is now `validate_push_notification_url` and returns `True` when the URL is safe. Rejects log with `logger.warning` in the validator and no longer travel as raw strings. The handler hook is `Callable[[str], Awaitable[bool]]`. Restored the unrelated blank line in the in-memory dispatch test.

sokoliva (approved):

SashaMIT:
Thanks Iva. Glad the dispatch-time check landed next to #1173. Appreciate the reviews from you and Mykyta.

Issue #666, opened 2026-02-03
[Task]: 1.0: Run TCK tests against 1.0 implementation


ishymko:
Currently some TCK tests are running in CI ([run-tck.yaml](https://github.com/a2aproject/a2a-python/blob/main/.github/workflows/run-tck.yaml)) for JSON-RPC (see #639).

This should be done as a part for 1.0 migration as well.

It is setup for Java 1.0 RC [here](https://github.com/a2aproject/a2a-java/blob/main/.github/workflows/run-tck-1.0-wip.yml) ~against a [a2a-tck/spec_1.0](https://github.com/a2aproject/a2a-tck/tree/spec_1.0) branch~ TCK merged 1.0 tests to `main`.

| Category | JSON-RPC | REST | gRPC |
| :--- | :---: | :---: | :---: |
| **Mandatory**| ⬜  | ⬜ | ⬜ |
| **Capabilities** | ⬜  | ⬜ | ⬜ |
| **Quality**| ⬜ | ⬜ | ⬜ |
| **Features** | ⬜ | ⬜ | ⬜ |

kuangmi-bit:
Hi — I'd like to pick this up. Quick context on why this is a good fit:

- We have 3 merged contributions in a2a-tck (#196 CARD-SIGN-003 protected-header structural tests, #197 transport-layer extension activation, plus the eventqueue conformtest package work in a2a-go), so we're familiar with the 1.0 TCK test structure and how to wire a SUT against it.
- Current state: `run-tck.yaml` pins `TCK_VERSION: 0.3.0.beta3` — still exercising the 0.3 TCK. The 1.0 TCK is available (a2a-tck is at version `1.0.0` on main, tag `1.0.0.alpha2`).

Proposed plan (mirroring the Java SDK's `run-tck-1.0-wip.yml`):

1. Bump `TCK_VERSION` to a 1.0 build and run the JSON-RPC SUT first (smallest surface, already has 0.3 coverage) — fix any compatibility gaps the 1.0 tests surface.
2. Then enable REST and gRPC rows in the matrix as they go green.
3. Report per-cell status against the checklist in this issue (Mandatory/Capabilities/Quality/Features × JSON-RPC/REST/gRPC).

Two questions before I start:

1. **Version pin** — use `1.0.0.alpha2`, or wait for a final `1.0.0` release? Happy to pin whatever maintainers prefer.
[...]

PR #1165, opened 2026-08-06
fix(tck): enqueue Task before TaskStatusUpdateEvent in SUT agent

kuangmi-bit:
## Summary

Single-file fix to the TCK SUT (`tck/sut_agent.py`) that unblocks the A2A 1.0 TCK conformance run (part of #666):

1. **Enqueue the Task before emitting `TaskStatusUpdateEvent`.** The 1.0 SDK's active-task machinery requires the Task object to be enqueued first (it raises `InvalidAgentResponseError` otherwise), and the 1.0 TCK `CORE-SEND-*` requirements assert this ordering. When the request context does not yet carry a task, the SUT now creates one via `new_task_from_user_message` and enqueues it before processing the message.

2. **AgentInterface metadata aligned with the 1.0 spec:** the HTTP interface now reports `protocol_binding='HTTP+JSON'` (the 1.0 binding name) instead of the legacy `'REST'`, and the gRPC interface URL is a bare `host:port` target (no scheme prefix), matching what the SDK's gRPC channel expects.

## Verification

Ran the 1.0 TCK (`a2a-tck` tag `1.0.0.alpha2`, jsonrpc transport, must level) against the SUT locally:

```
before: 53 failed, 16 passed
after:   6 failed, 67 passed
```

The remaining 6 failures are not transport issues — they are SUT feature gaps (artifact-carrying responses, DM-ART-001) and one SDK error-code mapping gap (`ContentTypeNotSupportedError` reported as `ParseError`). Those are tracked as follow-ups; this PR removes the transport-level blockers.

## Scope notes

- Trailing-slash endpoint handling and any client-side redirect changes are intentionally **not** part of this PR. Per review, the redirect-following half lives in the harness: a2aproject/a2a-tck#230 (merged).
- No SDK client defaults are changed; `ClientFactory` behavior is identical to `main`.

## Related

- #666 (1.0 TCK migration)
- a2aproject/a2a-tck#230 (harness-side `follow_redirects=True`)

BEGIN_COMMIT_OVERRIDE
ci: enqueue Task before TaskStatusUpdateEvent in SUT agent (https://github.com/a2aproject/a2a-python/pull/1165)
END_COMMIT_OVERRIDE


kuangmi-bit:
Gentle ping on review. This is a small, TCK-unblocking server-side change (part of #666): `create_jsonrpc_routes` now accepts the trailing-slash endpoint variant — httpx normalizes an empty request path to a trailing slash, so `POST /a2a/jsonrpc/` currently 404s — and enqueues the Task rather than failing the request. Coverage report attached, all checks green. Happy to address any review feedback.

mykytanetipa (approved):

src/a2a/server/routes/jsonrpc_routes.py
```diff
@@ -64,5 +64,10 @@ def create_jsonrpc_routes(
             path=rpc_url,
             endpoint=dispatcher.handle_requests,
             methods=['POST'],
-        )
+        ),
+        Route(
+            path=f'{rpc_url}/',
+            endpoint=dispatcher.handle_requests,
+            methods=['POST'],
+        ),
```
mykytanetipa: major: adding a duplicated route is a behavioral change for existing clients that rely on public create_jsonrpc_routes, I would avoid this.

the route mismatch issue should be fixed from the clients side e.g. by using follow_redirects=True on the httpx.Client.

src/a2a/server/routes/jsonrpc_routes.py
```diff
@@ -64,5 +64,10 @@ def create_jsonrpc_routes(
             path=rpc_url,
             endpoint=dispatcher.handle_requests,
             methods=['POST'],
-        )
+        ),
+        Route(
+            path=f'{rpc_url}/',
```
mykytanetipa: will introduce malformed unreachable path in case of rpc_url='/'

mykytanetipa (changes requested):
Retracting my earlier approval - it was submitted in error while I was reviewing itk PRs in parallel. Switching to request-changes; see the points below.

kuangmi-bit:
@mykytanetipa — both points are right, thank you for catching them. Reworked per your direction (head `0c2894c`).

## What changed

1. **Duplicated route removed.** `create_jsonrpc_routes` returns a single `Route` again — no behavioral change for callers. And you were spot-on about `rpc_url='/'`: the FastAPI mount (`fastapi_routes.py:125`) passes exactly that, so `f'{rpc_url}/'` would have registered a `'//'` route in the most common integration. That was a real break I missed.

2. **Client-side fix, as you suggested.** The factory-created httpx client now sets `follow_redirects=True` (`client_factory.py`). Rationale: Starlette's default `redirect_slashes` answers the trailing-slash variant with a **307**, which preserves method and body — so following it is safe for JSON-RPC POSTs, and the client becomes robust to either spelling without the server caring. Users who pass their own `httpx_client` keep their own redirect policy.

3. **Regression tests**: default client follows redirects; custom client policy is respected.

## One honest caveat

The original 53→6 TCK improvement came from the server answering the trailing-slash spelling directly. The client that actually *hits* that spelling in the TCK run is the **a2a-tck harness**, not the SDK: `tck/transport/jsonrpc_client.py` posts to `"/"` (i.e. `base_url + "/"`) with httpx defaults (no redirect following). That harness lives in a2a-tck, so I'll add `follow_redirects=True` to the a2a-tck harness clients (`jsonrpc_client.py`, `http_json_client.py`) as a small follow-up PR there — that is what actually unblocks the SUT conformance run. This PR's client change is the SDK-side half: correct HTTP behavior for any SDK user whose server redirects.

The task-enqueue fix in the SUT (`new_task_from_user_message` ordering) is unchanged — separate from the route question.

kuangmi-bit:
@mykytanetipa — gentle nudge for a re-review when you have a moment. Both points from your 08-20 review are addressed at head `0c2894c`:

1. **Duplicated route removed** — `create_jsonrpc_routes` returns a single `Route` again; no behavioral change for callers.
2. Second point (trailing-slash variant) reworked as you directed.

Per your earlier guidance, the test-harness half of the fix (httpx `follow_redirects=True` in the TCK clients) is split out as a2aproject/a2a-tck#230, so this PR stays a minimal server-side change. CI is green across all 15 checks.

src/a2a/client/client_factory.py
```diff
@@ -78,7 +78,13 @@ def __init__(
...
     ):
         config = config or ClientConfig()
-        httpx_client = config.httpx_client or httpx.AsyncClient()
+        # Follow redirects by default: a server may 307 a trailing-slash
+        # variant of the JSON-RPC/REST endpoint (Starlette's default
+        # redirect_slashes). 307 preserves method and body, so this is safe
+        # for POST payloads and keeps the client robust to either spelling.
+        httpx_client = config.httpx_client or httpx.AsyncClient(
+            follow_redirects=True
+        )
```
mykytanetipa: This is a behavioral change for existing clients.

The default now follows all redirects, not only the intended /x -> /x/ case. A server that previously returned a 301/302 (which the client used to surface as a redirect response) will now be transparently followed, and for 301/302 httpx downgrades POST -> GET and drops the body. Affects existing client that depended on seeing the 3xx e.g. custom redirect handling, or asserting on status codes.

src/a2a/server/routes/jsonrpc_routes.py
```diff
@@ -64,5 +64,5 @@ def create_jsonrpc_routes(
             path=rpc_url,
             endpoint=dispatcher.handle_requests,
             methods=['POST'],
-        )
+        ),
```
mykytanetipa: minor: trailing comma, please remove

src/a2a/client/client_factory.py
```diff
@@ -78,7 +78,13 @@ def __init__(
         config: ClientConfig | None = None,
     ):
         config = config or ClientConfig()
-        httpx_client = config.httpx_client or httpx.AsyncClient()
+        # Follow redirects by default: a server may 307 a trailing-slash
+        # variant of the JSON-RPC/REST endpoint (Starlette's default
+        # redirect_slashes). 307 preserves method and body, so this is safe
+        # for POST payloads and keeps the client robust to either spelling.
+        httpx_client = config.httpx_client or httpx.AsyncClient(
```
mykytanetipa: major: do we need to change the default here? Anyone who wants redirect-following can already opt in today without this PR:
```
ClientFactory(ClientConfig(httpx_client=httpx.AsyncClient(follow_redirects=True)))
```

This is what I ment previously by saying "the route mismatch issue should be fixed from the clients side"

mykytanetipa (commented):
not as important, but isn't the TCK's JSON-RPC client constructed directly rather than through a2a's ClientFactory? (https://github.com/a2aproject/a2a-tck/blob/5996b79f9cefa6fc390980e383e358a66fb9e49e/tck/transport/jsonrpc_client.py#L81-L85)

kuangmi-bit:
@mykytanetipa — thanks for both rounds of review, and good catch on the TCK client construction. Reworked at head `cd62e3b`, scope narrowed:

## What changed

1. **ClientFactory default reverted** (your 08-30 major point). `follow_redirects` is back to httpx's default (`False`); the PR no longer touches the client default at all. Anyone who wants redirect-following can opt in today via `ClientFactory(ClientConfig(httpx_client=httpx.AsyncClient(follow_redirects=True)))` — unchanged from `main`.
2. **Trailing comma removed** from `create_jsonrpc_routes` (minor point).
3. **Both redirect tests dropped** — they asserted the reverted default, so they'd be testing the old behavior.
4. **The PR now contains only the TCK SUT fix**: `tck/sut_agent.py` enqueues the Task object *before* emitting `TaskStatusUpdateEvent`, which the 1.0 SDK requires (and which #666's harness exercises).

## On your 08-30 question

Correct — the TCK's JSON-RPC client is constructed directly (`tck/transport/jsonrpc_client.py#L81-L85` builds `httpx.AsyncClient()` itself, not via `ClientFactory`), so a client-side default change in the SDK wouldn't even have reached the harness. That confirms the split-fix shape:

- **This PR**: server-side SUT change only (Task enqueue ordering) — no client behavior change.
- **a2a-tck #230**: harness-side `follow_redirects=True` on the direct httpx clients, so the trailing-slash variant (Starlette 307 `/` → `//`... actually `//` → `/`) is followed there. It's already open, green, and linked in my last comment on that repo.

So the remaining diff is a 4-line SUT correctness fix with no public-API impact. Appreciate a re-review when you have a moment.

tck/sut_agent.py
```diff
@@ -157,11 +165,11 @@ def serve(task_store: TaskStore) -> None:
             ),
             AgentInterface(
                 url=f'http://localhost:{http_port}{REST_URL}',
-                protocol_binding='REST',
+                protocol_binding='HTTP+JSON',
                 protocol_version='1.0.0',
             ),
             AgentInterface(
-                url=f'http://localhost:{grpc_port}',
```
mykytanetipa, line 164: wont this break current TCK CI on a2a 0.3.x (e.g. assertions like https://github.com/a2aproject/a2a-tck/blob/0.3.0.beta3/tests/optional/capabilities/test_agent_card_optional.py#L286)? 

This will leave CI both not working on 0.3.x and 1.x.x, I think we should submit the full 1.x.x migration work at the same time (i.e. with TCK_VERSION v1) but it would include more changes. (please provide explanation, I may lack context on the TCK migration plan)

mykytanetipa (commented):
pr description still mentions "create_jsonrpc_routes accepts the trailing-slash endpoint variant" and other work that was reworked

kuangmi-bit:
@mykytanetipa — thank you, you're right. PR description updated at head `cd62e3b` to match the actual scope. The stale items are gone: the trailing-slash route work and the client-default work were both reworked out of this PR per your earlier direction (the redirect-following half now lives in a2aproject/a2a-tck#230, merged 08-31). The description now covers only what the diff contains:

1. SUT enqueues the Task before `TaskStatusUpdateEvent` (1.0 ordering requirement, exercised by #666's harness).
2. AgentInterface metadata fixes: `HTTP+JSON` binding name and bare `host:port` gRPC URL.

No SDK client defaults are touched. Appreciate a re-review when you have a moment.

kuangmi-bit:
@mykytanetipa — gentle nudge for a re-review when you have a moment. Since your last COMMENTED rounds (08-30/08-31), both points are addressed at head `5a819ee`:

- PR description rewritten to match the actual scope (SUT enqueues Task before TaskStatusUpdateEvent; AgentInterface metadata corrections; trailing-slash/redirect work explicitly scoped out to the harness half, a2a-tck #230 — already merged 08-31)
- Branch just updated onto current `main` (now includes #1173/#1164/#1212), CI all green, no conflicts

The 08-20 CHANGES_REQUESTED was noted as an approval retraction ("submitted in error"), so this is just a request for a fresh look at the narrowed scope. Thanks for the review bandwidth.

kuangmi-bit:
@mykytanetipa — one more nudge on the narrowed scope at head `5a819ee`, plus a low-cost exit if the shape still isn't right.

Current state:

- Both points from your 08-20 review are addressed: single route restored (no duplicated route), and the trailing-slash/redirect work is explicitly scoped out to the harness half — `a2aproject/a2a-tck#230`, merged 08-31.
- PR description rewritten to match the actual diff (SUT enqueues Task before `TaskStatusUpdateEvent`; AgentInterface metadata corrections).
- Branch updated onto current `main`; CI green; `MERGEABLE`, no conflicts.
- The 08-20 `CHANGES_REQUESTED` was an approval retraction ("submitted in error"), so the only thing outstanding is a fresh look.

If the current shape isn't what the repo wants, **closing it with a one-line note is a perfectly fine outcome** — I'd rather that than leave it sitting in the queue. If it's close, tell me which part to cut and I'll narrow it further, or split it into a smaller PR.

mykytanetipa (approved):
LGTM (partial migration scope)

PR #1167, opened 2026-08-06
fix(grpc): correct error code mapping and enforce A2A-Version validation

### LINKED ISSUE #1166: [Bug]: grpc_handler maps 3 A2A errors to FAILED_PRECONDITION instead of UNIMPLEMENTED, and doesn't validate A2A-Version at all
### What happened?

`grpc_handler.py`'s `_ERROR_CODE_MAP` (`src/a2a/server/request_handlers/grpc_handler.py:84-98`) maps 3 of the 9 A2A error types to the wrong gRPC status code, diverging from the canonical binding table in the spec (`specification.md` "Custom Binding Requirements" / error-code-mappings section):

| A2A Error Type | Spec gRPC Status | SDK gRPC Status |
|---|---|---|
| `PushNotificationNotSupportedError` | `UNIMPLEMENTED` | `FAILED_PRECONDITION` ❌ |
| `UnsupportedOperationError` | `UNIMPLEMENTED` | `FAILED_PRECONDITION` ❌ |
| `VersionNotSupportedError` | `UNIMPLEMENTED` | `FAILED_PRECONDITION` ❌ |

```python
_ERROR_CODE_MAP: dict[type[A2AError], grpc.StatusCode] = {
    ...
    types.PushNotificationNotSupportedError: grpc.StatusCode.FAILED_PRECONDITION,
    types.UnsupportedOperationError: grpc.StatusCode.FAILED_PRECONDITION,
    ...
    types.VersionNotSupportedError: grpc.StatusCode.FAILED_PRECONDITION,
}
```

The other 6 mappings (`TaskNotFoundError`, `TaskNotCancelableError`, `ContentTypeNotSupportedError`, `InvalidAgentResponseError`, `ExtendedAgentCardNotConfiguredError`, `ExtensionSupportRequiredError`) are all correct against the same table.

### Second, related bug: gRPC transport doesn't validate `A2A-Version` at all

`jsonrpc_dispatcher.py` and `rest_dispatcher.py` both guard their handlers with `@validate_version(constants.PROTOCOL_VERSION_1_0)`, which raises `VersionNotSupportedError` for an unsupported `A2A-Version`. `grpc_handler.py` has no equivalent — a `SendMessage` gRPC call with metadata `a2a-version: 99.0` is processed normally instead of being rejected. Per spec (`specification.md:739`): "*Agents MUST process requests using the semantics of the requested `A2A-Version`... If the version is not supported by the interface, agents MUST return a `VersionNotSupportedError`.*"

### Reproduction

Against a standard `DefaultRequestHandlerV2`-backed gRPC server:

```python
import grpc
metadata = [("a2a-version", "99.0")]
client.stub.SendMessage(proto_request, metadata=metadata)
# succeeds — no VersionNotSupportedError, no version check at all
```

For the status-code mismatch, trigger any `PushNotificationNotSupportedError` (e.g. call `CreatePushNotificationConfig` when `capabilities.push_notifications` is `false`) over gRPC and inspect `e.code()` — it's `FAILED_PRECONDITION`, not `UNIMPLEMENTED`.

We hit both while running the official `a2a-tck` MUST-level gRPC suite against a reference SUT built on this SDK — `GRPC-ERR-002` fails on `test_push_not_supported_returns_unimplemented` and `test_version_not_supported_returns_unimplemented` for exactly these reasons.

### Expected

- Fix the 3 entries in `_ERROR_CODE_MAP` to `UNIMPLEMENTED`, matching the other 6 and the spec table.
- Apply the same version-validation the HTTP dispatchers already do to the gRPC transport (a `@validate_version`-equivalent, or an explicit check in `GrpcHandler` reading the `a2a-version` metadata key).

Happy to send a PR for both if useful.

avilleroy51:
## Summary

Fixes #1166 — two bugs in the gRPC transport, found while cross-checking `grpc_handler.py` against `specification.md`'s normative error-code table and against the version-validation behavior already enforced on JSON-RPC/REST.

1. **Wrong error code mapping.** `PushNotificationNotSupportedError`, `UnsupportedOperationError` and `VersionNotSupportedError` were mapped to `FAILED_PRECONDITION` in `_ERROR_CODE_MAP`. Per `specification.md:1178-1192` (the normative A2A-error → gRPC-status table) all three should map to `UNIMPLEMENTED`.
2. **`A2A-Version` never validated on gRPC.** JSON-RPC and REST dispatchers already enforce version compatibility via `@validate_version(PROTOCOL_VERSION_1_0)`. The gRPC transport had no equivalent — `DefaultGrpcServerCallContextBuilder.build()` never exposed `invocation_metadata()` to the rest of the request-handling code, so there was no way for any check to see the caller's declared version. This adds `state['headers']` (same shape the HTTP route builders already populate) and a single `_validate_a2a_version()` check applied where all 11 RPC methods funnel through `_build_call_context`, replicating `validate_version`'s semantics (missing header ⇒ treated as `0.3`; major version must match `1.0`).

## Test plan

- [x] `tests/server/request_handlers/test_grpc_handler.py` alone: 43/43 passed
- [x] Full suite (excluding `tests/integration`, `tests/install_smoke`, `tests/compat` — out of scope — and modules with pre-existing, unrelated `ModuleNotFoundError`: `respx`, `opentelemetry`, migrations): 997 passed, 0 failed
- [x] `ruff check` on both changed files: all checks passed
- [x] `mock_grpc_context` fixture now defaults `invocation_metadata.return_value` to `a2a-version: 1.0` (same convention `TestClient(..., headers={'A2A-Version': '1.0'})` already uses for JSON-RPC/REST tests)
- [x] The 2 pre-existing `test_abort_context_error_mapping` cases that asserted the old `FAILED_PRECONDITION` mapping updated to `UNIMPLEMENTED`
- [x] The 3 `TestGrpcExtensions` cases that override `invocation_metadata` updated to include `('a2a-version', '1.0')`

### DIFF
--- a/src/a2a/server/request_handlers/grpc_handler.py
+++ b/src/a2a/server/request_handlers/grpc_handler.py
@@ -3,7 +3,7 @@ import logging
 
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterable, Awaitable, Callable
-from typing import TypeVar
+from typing import Any, TypeVar
 
 
 try:
@@ -18,6 +18,7 @@ except ImportError as e:
 
 from google.protobuf import any_pb2, empty_pb2, message
 from google.rpc import error_details_pb2, status_pb2
+from packaging.version import InvalidVersion, Version
 
 import a2a.types.a2a_pb2_grpc as a2a_grpc
 
@@ -30,7 +31,7 @@ from a2a.extensions.common import (
 from a2a.server.context import ServerCallContext
 from a2a.server.request_handlers.request_handler import RequestHandler
 from a2a.types import a2a_pb2
-from a2a.utils import proto_utils
+from a2a.utils import constants, proto_utils
 from a2a.utils.errors import A2A_ERROR_REASONS, A2AError, TaskNotFoundError
 from a2a.utils.grpc_status import status_to_grpc
 from a2a.utils.proto_utils import validation_errors_to_bad_request
@@ -52,7 +53,19 @@ class DefaultGrpcServerCallContextBuilder(GrpcServerCallContextBuilder):
 
     def build(self, context: grpc.aio.ServicerContext) -> ServerCallContext:
         """Builds a ServerCallContext from a gRPC ServicerContext."""
-        state = {'grpc_context': context}
+        state: dict[str, Any] = {'grpc_context': context}
+        # Mirror what the HTTP route builders do (`state['headers'] =
+        # dict(request.headers)`, see a2a.server.routes.common): helpers such
+        # as `validate_version` read the A2A-Version header out of
+        # `context.state['headers']`. Without this key the gRPC transport has
+        # no way to see request metadata at all. Keys are lowercased because
+        # gRPC metadata keys are case-insensitive and normalized to lowercase.
+        state['headers'] = {
+            key.lower(): (
+                value.decode('utf-8') if isinstance(value, bytes) else value
+            )
+            for key, value in (context.invocation_metadata() or ())
+        }
         return ServerCallContext(
             user=self.build_user(context),
             state=state,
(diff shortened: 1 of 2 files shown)


avilleroy51:
Hi maintainers — checking in on this PR. It looks like no CI workflow has run yet on this commit (no check-suites registered), which is expected for a first-time external contributor — GitHub requires a maintainer to approve running workflows on fork PRs.

Could someone please approve the workflow run (or take a look) when you have a moment? Happy to address any feedback. Thanks!

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -88,16 +101,59 @@ def _get_metadata_value(
...
-    types.PushNotificationNotSupportedError: grpc.StatusCode.FAILED_PRECONDITION,
-    types.UnsupportedOperationError: grpc.StatusCode.FAILED_PRECONDITION,
+    types.PushNotificationNotSupportedError: grpc.StatusCode.UNIMPLEMENTED,
+    types.UnsupportedOperationError: grpc.StatusCode.UNIMPLEMENTED,
     types.ContentTypeNotSupportedError: grpc.StatusCode.INVALID_ARGUMENT,
     types.InvalidAgentResponseError: grpc.StatusCode.INTERNAL,
     types.ExtendedAgentCardNotConfiguredError: grpc.StatusCode.FAILED_PRECONDITION,
     types.ExtensionSupportRequiredError: grpc.StatusCode.FAILED_PRECONDITION,
-    types.VersionNotSupportedError: grpc.StatusCode.FAILED_PRECONDITION,
+    types.VersionNotSupportedError: grpc.StatusCode.UNIMPLEMENTED,
```
mykytanetipa, line 110: The error code remapping (FAILED_PRECONDITION -> UNIMPLEMENTED) contradicts the current spec (v1.0.1 and main), which reverted these to FAILED_PRECONDITION in spec PR #1627 (https://github.com/a2aproject/A2A/pull/1627) which was merged 2026-04-14. The SDK's existing values are correct; this change should be dropped.

src/a2a/server/request_handlers/grpc_handler.py
```diff
@@ -88,16 +101,59 @@ def _get_metadata_value(
...
 }
 
 
+def _validate_a2a_version(server_context: ServerCallContext) -> None:
```
mykytanetipa, line 114: This duplicates existing logic in src/a2a/utils/version_validator.py (_get_actual_version (partially), _is_version_compatible, and the identical VersionNotSupportedError message), which will lead to diverging implementations of the same rule.

Consider extracting version_validator.py helper functions to module scope and reusing them in both places

mykytanetipa (changes requested):

Issue #1170, opened 2026-08-08
Cancel reports success but writes no terminal state (DefaultRequestHandlerV2)


technicalpickles:
Cancelling a task can leave it with no terminal state at all. Two paths get there, and the V1 handler guarded both.

## What happens

Against `DefaultRequestHandlerV2` (what `DefaultRequestHandler` aliases to), with an executor whose `cancel()` is empty:

```
mid-run cancel:  cancel returned WORKING         later get_task: WORKING (forever)
parked task:     cancel returned INPUT_REQUIRED  task unchanged
```

Neither call errors. In the first case the task never reaches a terminal state at all, so a client that cancelled has no way to know it can stop polling. In the second, cancel reports success and does nothing.

## Path 1: mid-run, ordering in `ActiveTask.cancel`

`active_task.py:733-737` cancels the producer first, then awaits the executor's cancel:

```python
self._producer_task.cancel()
try:
    await self._agent_executor.cancel(request_context, self._event_queue_agent)
```

The producer is the task running `execute()`, and it's the only component that writes the task's terminal state. Killing it first leaves nothing to write one.

V1 does this the other way round: `default_request_handler.py:213` awaits `agent_executor.cancel(...)`, and line 224 cancels the producer after.

## Path 2: parked task, missing guard in `on_cancel_task`

`ActiveTask.cancel` only acts `if not self._is_finished.is_set() and self._producer_task` (`active_task.py:729`), so it treats cancelable as "has a running producer" rather than "is not terminal". A task parked in `input-required` has already returned from `execute()`, so it takes the `else` branch, logs "not cancelling", and returns the task untouched.

V1 caught that on the way out (`default_request_handler.py:233`):

```python
if result.status.state != TaskState.TASK_STATE_CANCELED:
    raise TaskNotCancelableError(...)
```

V2's `on_cancel_task` (`default_request_handler_v2.py:161-181`) returns `result` whatever it is. The terminal-state pre-check at `default_request_handler.py:195` doesn't cover this, since `input-required` isn't terminal.

I'm not assuming an answer on whether a parked task should be cancelable. Either outcome works as long as it's visible: cancel it, or raise `TaskNotCancelableError`. Reporting success and changing nothing is the one result a caller can't act on.

## Why the tests didn't catch it

[...]

Issue #1159, opened 2026-08-03
[Bug]: SubscribeToTask and CancelTask bypass owner-aware TaskStore for live tasks


andreeyka:
### What happened?

`DefaultRequestHandlerV2.on_subscribe_to_task` and `on_cancel_task` resolve the task through `ActiveTaskRegistry` by `task_id` alone. `get_or_create()` returns the cached `ActiveTask` before the owner-aware `TaskStore` is consulted, so a caller other than the task owner can subscribe to and cancel another user's live task.

Owner isolation itself works: `GetTask` and `SendMessage` consult `task_store.get(task_id, context)` first and correctly raise `TaskNotFoundError`. Only the two registry-backed paths skip it, so behaviour is inconsistent across request paths. This narrows the task-authorization half of #786, which predates owner-aware stores and `resolve_user_scope`.

Practical exploitability is low: `task_id` defaults to `uuid4()`, so it cannot be guessed or enumerated — the caller must already know the id.

**Expected:** `SubscribeToTask` / `CancelTask` for a task outside the caller's partition raise `TaskNotFoundError`, the same response as for a nonexistent task.

**Actual:** the call succeeds. The subscriber receives the owner's `Task` and event stream; `CancelTask` cancels the task and invokes `AgentExecutor.cancel()`.

**Affected:** a2a-sdk 1.1.2 (latest) and current `main`.

**Root cause**

- `active_task_registry.py:52-53` — `get_or_create()` returns `self._active_tasks[task_id]` with no owner check; `call_context` is not consulted on that path.
- `default_request_handler_v2.py:169` (`on_cancel_task`) and `:398` (`on_subscribe_to_task`) call `get_or_create()` without a prior `task_store.get(task_id, context)`. `_setup_active_task` does perform it (`:203`), which is why `SendMessage` is unaffected.

Preconditions: the task must still be live (present in `ActiveTaskRegistry`) and the caller must know its `task_id`. Tasks that are no longer active are isolated correctly.

**Side effect:** `ActiveTask.cancel()` assigns the caller's context to `TaskManager._call_context` (`active_task.py:718`), so the cancelled task is additionally persisted into the caller's store partition.

**Reproduction** — Python 3.13, `pip install a2a-sdk==1.1.2` (no extras):

```python
"""Minimal reproduction: owner-isolation bypass for live tasks in a2a-sdk 1.1.2.

`SubscribeToTask` and `CancelTask` resolve the task through `ActiveTaskRegistry`
[...]

PR #1171, opened 2026-08-08
test(server): add xfail repros for cancel leaving no terminal state

technicalpickles:
Adds two failing scenarios for #1170, both `xfail(strict=True)` so they sit green in CI today and flip loudly the moment someone fixes the underlying bug.

No fix here, just the repro. #1170 has the analysis and two possible fix shapes, and which one you want depends on calls that are yours to make (particularly whether the framework should close a task out when the executor didn't, which an existing TODO in this file already raises).

## What they cover

Both drive `DefaultRequestHandlerV2` through the real client with an executor whose `cancel()` is empty:

- `test_scenario_19_mid_run_cancel_reaches_a_terminal_state` cancels mid-run. The task stays `working` forever, so a caller has no way to know it can stop polling. The assertion is deliberately weak: any terminal state passes, not specifically `canceled`.
- `test_scenario_19_cancel_of_parked_task_does_not_silently_succeed` cancels a task parked in `input-required`. Cancel returns success with the task unchanged. This one asserts only that the state must not come back unchanged, so it stays agnostic on whether a parked task should be cancelable. Cancelling it or raising `TaskNotCancelableError` would both pass.

## Why an empty `cancel()`

`test_scenario_cancel_working_task_empty_cancel` already covers the mid-run case and passes, because its executor hand-enqueues the `CANCELED` event, right below a `# TODO: this should be done automatically by the framework ?` comment. These two are that same scenario with the hand-written event removed.

`InputRequiredAgent`, `SlowAgent` and `DummyAgentExecutor` in this file all define `cancel()` as `pass` already, so the shape isn't unusual. It's also what a real executor looks like when its cancel path only does teardown.

## Notes

Only `DefaultRequestHandlerV2`, no `use_legacy` parametrize. The legacy handler hangs on both of these rather than failing, since `on_cancel_task` waits in `consume_all` for an event the empty executor never sends. That seemed like a separate thing and not worth entangling with this.

Pure addition, 117 lines, no existing test touched and no new dependencies. `ruff check` and `ruff format --check` are clean. `tests/integration` goes from 427 passed / 1 xpassed to 427 passed / 2 xfailed / 1 xpassed, and the two new ones run in about 0.2s.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/integration/test_scenarios.py
+++ b/tests/integration/test_scenarios.py
@@ -2253,3 +2253,120 @@ async def test_scenario_context_id_visibility(use_legacy, streaming):
     # Verify that agent can see context_id the same as in task_1 and different task id
     assert agent_context2_id == agent_context1_id
     assert agent_task2_id != agent_task1_id
+
+
+# Scenario 19: cancel must close the task out when the executor does not.
+#
+# The empty-cancel scenario above only passes because its executor
+# hand-enqueues a CANCELED event, next to a "TODO: this should be done
+# automatically by the framework ?" comment. These two use an executor whose
+# cancel() really is empty, which is the shape InputRequiredAgent, SlowAgent
+# and DummyAgentExecutor all use in this file.
+_TERMINAL_STATES = {
+    TaskState.TASK_STATE_CANCELED,
+    TaskState.TASK_STATE_COMPLETED,
+    TaskState.TASK_STATE_FAILED,
+    TaskState.TASK_STATE_REJECTED,
+}
+
+
+async def _start_task(client, text):
+    it = client.send_message(
+        SendMessageRequest(
+            message=Message(
+                message_id='test-msg',
+                role=Role.ROLE_USER,
+                parts=[Part(text=text)],
+            ),
+            configuration=SendMessageConfiguration(return_immediately=True),
+        )
+    )
+    res = await it.__anext__()
+    return res.task.id if res.HasField('task') else res.status_update.task_id
+
+
+@pytest.mark.timeout(5.0)
+@pytest.mark.asyncio
+@pytest.mark.xfail(
+    strict=True,
+    reason='https://github.com/a2aproject/a2a-python/issues/1170 - '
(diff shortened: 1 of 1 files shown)


astrogilda:
Cross-linking a measurement, since a repro with no fix attached is only worth what someone does with it. I ran these two scenarios verbatim against #1172, which fixes #1170 and #1159 together, using two `git worktree` checkouts at `cff6727` and `a382ad5` with this file dropped into both.

On `main` both report `XFAIL` (exit 0). On #1172 both report `XPASS(strict)`, which pytest surfaces as a failure precisely because the bug they pin is gone, and with `--runxfail` both pass outright. So these hold as strict markers today and flip cleanly the moment the fix lands, which is exactly what you would want from them.

I have suggested on #1172 that these two be folded in there with the markers dropped, since they exercise the mid-run path through the client and transport rather than at the handler layer, which its own tests cover. Filing the repro separately and leaving the fix to someone else made the review of #1172 much cheaper than it would otherwise have been, so thank you for that.

technicalpickles:
#1172 already folds these two scenarios in directly (with the strict-xfail markers dropped, since it fixes what they pin), thanks for the credit there. I think this is superseded, happy to close in favor of #1172.

astrogilda:
Checked the fold rather than taking the summary for it, and it is faithful. Both scenarios are in the integration scenarios file on #1172 as `test_scenario_19_mid_run_cancel_reaches_a_terminal_state` and `test_scenario_19_cancel_of_parked_task_does_not_silently_succeed`, with the markers dropped and a comment saying why. That is the whole of what this issue was pinning, so there is nothing left here that #1172 does not carry.
Close it in favour of #1172, and thanks for the credit. Worth saying that the split is what made this cheap: a repro filed on its own, with no fix attached and a strict marker holding it honest, meant the fix could be reviewed against something that already failed for the stated reason instead of against a description of the bug.

technicalpickles:
Closing, it landed in a PR with the fix. Thanks!

PR #1172, opened 2026-08-09
fix: owner-scope cancel/subscribe and write terminal state on cancel (#1159, #1170)

### LINKED ISSUE #1159: [Bug]: SubscribeToTask and CancelTask bypass owner-aware TaskStore for live tasks

### LINKED ISSUE #1170: Cancel reports success but writes no terminal state (DefaultRequestHandlerV2)

chopmob-cloud:
## Summary

Two related fixes to `DefaultRequestHandlerV2`, shipped together because the first is a prerequisite for the second to be safe.

**Fixes #1159 (owner-scoping, the security base).** `on_cancel_task` and `on_subscribe_to_task` resolved a live task via `ActiveTaskRegistry.get_or_create(task_id)` by `task_id` alone, skipping the owner-aware `task_store.get(task_id, context)` that `on_get_task` and the send path already perform. A caller who knows another tenant's `task_id` could subscribe to its live stream and cancel it (CWE-639). Both handlers now consult the owner-aware store first and fail closed with `TaskNotFoundError` (masking existence, matching `on_get_task`).

**Fixes #1170 (cancel writes a terminal state).** `ActiveTask.cancel` cancelled the producer before `AgentExecutor.cancel`, so the component that owns the terminal state could not write it; and a task parked non-terminal (e.g. `input-required`) reported cancel-success without transitioning. The executor cancel now runs before the producer cancel (producer still cancelled on the error path), and cancellation that leaves the task non-terminal is closed out as `CANCELED`.

## Why one PR

#1170 alone makes #1159 worse: for a parked task, a cross-tenant cancel via the #1159 bypass changes from a harmless no-op into an actual cross-tenant `CANCELED` write. Landing #1159's owner check together (as the base commit) removes that window. Merging them together, or #1159 before #1170, is safe; #1170 must not land first.

## Tests

- Owner-scope regressions: non-owner cancel and non-owner subscribe are rejected (`TaskNotFoundError`), including a parked task; the owner still succeeds.
- Cancel-terminal regressions: mid-run cancel reaches a terminal state; cancel of a parked task does not silently succeed (the maintainer repros from #1171, converted from xfail to passing).

## Validation

- Owner-scope tests: reverting the #1159 handler change fails all three; restore passes. Full request-handler module: 62 passed.
- Cancel-terminal tests: reverting the #1170 change fails both scenario_19 tests; restore passes. Full integration module: no regression.
- Combined: with both applied, a cross-tenant cancel of a parked task is rejected rather than writing `CANCELED`.
- `ruff` clean at the pinned version.

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler_v2.py
+++ b/src/a2a/server/request_handlers/default_request_handler_v2.py
@@ -165,6 +165,15 @@ class DefaultRequestHandlerV2(RequestHandler):
     ) -> Task | None:
         task_id = params.id
 
+        # Owner-aware isolation: mirror on_get_task / _setup_active_task and
+        # consult the owner-scoped store before resolving the live task from
+        # the registry, which keys on task_id alone. Without this a non-owner
+        # who knows the uuid4 task_id could cancel another user's live task
+        # (issue #1159, CWE-639). Mask a partition mismatch as not-found so
+        # existence is not leaked.
+        if not await self.task_store.get(task_id, context):
+            raise TaskNotFoundError
+
         try:
             active_task = await self._active_task_registry.get_or_create(
                 task_id, call_context=context, create_task_if_missing=False
@@ -394,6 +403,13 @@ class DefaultRequestHandlerV2(RequestHandler):
         context: ServerCallContext,
     ) -> AsyncGenerator[Event, None]:
         task_id = params.id
+
+        # Owner-aware isolation before resolving the live task from the
+        # registry (which is keyed on task_id alone); see on_cancel_task.
+        # Prevents a non-owner from subscribing to another user's live
+        # task stream (issue #1159, CWE-639).
+        if not await self.task_store.get(task_id, context):
+            raise TaskNotFoundError
 
         active_task = await self._active_task_registry.get_or_create(
             task_id,
--- a/src/a2a/server/request_handlers/default_request_handler_v2.py
+++ b/src/a2a/server/request_handlers/default_request_handler_v2.py
@@ -165,15 +165,6 @@ class DefaultRequestHandlerV2(RequestHandler):
     ) -> Task | None:
         task_id = params.id
 
-        # Owner-aware isolation: mirror on_get_task / _setup_active_task and
-        # consult the owner-scoped store before resolving the live task from
-        # the registry, which keys on task_id alone. Without this a non-owner
-        # who knows the uuid4 task_id could cancel another user's live task
-        # (issue #1159, CWE-639). Mask a partition mismatch as not-found so
-        # existence is not leaked.
-        if not await self.task_store.get(task_id, context):
-            raise TaskNotFoundError
-
         try:
             active_task = await self._active_task_registry.get_or_create(
                 task_id, call_context=context, create_task_if_missing=False
@@ -403,13 +394,6 @@ class DefaultRequestHandlerV2(RequestHandler):
         context: ServerCallContext,
     ) -> AsyncGenerator[Event, None]:
         task_id = params.id
-
-        # Owner-aware isolation before resolving the live task from the
-        # registry (which is keyed on task_id alone); see on_cancel_task.
-        # Prevents a non-owner from subscribing to another user's live
-        # task stream (issue #1159, CWE-639).
-        if not await self.task_store.get(task_id, context):
-            raise TaskNotFoundError
 
         active_task = await self._active_task_registry.get_or_create(
             task_id,
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -730,7 +730,11 @@ class ActiveTask:
                 logger.debug(
                     'Cancel[%s]: Cancelling producer task', self._task_id
                 )
-                self._producer_task.cancel()
+                # Await the executor's cancel before cancelling the producer,
+                # so the component that owns the task's terminal state can
+                # still write to the still-open event queue. Cancelling the
+                # producer first can tear the queue down first and drop that
+                # write. Mirrors the V1 handler ordering.
                 try:
                     await self._agent_executor.cancel(
                         request_context, self._event_queue_agent
@@ -739,8 +743,10 @@ class ActiveTask:
                     logger.exception(
                         'Cancel[%s]: Agent cancel failed', self._task_id
                     )
+                    self._producer_task.cancel()
                     await self._mark_task_as_failed(e)
                     raise
+                self._producer_task.cancel()
             else:
                 logger.debug(
                     'Cancel[%s]: Task already finished [%s] or producer not started [%s], not cancelling',
@@ -753,6 +759,15 @@ class ActiveTask:
         task = await self._task_manager.get_task()
         if not task:
             raise RuntimeError('Task should have been created')
+        # A cleanup-only executor.cancel() may not write a terminal state, and
+        # a task parked in a non-terminal state (e.g. input-required) has no
+        # running producer to write one either. Close it out as CANCELED so a
+        # caller that cancelled is never left polling a live task. Mirrors the
+        # V1 handler, which made a non-cancelled outcome visible instead of
+        # reporting success and changing nothing.
+        if task.status.state not in TERMINAL_TASK_STATES:
+            task.status.state = TaskState.TASK_STATE_CANCELED
+            await self._task_manager.save_task_event(task)
         return task
 
     async def aclose(self) -> None:
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -734,7 +734,13 @@ class ActiveTask:
                 # so the component that owns the task's terminal state can
                 # still write to the still-open event queue. Cancelling the
                 # producer first can tear the queue down first and drop that
-                # write. Mirrors the V1 handler ordering.
+                # write. Mirrors the V1 handler ordering. The producer cancel
+                # is in a finally so it still runs when executor.cancel() raises
+                # a BaseException such as asyncio.CancelledError, which
+                # `except Exception` does not catch; otherwise the producer
+                # would leak as a pending task. Cancelling after
+                # _mark_task_as_failed also keeps the FAILED write ahead of the
+                # queue teardown, so it is not dropped as QueueShutDown.
                 try:
                     await self._agent_executor.cancel(
                         request_context, self._event_queue_agent
@@ -743,10 +749,10 @@ class ActiveTask:
                     logger.exception(
                         'Cancel[%s]: Agent cancel failed', self._task_id
                     )
-                    self._producer_task.cancel()
                     await self._mark_task_as_failed(e)
                     raise
-                self._producer_task.cancel()
+                finally:
+                    self._producer_task.cancel()
             else:
                 logger.debug(
                     'Cancel[%s]: Task already finished [%s] or producer not started [%s], not cancelling',
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -747,6 +747,15 @@ class ActiveTask:
                 # would leak as a pending task. Cancelling after
                 # _mark_task_as_failed also keeps the FAILED write ahead of the
                 # queue teardown, so it is not dropped as QueueShutDown.
+                #
+                # Residual: if executor.cancel() raises a BaseException (e.g.
+                # asyncio.CancelledError), it propagates out of cancel() from
+                # the finally before the _is_finished.wait() and the CANCELED
+                # write below, leaving the task non-terminal with no producer
+                # behind it. This is not the #1170 silent-success shape (the
+                # caller receives the exception), and it is recoverable: a
+                # later cancel() takes the else branch and reaches the
+                # terminal write.
                 try:
                     await self._agent_executor.cancel(
                         request_context, self._event_queue_agent
@@ -780,8 +789,12 @@ class ActiveTask:
         if task.status.state not in TERMINAL_TASK_STATES:
             # Write a copy rather than mutating the task in place: get_task()
             # returns the shared _task_manager._current_task, which may already
-            # have been yielded to a subscriber, and that reference must not
-            # change under the reader.
+            # have been yielded to a subscriber. Copying keeps THIS terminal
+            # write off the yielded reference. It is not a file-wide guarantee:
+            # ordinary status writes (save_task_event -> ensure_task -> CopyFrom,
+            # and the producer-failure write above) still update the shared
+            # object in place. Making every write copy-on-yield is the general
+            # property tracked in #1175 and #1191.
             updated = Task()
             updated.CopyFrom(task)
             updated.status.state = TaskState.TASK_STATE_CANCELED
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -707,6 +707,12 @@ class ActiveTask:
     async def cancel(self, call_context: ServerCallContext) -> Task:
         """Cancels the running active task.
 
+        The returned task carries a terminal state, which is written to the
+        task store. That write is not guaranteed to reach an active subscriber
+        stream: by the time cancel writes it the event queues may already be
+        closed, so a client that was streaming may have to re-read the task to
+        observe the terminal state.
+
         Concurrency Guarantee:
         Uses `_lock` to ensure we don't attempt to cancel a producer that is
         already winding down or hasn't started. It fires the cancellation signal
@@ -772,8 +778,15 @@ class ActiveTask:
         # V1 handler, which made a non-cancelled outcome visible instead of
         # reporting success and changing nothing.
         if task.status.state not in TERMINAL_TASK_STATES:
-            task.status.state = TaskState.TASK_STATE_CANCELED
-            await self._task_manager.save_task_event(task)
+            # Write a copy rather than mutating the task in place: get_task()
+            # returns the shared _task_manager._current_task, which may already
+            # have been yielded to a subscriber, and that reference must not
+            # change under the reader.
+            updated = Task()
+            updated.CopyFrom(task)
+            updated.status.state = TaskState.TASK_STATE_CANCELED
+            await self._task_manager.save_task_event(updated)
+            task = updated
         return task
 
     async def aclose(self) -> None:
--- a/src/a2a/server/agent_execution/active_task_registry.py
+++ b/src/a2a/server/agent_execution/active_task_registry.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
 
 from a2a.server.agent_execution.active_task import ActiveTask
 from a2a.server.tasks.task_manager import TaskManager
+from a2a.utils.errors import TaskNotFoundError
 
 
 logger = logging.getLogger(__name__)
@@ -50,25 +51,44 @@ class ActiveTaskRegistry:
         with self._lock:
             if self._closed:
                 raise RuntimeError('ActiveTaskRegistry is closed')
-            if task_id in self._active_tasks:
-                return self._active_tasks[task_id]
+            existing = self._active_tasks.get(task_id)
+            if existing is None:
+                task_manager = TaskManager(
+                    task_id=task_id,
+                    context_id=context_id,
+                    task_store=self._task_store,
+                    initial_message=initial_message,
+                    context=call_context,
+                )
 
-            task_manager = TaskManager(
-                task_id=task_id,
-                context_id=context_id,
-                task_store=self._task_store,
-                initial_message=initial_message,
-                context=call_context,
-            )
+                active_task = ActiveTask(
+                    agent_executor=self._agent_executor,
+                    task_id=task_id,
+                    task_manager=task_manager,
+                    push_sender=self._push_sender,
+                    on_cleanup=self._on_active_task_cleanup,
+                )
+                self._active_tasks[task_id] = active_task
 
-            active_task = ActiveTask(
-                agent_executor=self._agent_executor,
-                task_id=task_id,
-                task_manager=task_manager,
-                push_sender=self._push_sender,
-                on_cleanup=self._on_active_task_cleanup,
-            )
-            self._active_tasks[task_id] = active_task
+        if existing is not None:
+            # Owner-aware guard on the cache-hit path. The miss path below is
+            # owner-scoped by ActiveTask.start(), which reads through the task
+            # store with call_context and raises TaskNotFoundError when the
+            # task is not owned and create_task_if_missing is false. A cache
+            # hit returned before that check, so resolving a live task by id
+            # was an unauthenticated lookup (issue #1159, CWE-639): every
+            # call site had to guard first. Enforce it here instead, so the
+            # hit path is symmetric with the miss path and no future caller
+            # can reintroduce the gap. Skipped when create_task_if_missing is
+            # set, which is the on_message_send create path establishing
+            # ownership. Masked as not-found so existence is not leaked. Done
+            # outside _lock because the store read is I/O and the miss-path
+            # check runs outside the lock too.
+            if not create_task_if_missing and not await self._task_store.get(
+                task_id, call_context
+            ):
+                raise TaskNotFoundError
+            return existing
 
         await active_task.start(
             call_context=call_context,
--- a/tests/integration/test_scenarios.py
+++ b/tests/integration/test_scenarios.py
@@ -2253,3 +2253,105 @@ async def test_scenario_context_id_visibility(use_legacy, streaming):
     # Verify that agent can see context_id the same as in task_1 and different task id
     assert agent_context2_id == agent_context1_id
     assert agent_task2_id != agent_task1_id
+
+
+# Scenario 19 (issue #1170): cancel must leave the task in a state a caller can
+# act on. Both executors below define cancel() as an empty teardown, which is
+# what InputRequiredAgent, SlowAgent and DummyAgentExecutor use in this file and
+# what a real cleanup-only executor looks like.
+_TERMINAL_STATES = {
+    TaskState.TASK_STATE_CANCELED,
+    TaskState.TASK_STATE_COMPLETED,
+    TaskState.TASK_STATE_FAILED,
+    TaskState.TASK_STATE_REJECTED,
+}
+
+
+async def _start_task(client, text):
+    it = client.send_message(
+        SendMessageRequest(
+            message=Message(
+                message_id='test-msg',
+                role=Role.ROLE_USER,
+                parts=[Part(text=text)],
+            ),
+            configuration=SendMessageConfiguration(return_immediately=True),
+        )
+    )
+    res = await it.__anext__()
+    return res.task.id if res.HasField('task') else res.status_update.task_id
+
+
+@pytest.mark.timeout(5.0)
+@pytest.mark.asyncio
+async def test_scenario_19_mid_run_cancel_reaches_a_terminal_state():
+    """A caller who cancels should at least be able to stop polling."""
+    started = asyncio.Event()
+    hang = asyncio.Event()
+
+    class SilentCancelAgent(AgentExecutor):
+        async def execute(
+            self, context: RequestContext, event_queue: EventQueue
+        ):
+            task = new_task_from_user_message(context.message)
+            task.status.state = TaskState.TASK_STATE_WORKING
+            await event_queue.enqueue_event(task)
+            started.set()
+            await hang.wait()
+
+        async def cancel(
+            self, context: RequestContext, event_queue: EventQueue
+        ):
+            pass
+
+    client = await create_client(
+        create_handler(SilentCancelAgent(), use_legacy=False),
+        agent_card=agent_card(),
+        streaming=True,
+    )
+    task_id = await _start_task(client, 'hello')
+    await asyncio.wait_for(started.wait(), timeout=1.0)
+
+    await client.cancel_task(CancelTaskRequest(id=task_id))
+    task_after = await client.get_task(GetTaskRequest(id=task_id))
+
+    assert task_after.status.state in _TERMINAL_STATES
+
+
+@pytest.mark.timeout(5.0)
+@pytest.mark.asyncio
+async def test_scenario_19_cancel_of_parked_task_does_not_silently_succeed():
+    """input-required is non-terminal, so cancel should either work or raise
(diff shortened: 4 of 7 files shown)


chopmob-cloud:
Independently validated the #1170 half of this against the repros in #1171, which adds two strict-xfail scenarios (`test_scenario_19_mid_run_cancel_reaches_a_terminal_state` and `test_scenario_19_cancel_of_parked_task_does_not_silently_succeed`).

On a clean checkout of current `main` both xfail as expected (the bug is present); run against this branch they XPASS, and forced with `--runxfail` they pass outright: 2 passed on this branch versus 2 failed on `main`. This PR's own suite for the two changed files is green (174 passed, 0 failed).

So the repros in #1171 are effectively an independent check on the cancel-terminal-state fix here. If useful they could be folded into this PR with the xfail markers dropped, giving direct regression coverage alongside the owner-scoping (#1159) tests.

astrogilda (commented):
Outside contributor here, not a maintainer. I have one merged fix in `active_task.py` (#1105, the `aclose()` drain, shipped in 1.1.2) and one open PR in `event_queue_v2.py` (#1137), so this is a neighbourhood I have had my hands in rather than any kind of authority. I ran this branch rather than only reading it, and I found one thing I think is a real regression plus two things worth a maintainer's opinion. Everything below is reproducible from `a382ad5` and `cff6727`.

First, on whether the bugs are real. I took your three new handler tests and ran them against unmodified `main` with your source changes reverted: all three fail with `Failed: DID NOT RAISE <class 'a2a.utils.errors.TaskNotFoundError'>`, meaning Bob genuinely does cancel and subscribe to Alice's live task today. That is #1159 reproducing under its own tests, and it is worth saying plainly in a thread that has been quiet for six days: this is a live cross-tenant authorization gap, not a hypothetical. On this branch all three pass.

Second, the #1170 half, cross-checked against #1171 rather than against your own tests. I checked out `main` and this branch into two worktrees, dropped in #1171's `tests/integration/test_scenarios.py` verbatim in both, and ran only its two scenarios. On `main` both report `XFAIL` (exit 0). On this branch both report `XPASS(strict)`, which pytest surfaces as a failure precisely because the bug they pin is gone, and with `--runxfail` both pass outright (2 passed). So @technicalpickles' deliberately fix-free repro is an independent confirmation of this branch, and I would suggest folding those two tests in here with the markers dropped, since they cover the mid-run path at the client-and-transport level rather than at the handler level.

Third, no regressions in the existing suite: 1768 passed on `main` versus 1773 passed on this branch, same 90 skipped, same 3 xfailed, `--ignore=tests/integration/cross_version` on both because that directory needs network I do not have. So the five new tests are the entire delta.

Now the thing I think needs changing. The reordering leaves a path where the producer task is never cancelled at all. The new shape cancels the producer in the `except Exception` arm and again after the `try`, but `asyncio.CancelledError` is a `BaseException`, so if `agent_executor.cancel()` raises it, neither statement runs and the producer survives as a pending `asyncio.Task`. On `main` this could not happen, because the producer was cancelled unconditionally before the executor was awaited, so this is a regression rather than a pre-existing gap. I built a probe with a hanging `execute()` and a `cancel()` that raises `asyncio.CancelledError`, and asserted on `active._producer_task.done()` after the call: on `main` it is `True`, on this branch it is `False`. That is the same failure class @mykytanetipa flagged on my #1105, where a dispatcher cancelled but not awaited surfaced as `Task was destroyed but it is pending!`. A `try/finally` closes it and removes the duplicated call:

```python
try:
    await self._agent_executor.cancel(
        request_context, self._event_queue_agent
    )
except Exception as e:
    logger.exception('Cancel[%s]: Agent cancel failed', self._task_id)
    await self._mark_task_as_failed(e)
    raise
finally:
    self._producer_task.cancel()
```

The second point is about where the owner check belongs, and I raise it as a question because I may be missing a reason for the current placement. The registry's miss path is already owner-scoped: `get_or_create` on a miss calls `ActiveTask.start()`, which reads through `TaskManager.get_task()` with the caller's `call_context` and raises `TaskNotFoundError` when `create_task_if_missing` is false. I verified that by evicting a task from the registry while leaving it in the store and then having a non-owner cancel it on unmodified `main`: rejected with `TaskNotFoundError`. So the gap is exactly the cache-hit early return at `active_task_registry.py:52-53`, which returns before `call_context` is ever consulted. Guarding it there, conditioned on `not create_task_if_missing` so `on_message_send` is unaffected, would make the hit path symmetric with the miss path and would close the class for every present and future caller of the registry, rather than for the two handlers that happen to be known today. Two call-site checks fix the two known holes and leave `ActiveTaskRegistry` itself as an unauthenticated lookup. If there is a reason the check has to live in the handler, I would find that useful to know.

The third point is scope rather than correctness. The terminal state now reaches the store but never reaches the wire. I subscribed to a live task, cancelled it, and collected the events the subscriber actually received: the stream yields the initial `Task` in `WORKING` and then ends, while `task_store.get` afterwards returns `CANCELED`. A client that was streaming therefore still cannot tell that the task was cancelled without polling, and `push_sender` does not fire either, since `_update_task_state` only notifies on `PushNotificationEvent`. I do not think that blocks this PR, because at that point both queues are already closed and a direct store write is the only option left, which is also why the producer-failure path at `active_task.py:570-572` does exactly the same thing, under a comment that says so. But it is worth a sentence in the docstring or a follow-up issue, because "cancel now writes a terminal state" reads as a stronger promise than what a streaming client receives. Relatedly, mutating `task.status.state` in place mutates the shared `_task_manager._current_task`, which is the same object already handed to a subscriber, so an already-yielded reference changes underneath the reader. I noticed that only because it confounded my first measurement, and it may well be fine given the same pattern exists elsewhere in the file.

On splitting: I would keep these together. The repo has split bundled PRs before, including one of mine, but here your third test is the argument against it, because it shows the two changes interact. Landing #1170's unconditional terminal write without #1159's owner check would turn a read-only cross-tenant subscribe into a cross-tenant write, which is strictly worse than either bug alone. If a maintainer does want them separated, the safe order is #1159 first and #1170 second, never the reverse.

Reproduction details if any of the above is worth re-checking: two `git worktree` checkouts at `cff6727` and `a382ad5`, `uv sync --all-extras --group dev`, Python 3.10, `pytest -p no:randomly`.

chopmob-cloud:
Thanks, this is a good catch and you are right on the first point. It is a
regression we introduced and it should not ship as-is.

**The producer leak (point 1): confirmed, adopting your `try/finally`.**

The reorder moved the producer cancel to after the executor await, so the only
statements that cancel it are the `except Exception` arm and the trailing line.
`asyncio.CancelledError` has been a `BaseException` (not an `Exception`) since
3.8, so when `agent_executor.cancel()` raises it (a cancelled cancel, a
shutdown, or an executor that raises it directly, as in your probe) neither runs
and the producer is left pending. On `main` the unconditional pre-await
`self._producer_task.cancel()` made this impossible, so this is a regression, not
a pre-existing gap. I confirmed the control flow: with the branch shape the
producer cancel fires on neither path once `cancel()` raises `CancelledError`;
your `finally` restores it.

Taking your shape as written:

```python
try:
    await self._agent_executor.cancel(
        request_context, self._event_queue_agent
    )
except Exception as e:
    logger.exception('Cancel[%s]: Agent cancel failed', self._task_id)
    await self._mark_task_as_failed(e)
    raise
finally:
    self._producer_task.cancel()
```

One extra reason it is the right shape, beyond closing the `BaseException` hole:
it also fixes an ordering inconsistency the current `except` arm has with this
PR's own stated intent. Today that arm runs `self._producer_task.cancel()` and
then `await self._mark_task_as_failed(e)`, but `_mark_task_as_failed` enqueues
the `FAILED` `TaskStatusUpdateEvent` onto `_event_queue_agent` and swallows
`QueueShutDown`. Cancelling the producer first, then awaiting, lets the producer
tear the queue down at that await point and the FAILED write gets silently
dropped, which is the same "cancel producer first drops the terminal write"
hazard the reorder was meant to avoid. The `finally` writes FAILED to the
still-open queue first, then cancels, so both the `Exception` and
`CancelledError` paths are correct. I will add a regression test on your probe
(hanging `execute`, `cancel` raising `CancelledError`, assert the producer is
cancelled).

**Owner check placement (point 2).**

You are right that the miss path is already owner-scoped through
`ActiveTask.start()` -> `TaskManager.get_task(call_context)`, and that the only
hole is the cache-hit early return. We guarded at the two call sites to mirror
the existing handler-level checks (`on_get_task`, `_setup_active_task`) and keep
`ActiveTaskRegistry` auth-unaware, but I agree the registry-level guard, gated on
`not create_task_if_missing` as you note so `on_message_send` is unaffected, is
the stronger invariant and would let the two call-site checks go away. I would
rather a maintainer weigh in on whether auth belongs in the registry before I
move it there, since it shifts which layer owns the check. Happy to do it either
way.

**Wire visibility and shared mutation (point 3).**

Agreed it is out of scope for this PR: at that point both queues are closed, a
direct store write is the only option left, and it mirrors the producer-failure
path at `active_task.py:570-572` for exactly that reason. I will tighten the
docstring so "cancel writes a terminal state" reads as "reaches the store," not
"reaches an active stream," and open a follow-up for wire-level cancel
visibility (a synthetic terminal event to live subscribers plus the
`push_sender` path). On the in-place `task.status.state` mutation of the shared
`_current_task`: good catch, I will stop mutating the yielded object in place and
write a copy so an already-yielded reference does not change under a reader.

On splitting: agreed, keep them together, and thanks for spelling out why (the
#1170 write on the un-guarded path turns a cross-tenant read into a cross-tenant
write). If they are ever split, #1159 first.

I have folded in the `try/finally` plus a regression test that reproduces your
probe (a hanging `execute`, a `cancel` that raises `asyncio.CancelledError`,
asserting the producer receives the cancellation rather than leaking pending). It
fails on the current branch and passes with the fix, and the existing cancel
tests, the owner-scope handler tests, and the #1170 scenario tests stay green.

chopmob-cloud:
Opened #1175 to track the stream and push visibility follow-up from this review: the terminal state reaches the store but not an active subscriber stream. It covers both the cancel path and the parallel producer-failure FAILED write.

chopmob-cloud:
Two updates on this.

**Fold-in done, with attribution.** The two mid-run scenarios are
@technicalpickles' from #1171, not mine. I folded them in here verbatim minus
the strict-xfail markers, and I have now credited the origin properly: a
`Co-authored-by` trailer and a comment in `test_scenarios.py` pointing back at
#1171 (0959898). @technicalpickles you filed the repro separately on purpose, so
this is your call: if you would rather land #1171 on its own I will drop the two
scenarios from here and rebase to strip the markers once yours merges, so main
never goes red on the ordering. If you are happy for them to ride along they stay
co-authored to you. Either way the fix and the repro should land together, since
keeping #1170's terminal write and #1159's owner check in one change is what
stops the cross-tenant subscribe becoming a cross-tenant write.

**On where the owner check belongs (your second point): adopted.** You were right
that guarding the two handlers leaves `ActiveTaskRegistry` itself an
unauthenticated lookup, one forgotten call site away from regressing. I moved the
guard into `get_or_create` on the cache-hit early return, conditioned on `not
create_task_if_missing` so the `on_message_send` create path is untouched,
mirroring the miss path's own owner check in `ActiveTask.start`. It is masked as
`TaskNotFoundError` so existence is not leaked, and it runs outside the registry
lock since the store read is I/O and the miss-path check is outside the lock too.
Added a registry-level regression test that a non-owner cannot resolve another
user's cached task. The two handler-site guards stay as defence in depth.
(1bcd46e)

Thanks for the verbatim cross-check and for pushing on the placement, both made
this materially better.

astrogilda:
I re-checked all three findings from my review against 2ef51c9 and all three are closed. Here is what I checked them against, so the agreement rests on the diff and not on my nodding at your summary.

The producer cancel in the finally closes the leak I reported, and the second reason you gave for the ordering is the stronger one. One thing worth knowing when you write the regression test: Task.cancel() requests cancellation and does not complete it, so on the raising path there is no await between the finally firing and the exception leaving cancel(), and at that instant the producer is cancel-requested while done() is still False. On the ordinary path the later `await self._is_finished.wait()` and the store write give the loop its turn and it settles. If the test asserts done() immediately after a pytest.raises around cancel(), give it a loop turn or await the task. Otherwise it can read red on a correct fix, or green for a reason that has nothing to do with the fix, depending on where the nearest await lands.

I tried to break the copy before believing it. The usual way a call-site copy turns out decorative is that the layer underneath merges into the shared object, which moves the in-place write one frame down. Here the layer rebinds. `TaskManager._save_task` does `self._current_task = task`, so after your save_task_event(updated) the manager points at the copy and the object a subscriber already holds is left alone. The guarantee holds for that write.

The comment above it reads as a general rule the file does not keep, and that is the one thing I would change. subscribe(include_initial_task=True) yields whatever get_task() returns, which is `_task_manager._current_task`, and every ordinary status write reaches that same object in place. save_task_event on a TaskStatusUpdateEvent goes through ensure_task into ensure_task_id, which returns `_current_task`, then does task.status.CopyFrom(event.status). The producer-failure arm in `_run_producer` writes task.status.state = TaskState.TASK_STATE_FAILED on the same object before its own save_task_event(task). An already-yielded reference does still change under a reader on every status write except the one you copied. Keep the copy. I would have the comment say it protects this write, because a reader who takes "that reference must not change under the reader" as a property of the file will be wrong about the file, and the general version of that property is #1175 and #1191 territory.

One residual door, small enough that I am naming it for the record and not asking for anything. If executor.cancel raises a BaseException such as CancelledError, the finally fires and the exception leaves cancel() before `await self._is_finished.wait()`, so neither the FAILED write in the except arm nor the CANCELED write after the lock runs, and the task is left non-terminal with no producer behind it. That does not put #1170's shape back, because the caller gets an exception, and the shape #1170 describes is a silent success. A second cancel() takes the else branch, since the consumer's own finally has set `_is_finished` by then, and it does reach the terminal write, so the state is recoverable. Worth a sentence somewhere so the scope is written down where a reader can find it.

On the two folded-in scenarios, Josh answered over on #1171 that he considers it superseded and is happy to close in favour of this one. I checked the fold against your diff before agreeing, and the summary played no part in it. Both are in tests/integration/test_scenarios.py as test_scenario_19_mid_run_cancel_reaches_a_terminal_state and test_scenario_19_cancel_of_parked_task_does_not_silently_succeed, with the strict-xfail markers dropped and a comment recording why. So there is nothing for you to drop or rebase out on that question. Fifteen checks green at 2ef51c9.

Last thing, and it is bookkeeping. My review was COMMENTED because I am not a maintainer here and an approval from me carries no merge signal. Every point in it is now closed in code, so I am saying that plainly instead of leaving the only review on a green branch looking like an unresolved objection.

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -165,6 +165,15 @@ async def on_cancel_task(  # noqa: D102
...
         task_id = params.id
 
+        # Owner-aware isolation: mirror on_get_task / _setup_active_task and
+        # consult the owner-scoped store before resolving the live task from
+        # the registry, which keys on task_id alone. Without this a non-owner
+        # who knows the uuid4 task_id could cancel another user's live task
+        # (issue #1159, CWE-639). Mask a partition mismatch as not-found so
+        # existence is not leaked.
+        if not await self.task_store.get(task_id, context):
+            raise TaskNotFoundError
```
mykytanetipa: With the owner-aware guard inside get_or_create this handler-level task_store.get(id, context) check is now redundant

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -395,6 +404,13 @@ async def on_subscribe_to_task(  # noqa: D102
     ) -> AsyncGenerator[Event, None]:
         task_id = params.id
 
+        # Owner-aware isolation before resolving the live task from the
+        # registry (which is keyed on task_id alone); see on_cancel_task.
+        # Prevents a non-owner from subscribing to another user's live
+        # task stream (issue #1159, CWE-639).
+        if not await self.task_store.get(task_id, context):
+            raise TaskNotFoundError
```
mykytanetipa: same redundancy issue as in on_cancel_task

mykytanetipa (changes requested):
Thank you for tackling this one, correct fix for both issues. Agree with the decision to handle #1175 separately.

Only point before the approve can be granted, please resolve the owner guard duplication comments below.

src/a2a/server/request_handlers/default_request_handler_v2.py
chopmob-cloud: Done in 44b2af6. Removed this handler guard. The owner check now lives solely in `ActiveTaskRegistry.get_or_create`, which runs `task_store.get(task_id, call_context)` on the cache-hit early return (gated on `not create_task_if_missing`) and on the miss path via `ActiveTask.start`. `on_cancel_task` calls `get_or_create` with `create_task_if_missing=False`, so the cross-tenant cancel path stays closed (#1159).

Verified on the running code: the three #1159 owner-scope tests still pass with this guard gone, and neutralising the registry guard makes them fail, so it is the single load-bearing enforcement now.

src/a2a/server/request_handlers/default_request_handler_v2.py
chopmob-cloud: Also removed in 44b2af6, same reasoning as `on_cancel_task`: `on_subscribe_to_task` calls `get_or_create` with `create_task_if_missing=False`, so the registry guard covers the cache-hit and miss paths for it too. The owner-scope subscribe test still passes with this handler guard gone.

mykytanetipa (approved):

chopmob-cloud:
Thanks for re-deriving all three against the diff rather than the summary; that is the review I would want on my own code.

Both of your notes are now reflected on the branch. The copy comment no longer states a file-wide rule: it says the copy keeps this terminal write off the yielded reference, and that ordinary status writes (save_task_event to ensure_task to CopyFrom, and the producer-failure write) still update the shared object in place, with the general copy-on-yield property left to #1175 / #1191.

The BaseException scope you named is written down at the cancel site now: if executor.cancel() raises one, it propagates from the finally before the _is_finished.wait() and the CANCELED write, leaving the task non-terminal with no producer. The caller receives the exception, so it is not the #1170 silent-success shape, and it is recoverable, since a later cancel() takes the else branch and reaches the terminal write.

On the regression test, thank you for the timing note. Task.cancel() requesting rather than completing is exactly the trap; the test should give the assertion a loop turn (or await the task) after the pytest.raises so it is not green or red for the wrong reason.

And thanks for confirming the fold with Josh on #1171. That settles the last open question on the scenarios.

PR #1173, opened 2026-08-09
fix(server): validate push-notification URLs at config creation

SashaMIT:
## Summary

The `on_create_task_push_notification_config` handlers (v1 and v2) stored client-supplied URLs without any validation. A caller who can create a push notification config can point the server at loopback, private-network, link-local, or cloud-metadata hosts, and the server will POST task events to that URL on every state change.

## Root cause

`src/a2a/server/request_handlers/default_request_handler.py` and `default_request_handler_v2.py` call `push_config_store.set_info(task_id, params, context)` without checking `params.url`. The dispatch path (`BasePushNotificationSender._dispatch_notification`) is covered by #1164; this PR closes the **write path**.

## Fix

- Added `push_url_validation_error` to `base_push_notification_sender.py` (same logic as #1164: blocks non-http(s) schemes, loopback, private, link-local, multicast, reserved, and unresolvable hosts).
- Called it in both `on_create_task_push_notification_config` handlers before storing the config.
- Raises `InvalidParamsError` with a descriptive message on rejection.

## Testing

- `uv run pytest tests/server/request_handlers/ -k push_notification` — 55 passed.
- New test: `test_on_create_task_push_notification_config_rejects_invalid_url` covers loopback and `file:` scheme rejection.
- Updated existing tests that used unresolvable fixture URLs (`1.example.com`, `callback.com`) to use `example.com` (resolvable in CI).

## Relation to #1164

#1164 validates at dispatch time (read path). This PR validates at config creation time (write path). Both are needed: write-time validation fails fast and gives the client immediate feedback; dispatch-time validation is a defense-in-depth backstop.

Made with [Cursor](https://cursor.com)


src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -102,6 +105,7 @@ def __init__(  # noqa: PLR0913
             [AgentCard, ServerCallContext], Awaitable[AgentCard]
         ]
         | None = None,
+        allow_private_push_urls: bool = False,
```
sokoliva: Can we rather do something like 

`push_url_validator: Callable[...] | None = push_url_validation_error` ?

A callable lets developers supply their own policy. [The spec](https://a2a-protocol.org/latest/specification/#132-push-notification-security) lists these as checks that SHOULD be done, not MUST and says "implement URL allowlists where appropriate." So the right policy is deployment specific, and the A2A library shouldn't hard-code one.

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -91,12 +94,16 @@ def __init__(  # noqa: PLR0913
             [AgentCard, ServerCallContext], Awaitable[AgentCard]
         ]
         | None = None,
+        allow_private_push_urls: bool = False,
```
sokoliva: Same as above, something like:

`push_url_validator: Callable[...] | None = push_url_validation_error` ?

src/a2a/server/request_handlers/default_request_handler_v2.py
sokoliva, line 257: We should also add a check before this call.

src/a2a/server/request_handlers/default_request_handler.py
sokoliva, line 329: Also check before this call.

src/a2a/server/tasks/base_push_notification_sender.py
```diff
@@ -20,6 +23,51 @@
...
+        addr.is_private
+        or addr.is_loopback
+        or addr.is_link_local
+        or addr.is_multicast
+        or addr.is_reserved
+        or addr.is_unspecified
+    )
+
+
+def push_url_validation_error(url: str) -> str | None:
```
sokoliva: Can we make this async and use the loop's resolver?

Something like:
```
 async def push_url_validation_error(url: str) -> str | None:
   ...
   loop = asyncio.get_running_loop()
   infos = await loop.getaddrinfo(host, port, type=socket.SOCK_STREAM)
```
socket.getaddrinfo is blocking, and both call sites are async def handlers.

sokoliva (changes requested):

SashaMIT:
Thanks. The library default is now an injectable async `push_url_validator` (default `push_url_validation_error`, `None` skips). Validation runs before both `set_info` sites, including the send-message path. Resolver uses the running loop's `getaddrinfo`.

src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -102,6 +105,8 @@ def __init__(  # noqa: PLR0913
             [AgentCard, ServerCallContext], Awaitable[AgentCard]
         ]
         | None = None,
+        push_url_validator: Callable[[str], Awaitable[str | None]]
+        | None = push_url_validation_error,
```
sokoliva: Let's actually default to None here to avoid making a breaking change. Also this would make more sense considering [the spec states should](https://a2a-protocol.org/latest/specification/#132-push-notification-security).


This also fixes the ITK error: 
`ERROR - Test PR - star - push notification - jsonrpc failed with exception: JSON-RPC Error: {'code': -32602, 'message': "Invalid push notification URL: host '127.0.0.1' resolves to a non-public address", 'data': [{'@type': 'type.googleapis.com/google.rpc.ErrorInfo', 'reason': 'INVALID_PARAMS', 'domain': 'a2a-protocol.org', 'metadata': {}}]}`

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -91,12 +94,15 @@ def __init__(  # noqa: PLR0913
             [AgentCard, ServerCallContext], Awaitable[AgentCard]
         ]
         | None = None,
+        push_url_validator: Callable[[str], Awaitable[str | None]]
+        | None = push_url_validation_error,
```
sokoliva: Same as above: default to None.

sokoliva (changes requested):
Thank you for making the changes. I left 2 equivalent comments. Sorry for the inconvenience and thank you for the patience!

src/a2a/server/request_handlers/default_request_handler.py
SashaMIT: Done. Default is now None so this is not a breaking change and the ITK loopback webhook keeps working. Pass push_url_validation_error to opt in.

src/a2a/server/request_handlers/default_request_handler_v2.py
SashaMIT: Same change here. Default is None.

SashaMIT:
Thanks, no inconvenience. Default is now None on both handlers so existing ITK loopback webhooks keep working. Pass `push_url_validation_error` to opt in.

sokoliva (approved):

SashaMIT:
Thanks Iva, glad this landed. The None default keeps existing ITK loopback webhooks working.

PR #1176, opened 2026-08-10
fix: validate task ID length and message content

ez-lbz:
### 1. Reject over-long task IDs before persistence

**Problem:** No upper bound is enforced on `taskId`. A client can send a 1000+ character task ID in `tasks/get`, `tasks/cancel`, `message/send`, `SubscribeToTask` or any push-notification-config method and the ID is passed straight to the task store and persisted. There is also no check that a provided task ID is non-empty. Affected files: `src/a2a/server/request_handlers/default_request_handler.py`, `src/a2a/server/request_handlers/default_request_handler_v2.py`.

**Fix (src/a2a/utils/task.py, src/a2a/server/request_handlers/default_request_handler.py, src/a2a/server/request_handlers/default_request_handler_v2.py):**
- Added `MAX_TASK_ID_LENGTH = 1000` and `validate_task_id()` in `src/a2a/utils/task.py`; it raises `InvalidParamsError` for empty task IDs or IDs longer than 1000 characters.
- Applied `validate_task_id()` at the entry of every task-ID-taking handler method in both V1 (`LegacyRequestHandler`) and V2 (`DefaultRequestHandlerV2`): `on_get_task`, `on_cancel_task`, `on_subscribe_to_task`, `on_message_send`/`on_message_send_stream` (via `_setup_message_execution`/`_setup_active_task`, only when a task ID is supplied) and the four push-notification-config methods. Validation happens before any store read/write.

### 2. Reject messages with empty content

**Problem:** `MessageSend` accepts messages with no parts or with parts that carry no content (no `text`, `raw`, `url` or non-null `data`). Such empty payloads are persisted into task history. The framework's schema validation already rejects an empty `parts` list with `InvalidParamsError` ("Validation failed"), but a part with no content (e.g. only `media_type` set) passes through unchecked.

**Fix (src/a2a/utils/task.py, src/a2a/server/request_handlers/default_request_handler.py, src/a2a/server/request_handlers/default_request_handler_v2.py):**
- Added `validate_message_content()` in `src/a2a/utils/task.py`, which raises `InvalidParamsError` if the message has no parts or any part is empty (no `text`/`raw`/`url`, and `data` is unset or null).
- Called it in `_setup_message_execution` (V1) and `_setup_active_task` (V2), which both `on_message_send` and `on_message_send_stream` route through, so the check applies to the streaming path too.

## Testing

- `./.venv/Scripts/python -m pytest tests/utils/test_task.py tests/server/request_handlers/test_default_request_handler.py tests/server/request_handlers/test_default_request_handler_v2.py -q` → **159 passed** (includes 13 new regression tests: task-ID length limits and empty-content rejection at both the utility and handler level).
- `./.venv/Scripts/python -m ruff check` on all modified files: clean (the two `too-many-positional-arguments` findings on the handler `__init__` lines pre-date this change and exist on `main`).
- Behavior change: requests with over-long/empty task IDs and messages with empty content now return `InvalidParamsError` instead of being processed. This is the intended hardening; clients must send valid task IDs and non-empty message content.

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler.py
+++ b/src/a2a/server/request_handlers/default_request_handler.py
@@ -61,7 +61,9 @@ from a2a.utils.errors import (
 from a2a.utils.task import (
     apply_history_length,
     validate_history_length,
+    validate_message_content,
     validate_page_size,
+    validate_task_id,
 )
 from a2a.utils.telemetry import SpanKind, trace_class
 
@@ -146,6 +148,7 @@ class LegacyRequestHandler(RequestHandler):
     ) -> Task | None:
         """Default handler for 'tasks/get'."""
         validate_history_length(params)
+        validate_task_id(params.id)
 
         task_id = params.id
         task: Task | None = await self.task_store.get(task_id, context)
@@ -187,6 +190,7 @@ class LegacyRequestHandler(RequestHandler):
         Attempts to cancel the task managed by the `AgentExecutor`.
         """
         task_id = params.id
+        validate_task_id(task_id)
         task: Task | None = await self.task_store.get(task_id, context)
         if not task:
             raise TaskNotFoundError
@@ -265,6 +269,9 @@ class LegacyRequestHandler(RequestHandler):
         # Create task manager and validate existing task
         # Proto empty strings should be treated as None
         task_id = params.message.task_id or None
+        if task_id:
+            validate_task_id(task_id)
+        validate_message_content(params.message)
         context_id = params.message.context_id or None
         task_manager = TaskManager(
             task_id=task_id,
@@ -523,6 +530,7 @@ class LegacyRequestHandler(RequestHandler):
             raise PushNotificationNotSupportedError
 
         task_id = params.task_id
+        validate_task_id(task_id)
         task: Task | None = await self.task_store.get(task_id, context)
         if not task:
             raise TaskNotFoundError
@@ -555,6 +563,7 @@ class LegacyRequestHandler(RequestHandler):
 
         task_id = params.task_id
         config_id = params.id
+        validate_task_id(task_id)
         task: Task | None = await self.task_store.get(task_id, context)
         if not task:
             raise TaskNotFoundError
@@ -585,6 +594,7 @@ class LegacyRequestHandler(RequestHandler):
         Requires the task and its queue to still be active.
         """
         task_id = params.id
+        validate_task_id(task_id)
         task: Task | None = await self.task_store.get(task_id, context)
         if not task:
             raise TaskNotFoundError
@@ -635,6 +645,7 @@ class LegacyRequestHandler(RequestHandler):
             raise PushNotificationNotSupportedError
 
         task_id = params.task_id
+        validate_task_id(task_id)
         task: Task | None = await self.task_store.get(task_id, context)
         if not task:
             raise TaskNotFoundError
@@ -667,6 +678,7 @@ class LegacyRequestHandler(RequestHandler):
 
         task_id = params.task_id
         config_id = params.id
+        validate_task_id(task_id)
         task: Task | None = await self.task_store.get(task_id, context)
         if not task:
             raise TaskNotFoundError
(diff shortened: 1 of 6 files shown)


src/a2a/server/request_handlers/default_request_handler.py
```diff
@@ -187,6 +190,7 @@ async def on_cancel_task(
         Attempts to cancel the task managed by the `AgentExecutor`.
         """
         task_id = params.id
+        validate_task_id(task_id)
```
mykytanetipa: Why do we need validate_task_id in on_cancel_task?

task_id is looked up, not stored so the "prevent persisting huge IDs" rationale doesn't apply here

src/a2a/server/request_handlers/default_request_handler.py
mykytanetipa, line 1: please avoid changes in 0.3.x structures, LegacyRequestHandler is deprecated and wont receive new feature updates

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -128,6 +130,7 @@ async def on_get_task(  # noqa: D102
         context: ServerCallContext,
     ) -> Task | None:
         validate_history_length(params)
+        validate_task_id(params.id)
```
mykytanetipa, line 133: Why do we need validate_task_id in on_get_task?

task_id is looked up, not stored so the "prevent persisting huge IDs" rationale doesn't apply here

src/a2a/server/request_handlers/default_request_handler_v2.py
```diff
@@ -164,6 +167,7 @@ async def on_cancel_task(  # noqa: D102
         context: ServerCallContext,
     ) -> Task | None:
         task_id = params.id
+        validate_task_id(task_id)
```
mykytanetipa, line 170: Why do we need validate_task_id in on_cancel_task?

task_id is looked up, not stored so the "prevent persisting huge IDs" rationale doesn't apply here

src/a2a/server/request_handlers/default_request_handler_v2.py
mykytanetipa, line 133: instead of stacking this validation in the function body I would move this into decorator that validates request params

src/a2a/utils/task.py
```diff
@@ -5,11 +5,65 @@
 from base64 import b64decode, b64encode
 from typing import Literal, Protocol, runtime_checkable
 
-from a2a.types.a2a_pb2 import Task
+from a2a.types.a2a_pb2 import Message, Part, Task
 from a2a.utils.constants import MAX_LIST_TASKS_PAGE_SIZE
 from a2a.utils.errors import InvalidParamsError
 
 
+MAX_TASK_ID_LENGTH = 1000
```
mykytanetipa, line 13: I'd avoid setting MAX_TASK_ID_LENGTH just as a global constant. If this change lands it should be configurable, opt-in, with None default and set by some server config

src/a2a/utils/task.py
```diff
@@ -5,11 +5,65 @@
...
+    if not message.parts:
+        raise InvalidParamsError(
+            message='message must contain at least one part'
+        )
+    for part in message.parts:
+        if not _part_has_content(part):
+            raise InvalidParamsError(message='message parts must not be empty')
+
+
+def _part_has_content(part: Part) -> bool:
```
mykytanetipa, line 51: isn't spec-aligned

src/a2a/utils/task.py
```diff
@@ -5,11 +5,65 @@
...
+    """Validates that a message carries actual content.
+
+    A message must contain at least one part, and every part must have
+    content (text, raw bytes, a URL or data) rather than being empty.
+
+    Raises:
+        InvalidParamsError: If the message has no parts or contains an
+            empty part.
+    """
+    if not message.parts:
```
mykytanetipa, line 42: this seems redundant, Message.parts is REQUIRED in the spec proto and validate_request_params will reject an empty field

mykytanetipa (changes requested):
The task_id cap is a reasonable idea, but it's a public-facing behavioral change, so it should be reviewed and accepted by the A2A spec maintainers first. Please open a spec proposal PR at https://github.com/a2aproject/A2A before implementing it in the SDK.

PR #1177, opened 2026-08-10
fix: enforce upper bound on history_length

ez-lbz:
### 1. Enforce an upper bound on `historyLength`

**Problem:** `validate_history_length()` in `src/a2a/utils/task.py` only rejected negative values. A client could send `historyLength=999999999` and the server would materialize the full task history for every matching task (in `tasks/get`, `tasks/list`, and `message/send`), allowing an unbounded-history read via an absurdly large parameter.

**Fix (src/a2a/utils/task.py):**
- Added `MAX_HISTORY_LENGTH = 1000` with a comment explaining the cap is a pragmatic bound aligned with the other A2A SDKs' "large value ≈ return everything available" semantics.
- `validate_history_length()` now raises `InvalidParamsError` when `history_length > MAX_HISTORY_LENGTH` (in addition to the existing negative-value check). The validation already runs at the entry of `on_get_task`, `on_list_tasks`, and `on_message_send` (V1) and their V2 counterparts, so the cap applies to all JSON-RPC / REST / gRPC paths without further changes.

## Testing

- `./.venv/Scripts/python -m pytest tests/utils/test_task.py tests/server/request_handlers/test_default_request_handler.py tests/server/request_handlers/test_default_request_handler_v2.py -q` → **150 passed** (includes 8 new regression tests: `validate_history_length` unit tests plus handler-level `history_length` over-limit rejection for `on_get_task` and `on_message_send`).
- Behavior change: requests with `historyLength` above 1000 now return `InvalidParamsError` instead of returning full history. Requests at or below the cap are unaffected (existing tests all use values ≤ 10).

### DIFF
--- a/src/a2a/utils/task.py
+++ b/src/a2a/utils/task.py
@@ -10,6 +10,17 @@ from a2a.utils.constants import MAX_LIST_TASKS_PAGE_SIZE
 from a2a.utils.errors import InvalidParamsError
 
 
+MAX_HISTORY_LENGTH = 1000
+"""Maximum allowed value for a ``history_length`` request.
+
+Keeps a single request from asking the server to materialize an
+unbounded history (e.g. ``historyLength=999999999``). The value is a
+pragmatic cap aligned with the other A2A SDKs' semantics, where very
+large values effectively mean "return everything available"; clients
+requesting more than this cap get ``InvalidParamsError``.
+"""
+
+
 @runtime_checkable
 class HistoryLengthConfig(Protocol):
     """Protocol for configuration arguments containing history_length field."""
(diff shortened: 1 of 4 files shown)


src/a2a/utils/task.py
```diff
@@ -25,9 +36,13 @@ def HasField(self, field_name: Literal['history_length']) -> bool:  # noqa: N802
 
 
 def validate_history_length(config: HistoryLengthConfig | None) -> None:
```
mykytanetipa, line 38: The PR description states that right now server "allowing an unbounded-history read via an absurdly large parameter", but the change into validate_history_length doesn't affect this, full task history will still be loaded in memory.

example on_get_task:
```
# param validation happens here:
validate_history_length(params)                                                    
...
# task is loaded with full history here:
task: Task | None = await self.task_store.get(task_id, context)  
...
# history cap is applied here:
return apply_history_length(task, params)                                   
```

src/a2a/utils/task.py
```diff
@@ -10,6 +10,17 @@
 from a2a.utils.errors import InvalidParamsError
 
 
+MAX_HISTORY_LENGTH = 1000
```
mykytanetipa, line 13: I'd avoid setting MAX_HISTORY_LENGTH just as a global constant. If this change lands it should be configurable, opt-in, with None default and set by some server config

mykytanetipa (changes requested):
The history_length cap is a reasonable idea, but it's a public-facing behavioral change, so it should be reviewed and accepted by the A2A spec maintainers first. Please open a spec proposal PR at https://github.com/a2aproject/A2A before implementing it in the SDK.

PR #1178, opened 2026-08-10
fix: reject terminal tasks before persisting push config (V2)

ez-lbz:
### 1. Check terminal task state before persisting the push config (V2 handler)

**Problem:** In `DefaultRequestHandlerV2._setup_active_task()` (`src/a2a/server/request_handlers/default_request_handler_v2.py`), the push-notification config was written to the store via `set_info()` *before* the terminal-state check ran inside `ActiveTask.start()`. As a result, sending a message to an already-completed/canceled/failed/rejected task with a `task_push_notification_config` in the request still persisted the config for that terminal task. V1's handler checks terminal state before any persistence.

**Fix (src/a2a/server/request_handlers/default_request_handler_v2.py):**
- When `message.task_id` is provided and the task exists, the terminal-state check now runs immediately after loading the task and before the request-context build / `set_info()` call, matching V1's ordering (`default_request_handler.py` `_setup_message_execution`).
- Raises the same `InvalidParamsError("Task {id} is in terminal state: {state}")` that `ActiveTask.start()` raised before, so the error contract is unchanged; the `start()` check remains as a second line of defense.

## Testing

- `./.venv/Scripts/python -m pytest tests/server/request_handlers/test_default_request_handler_v2.py -q` → **63 passed**.
- `./.venv/Scripts/python -m pytest tests/server/request_handlers/test_default_request_handler.py tests/utils/test_task.py -q` → **82 passed** (V1 unaffected).
- Updated the two existing `on_message_send` / `on_message_send_stream` terminal-state tests to seed the terminal task through the task store (the check now runs earlier than the patched `TaskManager.get_task`), and added a new parametrized regression test `test_on_message_send_terminal_task_skips_push_config_persistence` asserting `set_info` is never awaited for terminal tasks.
- `./.venv/Scripts/python -m ruff check` on modified files: clean (the pre-existing `too-many-positional-arguments` finding on the handler `__init__` exists on `main`).
- Behavior change: push configs are no longer persisted for terminal tasks; the error raised to the client is unchanged.

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler_v2.py
+++ b/src/a2a/server/request_handlers/default_request_handler_v2.py
@@ -203,6 +203,17 @@ class DefaultRequestHandlerV2(RequestHandler):
             task = await self.task_store.get(original_task_id, call_context)
             if not task:
                 raise TaskNotFoundError(f'Task {original_task_id} not found')
+            # Reject terminal tasks before persisting anything (e.g. the push
+            # notification config below). Previously the terminal-state check
+            # only happened inside ActiveTask.start(), after set_info() had
+            # already written the config for an already-completed task.
+            if task.status.state in TERMINAL_TASK_STATES:
+                raise InvalidParamsError(
+                    message=(
+                        f'Task {task.id} is in terminal state: '
+                        f'{task.status.state}'
+                    )
+                )
 
         # Build context to resolve or generate missing IDs
         request_context = await self._request_context_builder.build(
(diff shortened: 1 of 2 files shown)


ConnorMoss02:
Ran the two existing terminal-state tests against main without your change and they still pass. They only prove an error comes back, not that it comes back before the config gets written, so `set_info.assert_not_awaited()` in the new test is the only assertion that fails if this fix is removed.

ez-lbz:
Exactly right — and thanks for running that check. The `set_info.assert_not_awaited()` assertion is the whole point of the new test: the two pre-existing tests pass on `main` precisely because the terminal-state rejection happens eventually, just not before the config is persisted — so they can't distinguish the buggy ordering from the fixed one. `assert_not_awaited` pins the ordering itself, which is the only observable difference that matters here.

PR #1179, opened 2026-08-10
fix: remove zombie registry entries when ActiveTask start fails

ez-lbz:
### 1. Remove registry entries synchronously when `ActiveTask.start()` fails

**Problem:** `ActiveTaskRegistry.get_or_create()` (`src/a2a/server/agent_execution/active_task_registry.py`) inserts the `ActiveTask` into `_active_tasks` *before* calling `start()`. If `start()` raises (e.g. the task is already in a terminal state), the only cleanup was the fire-and-forget `asyncio.create_task(self._remove_task(...))` scheduled via `_on_active_task_cleanup` inside `ActiveTask.start()`. That cleanup runs asynchronously and may not execute before another request checks the registry, leaving a zombie entry for a task that never started — and a concurrent `get_or_create` could return the half-started task.

**Fix (src/a2a/server/agent_execution/active_task_registry.py):**
- Wrapped the `start()` call in `get_or_create()` with `try/except`; on any failure the entry is removed synchronously under `_lock` before the exception propagates.
- The removal is idempotent (`pop`), so the fire-and-forget cleanup task scheduled by `ActiveTask.start()` remains harmless if it runs later.

## Testing

- `./.venv/Scripts/python -m pytest tests/server/agent_execution/test_active_task_registry.py -q` → **6 passed** (includes new `test_get_or_create_failed_start_removes_zombie_entry`, which seeds a completed task so `start()` raises `InvalidParamsError`, then asserts the registry no longer contains the task).
- `./.venv/Scripts/python -m pytest tests/server/agent_execution/ -q` → **73 passed**.
- `./.venv/Scripts/python -m pytest tests/server/request_handlers/test_default_request_handler_v2.py -q` → **59 passed** (registry consumers unaffected).
- Behavior change: none for the success path; on a failed `start()` the registry is now guaranteed not to retain the task.

### DIFF
--- a/src/a2a/server/agent_execution/active_task_registry.py
+++ b/src/a2a/server/agent_execution/active_task_registry.py
@@ -90,10 +90,26 @@ class ActiveTaskRegistry:
                 raise TaskNotFoundError
             return existing
 
-        await active_task.start(
-            call_context=call_context,
-            create_task_if_missing=create_task_if_missing,
-        )
+        try:
+            await active_task.start(
+                call_context=call_context,
+                create_task_if_missing=create_task_if_missing,
+            )
+        except Exception:
+            # Remove the entry synchronously instead of relying on the
+            # fire-and-forget _on_active_task_cleanup path scheduled by
+            # ActiveTask.start() on failure. That cleanup runs in a separate
+            # task and may not execute before another request checks the
+            # registry, leaving a zombie entry for a task that never started.
+            # The pop itself is non-blocking, so holding the (sync) RLock
+            # briefly inside the async path is safe.
+            with self._lock:
+                self._active_tasks.pop(task_id, None)
+                logger.debug(
+                    'Removed failed active task for %s from registry',
+                    task_id,
+                )
+            raise
         return active_task
 
     def _on_active_task_cleanup(self, active_task: ActiveTask) -> None:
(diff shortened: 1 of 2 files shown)


src/a2a/server/agent_execution/active_task_registry.py
```diff
@@ -69,10 +69,24 @@ async def get_or_create(
...
+                call_context=call_context,
+                create_task_if_missing=create_task_if_missing,
+            )
+        except Exception:
+            # Remove the entry synchronously instead of relying on the
+            # fire-and-forget _on_active_task_cleanup path scheduled by
+            # ActiveTask.start() on failure. That cleanup runs in a separate
+            # task and may not execute before another request checks the
+            # registry, leaving a zombie entry for a task that never started.
+            async with self._lock:
```
mykytanetipa: async lock doesn't make a lot of sense here for single sync pop with no await, you've likely kept it for consistency with codebase which already moved to threading.RLock.

mykytanetipa (changes requested):
Thank you for this contribution. Could you rebase this onto latest main? It's built on an older base (_lock is now threading.RLock, and get_or_create was restructured)

ez-lbz:
Rebased onto latest `main` — the zombie-entry cleanup now lives in the restructured `get_or_create` (sync `_lock` + `start()` outside the lock): on any `start()` failure the just-inserted entry is removed synchronously under the RLock before the error propagates, so a subsequent lookup can never observe a never-started task. Registry tests pass locally (`tests/server/agent_execution/`: 77 passed).

PR #1180, opened 2026-08-10
fix: make get_info_for_dispatch abstract to prevent silent notification drops

ez-lbz:
### 1. Make `get_info_for_dispatch` abstract on `PushNotificationConfigStore`

**Problem:** `PushNotificationConfigStore.get_info_for_dispatch()` (`src/a2a/server/tasks/push_notification_config_store.py`) was not `@abstractmethod`. Its default implementation fell back to `get_info(task_id, ServerCallContext())` with a synthetic empty context, which resolves to the empty-string owner partition and returns no configs — silently dropping every push notification in any deployment with multiple owners. A subclass that forgot to override the method only logged a warning per call, so the failure was easy to miss in production.

**Fix (src/a2a/server/tasks/push_notification_config_store.py):**
- `get_info_for_dispatch()` is now `@abstractmethod`; the silent owner-scoped fallback is removed. Any store implementation that does not provide the cross-owner dispatch read path now fails loudly at instantiation time instead of dropping notifications at runtime.
- Removed the now-unused module-level `logger`.
- Verified both built-in stores already implement the method: `InMemoryPushNotificationConfigStore` (`inmemory_push_notification_config_store.py`) and `DatabasePushNotificationConfigStore` (`database_push_notification_config_store.py`).

## Testing

- `./.venv/Scripts/python -m pytest tests/server/tasks/test_inmemory_push_notifications.py -q` → **20 passed** (includes 2 new tests: an incomplete subclass without `get_info_for_dispatch` cannot be instantiated; both built-in stores expose a callable dispatch read path).
- `./.venv/Scripts/python -m pytest tests/server/tasks/test_database_push_notification_config_store.py -q` → **23 passed, 44 skipped** (skips are DB-DSN-dependent tests, pre-existing).
- Behavior change: third-party subclasses of `PushNotificationConfigStore` that do not implement `get_info_for_dispatch` will now raise `TypeError` at instantiation instead of silently dropping notifications. This is the intended loud-failure hardening.

### DIFF
--- a/src/a2a/server/tasks/push_notification_config_store.py
+++ b/src/a2a/server/tasks/push_notification_config_store.py
@@ -1,12 +1,7 @@
-import logging
-
 from abc import ABC, abstractmethod
 
 from a2a.server.context import ServerCallContext
 from a2a.types.a2a_pb2 import TaskPushNotificationConfig
-
-
-logger = logging.getLogger(__name__)
 
 
 class PushNotificationConfigStore(ABC):
@@ -34,6 +29,7 @@ class PushNotificationConfigStore(ABC):
         context).
         """
 
+    @abstractmethod
     async def get_info_for_dispatch(
         self,
         task_id: str,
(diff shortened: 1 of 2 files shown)


mykytanetipa (commented):
Thank you for this contribution, agree the underlying bug is real and worth fixing.

However, making this method abstract is a breaking change to a public ABC, we're not planning a major release yet, so let's hold off. For now the existing logger.warning approach is acceptable.

I've created an issue so we could track this bug - https://github.com/a2aproject/a2a-python/issues/1231, leaving the PR open for now.

PR #1181, opened 2026-08-10
fix: cap push notification configs per task

ez-lbz:


src/a2a/server/tasks/inmemory_push_notification_config_store.py
```diff
@@ -7,11 +7,20 @@
...
 
+MAX_PUSH_NOTIFICATION_CONFIGS_PER_TASK = 50
+"""Maximum number of push notification configs a single task may have.
+
+Prevents a client from registering an unbounded number of webhooks for
+one task (DoS / resource exhaustion).
+"""
+
+
 class InMemoryPushNotificationConfigStore(PushNotificationConfigStore):
```
mykytanetipa, line 24: The push notification cap is only present in InMemoryPushNotificationConfigStore

DatabasePushNotificationConfigStore and any custom PushNotificationConfigStore subclasses have no cap. I think that limit should belong at the handler layer, not in one store implementation.

src/a2a/server/tasks/inmemory_push_notification_config_store.py
```diff
@@ -7,11 +7,20 @@
     PushNotificationConfigStore,
 )
 from a2a.types.a2a_pb2 import TaskPushNotificationConfig
+from a2a.utils.errors import InvalidParamsError
 
 
 logger = logging.getLogger(__name__)
 
 
+MAX_PUSH_NOTIFICATION_CONFIGS_PER_TASK = 50
```
mykytanetipa, line 16: I'd avoid setting MAX_PUSH_NOTIFICATION_CONFIGS_PER_TASK just as a global constant. If this change lands it should be configurable, opt-in, with None default and set by some server config

mykytanetipa (commented):
The push notifications configs per task cap is a reasonable idea, but it's a public-facing behavioral change, so it should be reviewed and accepted by the A2A spec maintainers first. Please open a spec proposal PR at https://github.com/a2aproject/A2A before implementing it in the SDK.

mykytanetipa (changes requested):

PR #1182, opened 2026-08-10
fix: validate task state transitions (block terminal escape)

ez-lbz:
### 1. Validate task state transitions before persisting status updates

**Problem:** `TaskManager.save_task_event()` (`src/a2a/server/tasks/task_manager.py`) blindly overwrote the task status with `task.status.CopyFrom(event.status)` for every `TaskStatusUpdateEvent`. A misbehaving or malicious agent could move a task that is already in a terminal state (COMPLETED/CANCELED/FAILED/REJECTED) back to SUBMITTED (or to another state), rewriting the persisted terminal outcome. `active_task.py:571` was already guarded (`if task.status.state not in TERMINAL_TASK_STATES`) and needed no change.

**Fix (src/a2a/server/tasks/task_manager.py):**
- Added `TERMINAL_TASK_STATES` and `validate_state_transition(current_state, new_state)`, which raises `InvalidAgentResponseError` when a terminal state would transition to any *different* state (including back to SUBMITTED). Re-persisting the same terminal state is tolerated for idempotency, and all forward transitions (SUBMITTED→WORKING→terminal, interrupted states, etc.) are unaffected.
- `save_task_event()` now calls `validate_state_transition()` before `task.status.CopyFrom(...)`. Both the V1 path (ResultAggregator → `TaskManager`) and the V2 path (`ActiveTask` EventConsumer → `TaskManager`) flow through this single chokepoint.

## Testing

- `./.venv/Scripts/python -m pytest tests/server/tasks/test_task_manager.py -q` → **29 passed** (includes new tests: parametrized rejection of every terminal→other transition, acceptance of legal forward/idempotent transitions, and an end-to-end `save_task_event` regression test asserting a completed task cannot be moved back to SUBMITTED and the store is unchanged).
- `./.venv/Scripts/python -m pytest tests/server/request_handlers/ tests/server/agent_execution/ tests/server/tasks/ tests/server/events/ -q` → **551 passed, 90 skipped, 3 xfailed** (pre-existing xfails reference upstream issue #869).
- Behavior change: status updates that would move a terminal task to a different state now fail with `InvalidAgentResponseError` (-32006) instead of silently overwriting the persisted state. All legitimate agent flows (including REJECTED/CANCELED) are unaffected.

### DIFF
--- a/src/a2a/server/tasks/task_manager.py
+++ b/src/a2a/server/tasks/task_manager.py
@@ -40,10 +40,7 @@ def validate_state_transition(
         InvalidAgentResponseError: If the transition moves a task away from
             a terminal state.
     """
-    if (
-        current_state in TERMINAL_TASK_STATES
-        and new_state != current_state
-    ):
+    if current_state in TERMINAL_TASK_STATES and new_state != current_state:
         raise InvalidAgentResponseError(
             message=(
                 f'Illegal state transition from terminal state '
--- a/src/a2a/server/tasks/task_manager.py
+++ b/src/a2a/server/tasks/task_manager.py
@@ -19,6 +19,40 @@ from a2a.utils.telemetry import trace_function
 logger = logging.getLogger(__name__)
 
 
+TERMINAL_TASK_STATES = {
+    TaskState.TASK_STATE_COMPLETED,
+    TaskState.TASK_STATE_CANCELED,
+    TaskState.TASK_STATE_FAILED,
+    TaskState.TASK_STATE_REJECTED,
+}
+
+
+def validate_state_transition(
+    current_state: TaskState, new_state: TaskState
+) -> None:
+    """Validates a task state transition before it is persisted.
+
+    Terminal states are final: a task in COMPLETED/CANCELED/FAILED/REJECTED
+    must never move to a different state (including back to SUBMITTED).
+    Re-persisting the same terminal state is tolerated for idempotency.
+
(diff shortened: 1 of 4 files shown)


src/a2a/server/tasks/task_manager.py
```diff
@@ -19,6 +19,37 @@
 logger = logging.getLogger(__name__)
 
 
+TERMINAL_TASK_STATES = {
```
mykytanetipa: I'd avoid duplicating TERMINAL_TASK_STATES here, it's already defined in src/a2a/server/agent_execution/active_task.py, should be unified

mykytanetipa (changes requested):
Thank you for this contribution, the fix is correct. Left minor refactoring comment

ez-lbz:
Done — the canonical `TERMINAL_TASK_STATES` now lives in `a2a/utils/task.py` (dependency-free), and both `task_manager.py` and `active_task.py` import it from there. `active_task` keeps re-exporting the name, so existing imports (`default_request_handler_v2`) are unchanged. I put it in utils rather than importing `active_task` from `task_manager` because the latter import direction is order-fragile: `active_task` imports from the `a2a.server.tasks` package (`PushNotificationEvent`), so a module-level `task_manager → active_task` import breaks when `active_task` is imported first. One definition, no cycles. Lint + full suite green locally (1781 passed).

PR #1183, opened 2026-08-10
fix: sanitize internal error messages in JSON-RPC and gRPC paths

ez-lbz:
### 1. Stop leaking internal exception details in JSON-RPC errors

**Problem:** The JSON-RPC dispatcher (`src/a2a/server/routes/jsonrpc_dispatcher.py`) embedded the raw exception text into client-visible errors in four places: base request validation (`InvalidRequestError(data=str(e))`), param parsing (`InvalidParamsError(data=str(e))`), the unhandled-exception path (`InternalError(message=str(e))`), and the SSE stream error path (`InternalError(message=str(e))`). The generic defensive conversion in `_generate_error_response()` also used `InternalError(message=str(error))`. This leaks internal paths, library versions, and stack details to clients (fingerprinting / information disclosure).

**Fix (src/a2a/server/routes/jsonrpc_dispatcher.py):**
- All four call sites now return the corresponding A2A error with its default (generic) message; the original exception is still logged server-side via `logger.exception(...)`.
- `_generate_error_response()` converts unknown errors to a bare `InternalError()` ("Internal error") instead of embedding `str(error)`.

### 2. Stop leaking unknown exception details in the gRPC handler

**Problem:** `GrpcHandler` (`src/a2a/server/request_handlers/grpc_handler.py`) leaked exception details in two ways: non-A2A exceptions were not caught by `_handle_unary`/`_handle_stream` and propagated to the gRPC framework with their message text; and the `abort_context()` fallback sent `Unknown error type: {error}`, embedding the exception repr.

**Fix (src/a2a/server/request_handlers/grpc_handler.py):**
- `_handle_unary` and `_handle_stream` now catch generic `Exception`, log it, and abort with `InternalError` (INTERNAL / "Internal error").
- `abort_context()`'s fallback now logs the unknown error and aborts with a generic `"Unknown error"` message instead of embedding the error repr.

## Testing

- `./.venv/Scripts/python -m pytest tests/server/routes/test_jsonrpc_dispatcher.py tests/server/request_handlers/test_grpc_handler.py tests/server/test_integration.py -q` → **97 passed** (includes new sanitization tests for the unhandled-exception, malformed-params and invalid-base-request paths, plus gRPC unhandled-exception and unknown-A2A-error paths; all assert the internal text does not appear in the response).
- `./.venv/Scripts/python -m pytest tests/compat/v0_3/ -q` → **250 passed** (v0.3 compat unaffected).
- Updated `tests/server/test_integration.py::test_unhandled_exception` to assert the generic `"Internal error"` message and the absence of the exception text (previously it asserted the leaked message).
- Behavior change: non-A2A exception messages are no longer sent to clients over JSON-RPC or gRPC; clients now receive `Internal error` / `Unknown error`. Error codes are unchanged.

### DIFF
--- a/src/a2a/server/request_handlers/grpc_handler.py
+++ b/src/a2a/server/request_handlers/grpc_handler.py
@@ -136,6 +136,9 @@ class GrpcHandler(a2a_grpc.A2AServiceServicer):
             result = await handler_func(server_context)
         except A2AError as e:
             await self.abort_context(e, context)
+        except Exception:
+            logger.exception('Unhandled exception in gRPC handler')
+            await self.abort_context(types.InternalError(), context)
         else:
             return result
         return default_response
@@ -153,6 +156,9 @@ class GrpcHandler(a2a_grpc.A2AServiceServicer):
                 yield item
         except A2AError as e:
             await self.abort_context(e, context)
+        except Exception:
+            logger.exception('Unhandled exception in gRPC handler')
+            await self.abort_context(types.InternalError(), context)
 
     async def SendMessage(
         self,
@@ -413,9 +419,13 @@ class GrpcHandler(a2a_grpc.A2AServiceServicer):
             context.set_trailing_metadata(tuple(new_metadata))
             await context.abort(rich_status.code, rich_status.details)
         else:
+            logger.error(
+                'Unknown error type during request handling',
+                exc_info=error,
+            )
             await context.abort(
                 grpc.StatusCode.UNKNOWN,
-                f'Unknown error type: {error}',
+                'Unknown error',
             )
 
     def _build_call_context(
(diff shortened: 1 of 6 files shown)


ConnorMoss02:
`grpc_handler.py:374-377` sets only `reason` and `domain` on `ErrorInfo`, and the client at `transports/grpc.py:68-72` reads back only those two, so `error.data` never makes it across gRPC. `build_error_details` says it carries `data` as `metadata` so every transport surfaces field-level violations the same way, which holds for JSON-RPC and REST but not gRPC.

ez-lbz:
Good catch — and you're right, that's a real gap in this PR. `build_error_details` claims `error.data` rides along as `ErrorInfo.metadata` so every transport surfaces the same details, but `abort_context` builds `ErrorInfo` with `reason`/`domain` only and `client/transports/grpc.py` read back only those two fields, so for anything that isn't an `InvalidParamsError` (those travel as `BadRequest`, unchanged) the data was dropped on the gRPC path.

Fixed in 048555c: the server now projects `error.data` onto `ErrorInfo.metadata` (string-valued entries, matching the JS and Rust SDKs) and the client restores it into `A2AError.data`, with `BadRequest` details keeping precedence. Tests on both sides fail without the source change.

PR #1184, opened 2026-08-10
fix: harden SSE transport and enforce request body size limit

ez-lbz:
### 1. Configure SSE heartbeat and stream timeout explicitly

**Problem:** `EventSourceResponse` was created without explicit `ping`/`send_timeout` in `src/a2a/server/routes/jsonrpc_dispatcher.py` (`_create_response`) and `src/a2a/server/routes/rest_dispatcher.py` (`_handle_streaming`). With the installed `sse-starlette`, the default `ping` is `None` — no heartbeat is emitted — so idle streams send nothing and dead clients are never detected; there is also no cap on how long a stream may sit without producing output.

**Fix (src/a2a/server/routes/jsonrpc_dispatcher.py, src/a2a/server/routes/rest_dispatcher.py):**
- Added `SSE_PING_INTERVAL_SECONDS = 15` and `SSE_SEND_TIMEOUT_SECONDS = 300` to `src/a2a/utils/constants.py`.
- Every `EventSourceResponse` created by both dispatchers now passes `ping=15` (heartbeat frame every 15s) and `send_timeout=300` (tear down a stream that makes no progress for 5 minutes).

### 2. Enforce a request body size limit

**Problem:** No limit was enforced on the HTTP request body. A client could POST an arbitrarily large payload, which the JSON-RPC (`request.json()`) and REST (`request.body()`) paths buffer fully in memory — a memory-exhaustion / DoS vector.

**Fix (src/a2a/server/routes/common.py, src/a2a/server/routes/jsonrpc_dispatcher.py, src/a2a/server/routes/rest_dispatcher.py):**
- Added `MAX_REQUEST_BODY_SIZE = 10 MiB` to `src/a2a/utils/constants.py`.
- Added `read_request_body_with_limit()` in `routes/common.py`: it fast-rejects via the `Content-Length` header when present, and otherwise streams the body in chunks with an incremental cap, so oversized chunked bodies are rejected while being read rather than buffered unboundedly. The body is cached on the request so existing `request.body()`/`stream()` consumers keep working.
- JSON-RPC: `handle_requests` now reads through the limited reader; the existing `HTTPException(413)` handling returns `InvalidRequestError("Payload too large")`.
- REST: all four body-reading sites (`_handle_streaming` pre-consume, `on_message_send`, `on_message_send_stream`, `set_push_notification`) now use a wrapper that converts the 413 into `InvalidRequestError("Payload too large")` (400).

## Testing

- `./.venv/Scripts/python -m pytest tests/server/routes/test_jsonrpc_dispatcher.py tests/server/routes/test_rest_dispatcher.py -q` → **39 passed** (includes new tests: oversized body rejected via body content and via `Content-Length`, and `ping_interval`/`send_timeout` assertions on the returned `EventSourceResponse` for both dispatchers).
- `./.venv/Scripts/python -m pytest tests/server/ tests/compat/ -q` → **924 passed, 90 skipped, 3 xfailed** (pre-existing xfails reference upstream issue #869).
- `./.venv/Scripts/python -m ruff check` on all modified files: clean.
- Behavior change: SSE streams now emit a heartbeat comment every 15s (previously none with the installed sse-starlette) and are torn down after 300s without progress; request bodies over 10 MiB are rejected with 413/`"Payload too large"` instead of being buffered.

### DIFF
--- a/src/a2a/server/routes/common.py
+++ b/src/a2a/server/routes/common.py
@@ -50,9 +50,7 @@ async def read_request_body_with_limit(request: Request) -> bytes:
     if content_length:
         try:
             if int(content_length) > MAX_REQUEST_BODY_SIZE:
-                raise HTTPException(
-                    status_code=HTTP_413_CONTENT_TOO_LARGE
-                )
+                raise HTTPException(status_code=HTTP_413_CONTENT_TOO_LARGE)
         except ValueError:
             pass
 
--- a/src/a2a/server/routes/common.py
+++ b/src/a2a/server/routes/common.py
@@ -13,12 +13,63 @@ else:
         Request = Any
         BaseUser = Any
 
+    try:
+        from starlette.exceptions import HTTPException
+    except ImportError:
+        HTTPException = Any
+
 from a2a.auth.user import UnauthenticatedUser, User
 from a2a.extensions.common import (
     HTTP_EXTENSION_HEADER,
     get_requested_extensions,
 )
 from a2a.server.context import ServerCallContext
+from a2a.utils.constants import MAX_REQUEST_BODY_SIZE
+
+
+try:
+    from starlette.status import HTTP_413_CONTENT_TOO_LARGE
+except ImportError:
+    HTTP_413_CONTENT_TOO_LARGE = Any
+
+
+async def read_request_body_with_limit(request: Request) -> bytes:
+    """Reads a request body, rejecting bodies over ``MAX_REQUEST_BODY_SIZE``.
+
+    The content-length header is checked first (fast reject), and the body
+    is then streamed in chunks with an incremental cap so oversized
+    chunked bodies are rejected while being read instead of buffered
+    unboundedly. The body is cached on the request (``request._body``) so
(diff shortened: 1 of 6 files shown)


ConnorMoss02:
There are five `EventSourceResponse` sites and this covers three. `compat/v0_3/jsonrpc_adapter.py:280` and `compat/v0_3/rest_adapter.py:99` still construct it bare, so with `enable_v0_3_compat=True` those streams keep the `ping=None` behavior.

ez-lbz:
You're right — five construction sites and I covered three. `compat/v0_3/jsonrpc_adapter.py:280` and `compat/v0_3/rest_adapter.py:99` built the response bare, so with `enable_v0_3_compat=True` those streams kept `ping=None` and the unbounded `send_timeout` this PR removes elsewhere.

Fixed in 17c6f7c: both adapters now pass `ping`/`send_timeout` from `a2a.utils.constants`, same as the v1 routes, with a test per adapter asserting the response carries them (both fail with `assert None == 300` if the source change is reverted).

PR #1185, opened 2026-08-10
fix: serialize concurrent cancel requests with a per-task lock

ez-lbz:
### 1. Serialize the cancel check-then-act with a per-task lock

**Problem:** `LegacyRequestHandler.on_cancel_task()` (`src/a2a/server/request_handlers/default_request_handler.py`) performed "load task → check terminal state → call `agent_executor.cancel()`" without any synchronization. Two concurrent cancels of the same task could both pass the terminal-state check and both issue a cancel, duplicating the cancellation side effects (TOCTOU race).

**Fix (src/a2a/server/request_handlers/default_request_handler.py):**
- Added a per-task lock map (`_cancel_locks`) so the entire check-then-act critical section in `on_cancel_task` is atomic per task. The first cancel to acquire the lock performs the cancel; any concurrent cancel waits, then re-reads the task state and fails with `TaskNotCancelableError` if the task is now terminal.
- Each entry carries a user count so the lock entry is garbage-collected once the last caller finishes (no unbounded growth), and entries are only removed when no other caller can be waiting on them.
- V2's `DefaultRequestHandlerV2` is unaffected: cancels there are already serialized per `ActiveTask` by `ActiveTaskRegistry`/`ActiveTask`.

## Testing

- `./.venv/Scripts/python -m pytest tests/server/request_handlers/test_default_request_handler.py -q` → **75 passed** (includes new `test_on_cancel_task_serializes_concurrent_cancels`: two concurrent cancels, the first blocks inside the executor, and the second must observe the resulting terminal state and raise `TaskNotCancelableError` without ever reaching the executor; also asserts the per-task lock entry is cleaned up afterwards).
- `./.venv/Scripts/python -m ruff check` on modified files: clean (the pre-existing `too-many-positional-arguments` finding on the handler `__init__` exists on `main`).
- Behavior change: concurrent cancels of the same task are now serialized; the losing request gets `TaskNotCancelableError` instead of issuing a duplicate cancel.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -151,3 +151,4 @@ typeerror
 UIDs
 vulnz
 whl
+TOCTOU
--- a/src/a2a/server/request_handlers/default_request_handler.py
+++ b/src/a2a/server/request_handlers/default_request_handler.py
@@ -203,7 +203,7 @@ class LegacyRequestHandler(RequestHandler):
 
         The terminal-state check and the cancel call are serialized with a
         per-task lock so concurrent cancels cannot both pass the check
-        (TOCTOU, BUG-44). V2's ``ActiveTaskRegistry`` already serializes
+        (TOCTOU). V2's ``ActiveTaskRegistry`` already serializes
         cancels per ``ActiveTask`` and is not affected.
         """
         task_id = params.id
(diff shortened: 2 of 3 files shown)


mykytanetipa:
Thank you for this contribution. LegacyRequestHandler is a part of deprecated 0.3.x codebase, path that exists purely for compatibility of un-migrated 1.x users and 0.3.x-compat clients.

Please let us know if it causes a real production issue for you; otherwise the recommended path is migrating to 1.x.

Issue #1175, opened 2026-08-10
cancel() and producer-failure write a terminal state to the store but not to active subscriber streams


chopmob-cloud:
Follow-up from the review of #1172 (raised by @astrogilda).

### What happens

`ActiveTask.cancel()` now closes a task out to a terminal `CANCELED` state when
the executor's `cancel()` is cleanup-only or the task was parked in a
non-terminal state (e.g. `input-required`). That terminal state is written
directly to the task store via `save_task_event`, because by that point both
the agent and subscriber event queues are already closed.

The consequence is that the terminal state reaches the **store** but not an
**active subscriber stream**. A client that was streaming the task
(`SubscribeToTask` / `message/stream`) receives the initial `Task` in `WORKING`
and then the stream ends; it never sees a `CANCELED` event and has to re-read
the task to learn the outcome. The push path does not fire either, since
`_update_task_state` only notifies on `PushNotificationEvent`.

### Why it was left out of #1172

At the point of the direct store write, both queues are already closed, so a
store write is the only option left. The same pattern already exists for the
producer-failure path (`active_task.py`, the `FAILED` write in
`_run_producer`'s `except Exception` arm), under a comment that says so. #1172
kept scope to the authorization fix (#1159) and the "cancel leaves an actionable
terminal state" fix (#1170); this stream-visibility gap is orthogonal.

### Proposed follow-up

Give a streaming or push-subscribed client a way to observe the terminal state
without polling. Options to weigh:

- Emit a synthetic terminal `TaskStatusUpdateEvent` to `_event_queue_subscribers`
  before the queues are torn down, so an active subscriber sees the transition.
- Route the terminal store write through the push path so `push_sender` fires.

Both the `cancel()` CANCELED write and the producer-failure `FAILED` write share
this gap and should get the same treatment.

### Repro

Subscribe to a live task whose executor's `cancel()` is cleanup-only, cancel it,
and collect the events the subscriber receives: the stream yields the initial
`WORKING` task and ends, while `task_store.get` afterwards returns `CANCELED`.

anthonyrs06:
I'll take the follow-up that #1172 scoped out.

Long-lived A2A clients that subscribe (`message/stream` / `SubscribeToTask`) need to observe a terminal outcome without polling. Today `cancel()` and the producer-failure path write `CANCELED`/`FAILED` to the store after both queues are closed, so a subscriber sees `WORKING` and then a hung-up stream.

[...]

PR #1191, opened 2026-08-18
fix: publish terminal task state to subscriber streams (#1175)

anthonyrs06:
## Summary

Follow-up to #1172 / #1175. After #1172, `cancel()` writes `CANCELED` to the store *after* `_is_finished.wait()`, when both event queues are already closed. A live `SubscribeToTask` / `message/stream` client saw `WORKING` and then a hung-up stream. Push did not fire either. The same store-only write existed on the producer-failure path.

This persists a *copy* of the current task (the shared `get_task()` object is not mutated) and notifies observers **before** teardown:

- **`cancel()`** — if the executor's `cancel()` is cleanup-only (or the task is still non-terminal), emit a synthetic `TaskStatusUpdateEvent(CANCELED)` to `_event_queue_subscribers` and `push_sender` in `finally` *before* `_producer_task.cancel()`, while `_run_producer.finally` cannot have closed the queues yet. #1172's cancel ordering and BaseException/`CancelledError` residual are preserved.
- **Producer failure** — persist `FAILED` to the store and fire `push_sender` before the `finally` closes the queues. A `FAILED` status event is *not* enqueued on the subscriber stream: blocking `on_message_send` treats a terminal `Task` as a successful result, so the crash must still surface as the producer exception (scenarios 9 / 12). The leftover — a stream that cannot distinguish crash from disconnect without `get_task()` — is #1228.

The helper returns early when the task is already terminal (no duplicate store/push carrying an earlier state).

Rebased onto current main after #1172.

## Test plan

- [x] `./scripts/lint.sh`
- [x] `uv run pytest` (1988 passed locally after rebase)
- [x] `uv run pytest --cov=src --cov-report=term-missing` (93%)
- [x] New: live subscriber sees `CANCELED` on `cancel()`; shared task object is not mutated; push is notified
- [x] New: producer crash persists `FAILED` + push, subscriber receives the exception (no hang)
- [x] New: already-terminal persist is a no-op (no store write, no push)
- [x] Existing #1172 cancel tests (`CancelledError` still cancels producer; cancel does not mutate shared task)
- [x] Existing blocking error scenarios 9 and 12 still raise `A2AClientError`

### DIFF
--- a/src/a2a/server/agent_execution/active_task.py
+++ b/src/a2a/server/agent_execution/active_task.py
@@ -36,6 +36,7 @@ Data Flow and Event Handling:
 from __future__ import annotations
 
 import asyncio
+import inspect
 import logging
 import uuid
 
@@ -560,17 +561,26 @@ class ActiveTask:
                 'Producer[%s]: Execution failed',
                 self._task_id,
             )
-            # Persist the failure directly instead of relying on the closing
-            # event queue to carry a final status update.
+            # Persist FAILED (store + push) before finally closes the
+            # queues (issue #1175). Do not emit a FAILED status event:
+            # blocking on_message_send would treat that Task as success.
+            # The producer exception is the stream signal.
             if request_context:
-                task = await self._task_manager.ensure_task_id(
-                    self._task_id,
-                    request_context.context_id or '',
-                )
-                if task.status.state not in TERMINAL_TASK_STATES:
-                    task.status.state = TaskState.TASK_STATE_FAILED
-                    await self._task_manager.save_task_event(task)
-                self._task_created.set()
+                try:
+                    await self._task_manager.ensure_task_id(
+                        self._task_id,
+                        request_context.context_id or '',
+                    )
+                    await self._persist_and_publish_terminal(
+                        TaskState.TASK_STATE_FAILED,
+                        publish_to_stream=False,
+                    )
+                    self._task_created.set()
+                except Exception:
+                    logger.exception(
+                        'Producer[%s]: Failed to persist FAILED state',
+                        self._task_id,
+                    )
             await self._event_queue_agent.enqueue_event(cast('Event', e))
 
         finally:
@@ -711,11 +721,11 @@ class ActiveTask:
     async def cancel(self, call_context: ServerCallContext) -> Task:
         """Cancels the running active task.
 
-        The returned task carries a terminal state, which is written to the
-        task store. That write is not guaranteed to reach an active subscriber
-        stream: by the time cancel writes it the event queues may already be
-        closed, so a client that was streaming may have to re-read the task to
-        observe the terminal state.
+        The returned task carries a terminal state written to the store.
+        When a producer is still running, that state is also published to
+        the live subscriber stream and push *before* the producer is
+        cancelled (issue #1175). If the producer has already finished, the
+        subscriber queue is already closed, so the write is store plus push.
 
         Concurrency Guarantee:
         Uses `_lock` to ensure we don't attempt to cancel a producer that is
@@ -755,7 +765,7 @@ class ActiveTask:
                 # Residual: if executor.cancel() raises a BaseException (e.g.
                 # asyncio.CancelledError), it propagates out of cancel() from
                 # the finally before the _is_finished.wait() and the CANCELED
-                # write below, leaving the task non-terminal with no producer
+                # fallback below, leaving the task non-terminal with no producer
                 # behind it. This is not the #1170 silent-success shape (the
                 # caller receives the exception), and it is recoverable: a
                 # later cancel() takes the else branch and reaches the
@@ -771,6 +781,21 @@ class ActiveTask:
                     await self._mark_task_as_failed(e)
                     raise
                 finally:
+                    try:
+                        # Cleanup-only executor.cancel() or a parked
+                        # input-required task leaves no terminal event.
+                        # Publish CANCELED while the subscriber queue is
+                        # still open — producer teardown lives in
+                        # _run_producer.finally and cannot run until
+                        # _producer_task.cancel() below.
+                        await self._persist_and_publish_terminal(
+                            TaskState.TASK_STATE_CANCELED
+                        )
+                    except Exception:
+                        logger.exception(
+                            'Cancel[%s]: Failed to persist CANCELED state',
+                            self._task_id,
+                        )
                     self._producer_task.cancel()
             else:
                 logger.debug(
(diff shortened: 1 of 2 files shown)


astrogilda (commented):
Reviewed this because it carries the third point from my #1172 review, and #1175 credits it. Read of the diff, not a run of the branch.

The `cancel()` arm is right and the ordering is the part I'd have got wrong. Emitting inside finally before self._producer_task.cancel() means the `_event_queue_subscribers` queue is still open when the TaskStatusUpdateEvent goes in, because teardown lives in _run_producer's own finally and that can't run until the producer task is actually cancelled. Putting the cancel back at the end of the finally rather than before the try also keeps the BaseException property from #1172 intact: if executor.cancel raises CancelledError, the except Exception arm doesn't run but finally does, so the producer still gets cancelled. Both of those are easy to lose in a reorder and neither is lost here.

The `publish_to_stream=False` choice on the producer-failure arm is correct for the reason in the comment, and I want to be explicit that I agree, because it looks like an inconsistency and isn't. A blocking `on_message_send` that receives a FAILED Task treats it as a successful terminal result, so emitting a FAILED status event there would convert a crash into a success for every non-streaming caller. Surfacing the producer exception instead is the right call.

The consequence is worth stating in #1175 rather than leaving implicit, because it's a scope statement about what the issue closes. #1175 asks for a terminal outcome without polling. After this PR that holds for cancel() and does not hold for producer failure: a streaming client that survives the exception still cannot distinguish "the producer crashed" from "the connection dropped" without a get_task(), which is the workaround @rohityan gave above as a stopgap. So either the issue records that the producer-failure half stays polling-dependent, or the actual root cause gets its own issue. The blocking path conflating a FAILED terminal Task with success is a defect in on_message_send, not a constraint this PR has to design around. I'd rather see that filed than have the workaround become the contract. Happy to write it up if you'd rather not.

Three smaller things.

The `_persist_and_publish_terminal` helper guards only the **store write** with `if task.status.state not in TERMINAL_TASK_STATES`. The TaskStatusUpdateEvent construction, the stream publish and the push_sender.send_notification call all run unconditionally below it. Both cancel() call sites check terminality before calling, so they're fine, but the producer-failure call site doesn't. If the task is already terminal there, from a cancel that landed first, the store write is correctly skipped and push then fires anyway, carrying task.status.state, which is whatever the earlier terminal state was rather than FAILED. That's a duplicate notification with a state the caller didn't ask to send. Moving the guard inside the helper and returning early makes all three call sites safe by construction.

The else branch, which covers a producer that already finished or never started, calls _persist_and_publish_terminal(CANCELED) with the default publish_to_stream=True. If the producer has finished, its finally has already closed the subscriber queue, and EventQueue.enqueue_event on a closed queue logs "Queue is closed. Event will not be enqueued." and returns rather than raising (`event_queue.py:119-122`). So that branch is store plus push, never stream. Benign, since no live subscriber remains, but the code and the docstring both read as publishing there. One line either way.

This call puts a tuple on a queue whose declared element type is Event, with the cast doing the silencing:

```
await self._event_queue_subscribers.enqueue_event(cast('Any', (event, updated_task_copy)))
```

If (event, task) is genuinely the subscriber queue's contract, the type wants widening rather than a cast at the call site, because a cast('Any', ...) here is the thing that lets the next wrong shape through without a diagnostic. If it isn't the contract, this is the bug. Which is it?

mykytanetipa (changes requested):
#1172 was merged so for this PR to be reviewed, please rebase onto current main. There are multiple conflicting changes.

anthonyrs06:
Rebased onto current main after #1172 and addressed @astrogilda's review.

**Review follow-ups**

1. **Already-terminal guard** — `_persist_and_publish_terminal` now returns early (no store write, no stream publish, no push) if the task is already terminal. Covers the producer-failure-after-cancel duplicate-notification case. Test: `test_persist_terminal_skips_when_already_terminal`.
2. **Else / already-finished path** — docstring and the post-`wait()` fallback now say store + push only; the subscriber queue is typically already closed there. The live-stream publish happens in `finally` *before* `_producer_task.cancel()`, while `_run_producer.finally` cannot have run yet.
3. **Tuple on the subscriber queue** — `(event, updated_task)` is the existing subscriber-queue contract (`EventConsumer._enqueue_to_subscribers` and `subscribe()` already dequeue that shape). `Event` is the *agent*-queue type; widening it would lie about that queue. Left the `cast('Any', ...)` with a comment pointing at the consumer. Happy to follow up with a dedicated subscriber-queue type if you want that split.

**#1175 scope**

`publish_to_stream=False` on producer-failure stays. The blocking `on_message_send` FAILED-as-success defect is now #1228 so this PR does not make the workaround the contract.

@mykytanetipa ready for re-review.

chopmob-cloud:
As the reporter of #1175 I verified this against current main and it closes the gap I filed.

The bug was that a cancel or producer-failure wrote the terminal state to the store but never to an active subscriber stream, so a streaming client saw WORKING and then the stream simply ended. The new `_persist_and_publish_terminal` emits a `TaskStatusUpdateEvent` to the subscriber queue before `_producer_task.cancel()` runs, and `test_cancel_publishes_canceled_to_subscriber_stream` pins exactly that: a live tap now receives CANCELED. That is the behaviour #1175 asked for.

I also checked the fallback path (producer already finished or never started), which calls the helper with the default `publish_to_stream=True`. That is safe even though the subscriber queue is usually closed by then: `EventQueue.enqueue_event` takes `_lock`, returns early on `_is_closed`, and separately swallows `QueueShutDown`, so the stream publish is a no-op there and the effective behaviour is store plus push, which matches the docstring.

The already-terminal early return in the helper is the right guard for the producer-failure-after-cancel case: it stops a second notification carrying the earlier state. Keeping `publish_to_stream=False` on producer-failure is also correct, since a blocking `on_message_send` treats a FAILED Task as a successful terminal result, so the crash must stay the stream signal; the separate blocking-send defect is tracked in #1228.

mykytanetipa's one requested change (rebase onto main after #1172) is done, the tree is mergeable, and the added tests plus the full `test_active_task.py` module pass locally. Reads ready to land.

src/a2a/server/agent_execution/active_task.py
```diff
@@ -771,6 +781,21 @@ async def cancel(self, call_context: ServerCallContext) -> Task:
...
+                    try:
+                        # Cleanup-only executor.cancel() or a parked
+                        # input-required task leaves no terminal event.
+                        # Publish CANCELED while the subscriber queue is
+                        # still open — producer teardown lives in
+                        # _run_producer.finally and cannot run until
+                        # _producer_task.cancel() below.
+                        await self._persist_and_publish_terminal(
+                            TaskState.TASK_STATE_CANCELED
+                        )
```
mykytanetipa: This may introduce state race between await "self._mark_task_as_failed(e)" and "await self._persist_and_publish_terminal(...)" in case of  _agent_executor.cancel raising 

result: subscriber sees CANCELED, but FAILED will be in the store

src/a2a/server/agent_execution/active_task.py
```diff
@@ -784,26 +809,15 @@ async def cancel(self, call_context: ServerCallContext) -> Task:
...
-            # object in place. Making every write copy-on-yield is the general
-            # property tracked in #1175 and #1191.
-            updated = Task()
-            updated.CopyFrom(task)
-            updated.status.state = TaskState.TASK_STATE_CANCELED
-            await self._task_manager.save_task_event(updated)
-            task = updated
+            persisted = await self._persist_and_publish_terminal(
+                TaskState.TASK_STATE_CANCELED
+            )
```
mykytanetipa, line 818: _persist_and_publish_terminal is called after await self._is_finished.wait() -> the producer's finally closes the subscriber queue -> await self._event_queue_subscribers.enqueue_event (line 932) -> logs a WARNING each time

Should we have here publish_to_stream=False?

mykytanetipa (changes requested):

astrogilda:
Yes to `publish_to_stream=False` on that path. By the time the terminal publish runs after wait(), the producer's finally block has already closed the subscriber queue, so that enqueue only warns. Anthony, I don't think the split is needed; the comment pointing at the consumer is enough.

PR #1193, opened 2026-08-20
fix(signing): canonicalize agent cards per RFC 8785

### LINKED ISSUE #1174: [Bug]: _canonicalize_agent_card is not RFC 8785, so signed cards with non-ASCII content fail cross-SDK verification
**What happened?**

`_canonicalize_agent_card` in `src/a2a/utils/signing.py` finishes with `json.dumps(cleaned_dict, separators=(',', ':'), sort_keys=True)`. Spec §8.4.1 requires RFC 8785, and that call is not RFC 8785 on three axes, each of which is reachable from a conformant Agent Card and each of which breaks verification against `@a2a-js/sdk`.

First, `json.dumps` defaults to `ensure_ascii=True`, so every non-ASCII character in any string is emitted as a `\uXXXX` escape. RFC 8785 §3.2.2.2 requires literal UTF-8 outside the mandatory escape set. Any card with an accent in `name`, `description` or a skill description is affected, which in practice means any card not written in English. Second, `sort_keys=True` orders keys by Unicode code point, while RFC 8785 §3.2.3 requires ordering on UTF-16 code units. The orders differ whenever one key contains a non-BMP character and another contains a BMP character at or above U+E000, since the astral character's leading surrogate 0xD83D sorts below 0xFF01 in UTF-16 but its code point U+1F600 sorts above U+FF01. Arbitrary keys reach the canonicalizer through `AgentCard.security_schemes`, which is a `map<string, SecurityScheme>`, and through `AgentExtension.params`, which is a `google.protobuf.Struct`. Third, `json.dumps` serializes floats with `repr`, so `0.000001` becomes `1e-06`, while RFC 8785 §3.2.2.3 mandates the ECMAScript `Number::toString` algorithm and therefore `0.000001`. Numbers reach the canonicalizer through `AgentExtension.params`.

The consequence is a live cross-SDK break rather than a theoretical one. I signed six cards with `create_agent_card_signer` from `a2a-sdk` 1.1.2 and verified them with `verifyAgentCardSignature` from `@a2a-js/sdk` 1.0.1 using the matching public JWK. The ASCII-only control passes, which establishes that the keys, the JWK export, the JWS construction and the proto round-trip are all fine. The non-ASCII card, the astral-key card and the numeric-params card are all rejected. The mirror direction fails identically: cards signed by the JS SDK and verified with `create_signature_verifier` raise `InvalidSignaturesError` for the same three cases and pass for the ASCII control. Every one of the six cards verifies under this SDK's own verifier, so this is not a signer/verifier asymmetry inside the SDK; the SDK is internally consistent and disagrees with the other implementation.

To settle which side is conformant rather than argue it, I re-canonicalized both SDKs' output with two RFC 8785 implementations written by neither SDK author, `rfc8785.py` in Python and `gowebpki/jcs` v1.0.1 in Go. Both agree on every case: the JS output is RFC 8785 on all three axes and this SDK's output is not on any of them. The Go oracle, fed this SDK's canonical string as input, reproduces the JS canonical string byte for byte.

A minimal reproduction, needing only `pip install "a2a-sdk[signing]" rfc8785`:

```python
import json, rfc8785
from a2a.types import AgentCard, AgentCapabilities, AgentInterface, AgentSkill
from a2a.utils.signing import _canonicalize_agent_card

card = AgentCard(
    name="Café Agent", description="Planifie des itinéraires.", version="1.0.0",
    supported_interfaces=[AgentInterface(url="https://example.com/a2a/v1",
                                         protocol_binding="JSONRPC", protocol_version="1.0")],
    capabilities=AgentCapabilities(streaming=True),
    default_input_modes=["text/plain"], default_output_modes=["text/plain"],
    skills=[AgentSkill(id="route", name="route", description="Plan a route.", tags=["maps"])],
)
sdk = _canonicalize_agent_card(card)
ref = rfc8785.dumps(json.loads(sdk)).decode()
print("a2a-python:", sdk)
print("RFC 8785  :", ref)
print("equal     :", sdk == ref)   # False
```

The fix I would suggest is to replace the `json.dumps` call with a real RFC 8785 implementation rather than to patch the three symptoms, because the three axes found here are the ones reachable from an Agent Card today and the RFC has more rules than that; a card gaining a numeric or exotic-key field later would reintroduce the class. `rfc8785` on PyPI is a pure-Python implementation with no C dependency and would be a one-line change inside `_canonicalize_agent_card`. If a new dependency in the `signing` extra is unwelcome I am happy to send the hand-rolled version instead, but it will be materially longer than three lines because ES6 number serialization is the hard part.

Note that fixing this alone changes the bytes this SDK signs, so it is a wire-visible change for anyone already signing non-ASCII cards. I would expect it to ship with the same release discipline as any other canonical-form change, and it is worth landing alongside the specification clarification I have filed at a2aproject/A2A#2122, since a second, separate divergence exists where both SDKs agree with each other and disagree with the specification's worked example. That one is not a Python-specific bug and I have raised it against the spec rather than here.

astrogilda:
Fixes #1174.

The `_canonicalize_agent_card` function builds the bytes that a signature covers, so whatever it produces is what a verifier in another SDK has to reproduce exactly. Today it ends in one call:

```
json.dumps(cleaned_dict, separators=(',', ':'), sort_keys=True)
```

That does not produce RFC 8785 on any of the three axes the scheme exists to fix, and I ran each one against the code as it stands before changing it rather than reasoning about it. A card named Café Agent serializes its name as `{"name":"Caf\u00e9 Agent"}`, where JCS section 3.2.2.2 asks for the literal UTF-8 bytes `{"name":"Café Agent"}` and escapes only the C0 range, the quote and the backslash. sort_keys=True orders keys by code point, where section 3.2.3 asks for UTF-16 code unit order, so a card whose extension params carry the keys U+1F600 and U+FF01 comes out in the opposite order from every conforming implementation. And repr formats numbers, where section 3.2.2.3 asks for the ECMAScript Number::toString algorithm, so an extension param of 0.000001 is written 1e-06 and an integral param of 1 is written 1.0. Every one of those four canonical forms disagrees with an independent implementation of the RFC; after the change all four agree byte for byte.

There is a fourth divergence underneath the three in the issue, and it is the one that would have made a fix on the first three ineffective. The signer parsed the canonical string back into a dict and handed the dict to jwt.encode, which serializes its payload with its own json.dumps and therefore re-applies ensure_ascii=True to bytes that had just been canonicalized. The canonical form would have been correct and the bytes actually signed would still have been escaped. The signer now hands the canonical bytes to the JWS layer directly. The protected header is byte-identical either way, which the tests assert against a jwt.encode reference so that this cannot regress into a silent compatibility break in the header a verifier reads kid and alg out of.

On the question you left open, of vendoring a package versus writing the algorithm here: I wrote it here, and I put the package in as a test-only oracle instead, because that is strictly stronger than either option on its own. The `rfc8785` package on PyPI is Trail of Bits' implementation, Apache-2.0, a pure-Python py3-none-any wheel with no runtime dependencies of its own and requires-python >=3.8, so it installs cleanly across the whole 3.10 to 3.14 matrix; its most recent release is 0.1.4 from 2024-09-27, which for a finished implementation of a frozen RFC is a reasonable place to sit rather than a warning sign. The reason not to make it a runtime dependency of `a2a-sdk[signing]` is that the signing extra is the code that decides whether a card is trustworthy, and a dependency there is one more package an attacker has to compromise to change what a signature means. The reason to use it anyway is that if it were the implementation, nothing would be checking it. As the oracle it checks this implementation on every accepted vector, alongside a corpus whose expected values two other implementations already agree on, so the change lands with three-way agreement instead of one-way trust. The expensive half is number formatting and I did not write that from scratch either: it is adapted from the reference implementation at `cyberphone/json-canonicalization`, Apache-2.0, credited in the module docstring, and then checked against the oracle over twenty thousand seeded random doubles and every mantissa-and-exponent pair near the notation thresholds. If you prefer that the SDK depend on rfc8785 at runtime, say so and I will make that swap; it is a one-line change to _canonicalize_agent_card and the entire test suite here stays valid, because the tests are written against the RFC rather than against this implementation.

The tests are lifted rather than invented. The 57 vectors in `tests/utils/jcs_vectors.json` come verbatim from the language-neutral `a2a-jcs-v01` corpus proposed in a2aproject/a2a-tck#228, which is 47 accept, 10 reject, and five groups covering signature exclusion, key ordering, string serialization, number serialization, and arrays and nesting. Each vector carries its expected canonical form as a UTF-8 hex string, so a transcription mistake cannot hide inside an editor's encoding, and each expected value was produced by two independent implementations written by neither SDK author which agree byte for byte. I verified the lift programmatically rather than by eye: all 57 ids present, no extras, and zero mismatches on disposition, input and expected hex against the vector files in that PR. On top of the corpus the suite adds the cases this ecosystem has already found implementations disagreeing on: unpaired surrogates in keys and in values, noncharacters, characters above the BMP, U+2028 and U+2029, negative zero, denormals, the 1e21 and 1e-7 notation boundaries, integers past the exact-integer range, and nesting on both sides of the depth bound. That is 165 tests, and the full suite is 1950 passed with ruff check, ruff format --check and ty check all clean.

Then I tried to break it, which is where the depth bound comes from. Nesting is attacker-controlled: `AgentExtension.params` is a `google.protobuf.Struct` and nests arbitrarily, and an unbounded recursive serializer turns a hostile card into a stack exhaustion in whoever verifies it rather than a rejected signature. Serialization stops at 128 levels, and the `_clean_empty` helper needed the same bound because it runs first and would otherwise be the crash site while the serializer's bound sat unreachable behind it. A card past the bound now fails as `InvalidSignaturesError` on the verify path rather than as a new exception type at callers. I also checked the mutations: reversing the key order to code point, restoring ensure_ascii, restoring repr for numbers, removing the depth cap, emitting -0, and removing the surrogate rejection each turn the suite red, at 7, 35, 19, 4, 4 and 6 failures respectively. These tests fail for the reasons they claim to exist.

Two things worth knowing that are not this change. The compatibility break is real but narrow, and I measured its edges rather than describing them: a pure-ASCII card with no extension params canonicalizes identically before and after, and so does one whose extension params are ASCII strings, so those signatures keep verifying. A card with any non-ASCII text anywhere, or any numeric extension param, produces different bytes and its existing signatures will not verify, and that is the point: those are exactly the cards that do not verify in other implementations today. Separately, while attacking the signing path I found that _clean_empty lets a card carry content the signature does not cover. A card whose extension params include `"policy": ""` canonicalizes to the same bytes as a card with no policy key at all, so a signature made over the second verifies the first. That behaviour predates this change and fixing it would be a second and larger break, so I have deliberately left it alone here; I have a reproducer and I am happy to open it as its own issue if you want it tracked.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -37,9 +37,12 @@ coc
 codegen
 coro
 culsans
+cyberphone
 datamodel
 datapart
 deepwiki
+denormal
+denormals
 drivername
 DSNs
 dunders
@@ -56,6 +59,7 @@ GBP
 genai
 getkwargs
 gle
+gowebpki
 GVsb
 hazmat
 HS256
@@ -91,6 +95,8 @@ middleware
 mikeas
 mockurl
 mysqladmin
+noncharacter
+noncharacters
 notif
 npx
 oauthoidc
@@ -128,6 +134,7 @@ resub
 rmi
 RS256
 RUF
+Rundgren
 SECP256R1
 SFIXED
 SLF
@@ -135,6 +142,7 @@ socio
 sse
 starlette
 Starlette
+stringified
 subgids
 subuids
 sut
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,10 @@ dev = [
   "pytest-mock>=3.14.0",
   "pytest-xdist>=3.6.1",
   "respx>=0.20.2",
+  # Test-only oracle. The RFC 8785 conformance tests check the canonicalizer
+  # against an independent implementation, so a mistake shared between the
+  # implementation and its tests still shows up. Not a runtime dependency.
+  "rfc8785>=0.1.4",
   "ruff>=0.12.8",
   "uv-dynamic-versioning>=0.8.2",
   "types-protobuf<7",
--- a/src/a2a/utils/_jcs.py
+++ b/src/a2a/utils/_jcs.py
@@ -1,0 +1,227 @@
+"""RFC 8785 JSON Canonicalization Scheme (JCS).
+
+The Agent Card signature covers a canonical serialization of the card, so the
+bytes produced here are the bytes that get signed. `json.dumps` cannot produce
+them: it escapes non-ASCII by default, orders keys by code point rather than by
+UTF-16 code unit, and formats numbers with `repr` rather than with the
+ECMAScript `Number::toString` algorithm.
+
+Number formatting follows ECMA-262 7.1.12.1 as amended by RFC 8785 section
+3.2.2.3. That part of the algorithm is adapted from Anders Rundgren's reference
+implementation, which is published under the Apache License 2.0:
+https://github.com/cyberphone/json-canonicalization
+
+Serialization is depth-limited. Nesting is attacker-controlled through
+`AgentExtension.params`, which is a `google.protobuf.Struct` and so may nest
+arbitrarily, and an unbounded recursive serializer turns that into a crash in
+whatever process verifies the card.
+"""
+
+from __future__ import annotations
+
+import math
+import re
+
+from typing import Any
+
+
+MAX_DEPTH = 128
+"""Maximum object/array nesting accepted by `canonicalize`.
+
+Deep enough that no plausible Agent Card reaches it, shallow enough that the
+recursion cannot exhaust the interpreter stack.
+"""
+
+# JSON numbers are IEEE 754 double-precision floats (RFC 8785 section 3.2.2.3),
+# so integers outside this range have no interoperable representation.
+_SAFE_INT_MAX = 2**53 - 1
+_SAFE_INT_MIN = -(2**53) + 1
+
+# ECMA-262 7.1.12.1 steps 6 and 8: outside this exponent window a number is
+# written in exponential notation, inside it the digits are written out in full.
+_EXPONENT_EXPANSION_CEILING = 21
+_EXPONENT_EXPANSION_FLOOR = -7
+
+# RFC 8785 section 3.2.2.2: escape the two mandatory characters and the C0
+# control range, and emit everything else as literal UTF-8.
+_ESCAPE_RE = re.compile(r'[\x00-\x1f\\"]')
+_ESCAPE_MAP = {
+    '\\': '\\\\',
+    '"': '\\"',
+    '\b': '\\b',
+    '\f': '\\f',
+    '\n': '\\n',
+    '\r': '\\r',
+    '\t': '\\t',
+}
+for _codepoint in range(0x20):
+    _ESCAPE_MAP.setdefault(chr(_codepoint), f'\\u{_codepoint:04x}')
+del _codepoint
+
+
+class CanonicalizationError(ValueError):
+    """Raised when a value has no RFC 8785 canonical form."""
+
+
+def canonicalize(obj: Any) -> str:
+    """Serializes `obj` to its RFC 8785 canonical form.
+
+    Args:
+        obj: A JSON value: `None`, `bool`, `int`, `float`, `str`, or a list or
+            dict of the same. Dict keys must be strings.
+
+    Returns:
+        The canonical serialization. The result is `str` rather than `bytes`
+        for the caller's convenience; RFC 8785 defines the canonical form as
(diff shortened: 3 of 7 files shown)


mykytanetipa (approved):
LGTM

PR #1194, opened 2026-08-20
fix: prevent first-owner write loss in in-memory stores

anxkhn:
## Summary

`_InMemoryTaskStoreImpl.save` and `InMemoryPushNotificationConfigStore.set_info` created owner buckets before acquiring their existing `RLock`. Two threads could both observe a missing owner, create separate buckets, and overwrite the bucket containing the first completed write.

This moves owner-bucket creation inside the lock with `setdefault` in both stores. Threaded regression tests cover concurrent first writes and confirm that both records remain available through the public retrieval methods.

## Testing

`uv run pytest tests/server/tasks/test_inmemory_task_store.py tests/server/tasks/test_inmemory_push_notifications.py -q`

Result: 40 passed.


mykytanetipa (approved):
LGTM

Issue #1192, opened 2026-08-19
message/send JSON-RPC result is wrapped in the SendMessageResponse oneof instead of returning the Task/Message directly


scourtney-godaddy:
## Environment
- a2a-sdk 1.0.3
- Python 3.13
- Transport: JSON-RPC (`create_jsonrpc_routes`, `DefaultRequestHandlerV2`), `enable_v0_3_compat=True`

## Summary
For a 1.x `message/send` call, the JSON-RPC `result` comes back as `{"task": {…}}` (or `{"message": {…}}`) rather than the Task or Message object itself. A client that reads `result.status` and `result.artifacts`, which the A2A 1.0 spec describes for the unary send, finds nothing, because the payload sits one level deeper under `result.task`.

The `task`/`message` oneof belongs to the streaming `StreamResponse`. The unary `message/send` should return the Task or Message directly, the way `tasks/get` and `tasks/cancel` already do.

## Expected
```json
{ "jsonrpc": "2.0", "id": "1", "result": { "id": "…", "contextId": "…", "status": {…}, "artifacts": [ … ] } }
```

## Actual
```json
{ "jsonrpc": "2.0", "id": "1", "result": { "task": { "id": "…", "contextId": "…", "status": {…}, "artifacts": [ … ] } } }
```

## Root cause
In `a2a/server/routes/jsonrpc_dispatcher.py`, `_handle_send_message` wraps the return value in a `SendMessageResponse` before serializing:

```python
async def _handle_send_message(self, request_obj, context):
    task_or_message = await self.request_handler.on_message_send(request_obj, context)
    if isinstance(task_or_message, Task):
        return MessageToDict(SendMessageResponse(task=task_or_message))
    return MessageToDict(SendMessageResponse(message=task_or_message))
```

`MessageToDict` on a `SendMessageResponse` emits the oneof field name (`task` or `message`) as a key, which becomes the JSON-RPC `result`. The sibling handlers serialize the object directly and do not show this wrapper:

```python
async def _handle_get_task(self, request_obj, context):
    task = await self.request_handler.on_get_task(request_obj, context)
    return MessageToDict(task, preserving_proto_field_name=False)
```

## Impact
A2A 1.0 clients that follow the spec read the Task at `result`, not at `result.task`. Against this SDK they see an empty result and report the call as incomplete. The two shapes also disagree within one SDK: `tasks/get` returns the Task un-nested, `message/send` nests it, so a client cannot use one code path for both.

## Reproduce
[...]

PR #1196, opened 2026-08-21
fix(server,client): return Task/Message directly from message/send

richardsonlima:
**Fixes** #1192

Right now message/send on the JSON-RPC transport returns the result nested one level too deep, like {"task": {...}} instead of just the Task object. Turns out jsonrpc_dispatcher.py was wrapping the return value in a SendMessageResponse before serializing it, but that oneof is meant for the streaming StreamResponse, not for the unary send. tasks/get and tasks/cancel already return their payload unwrapped, so message/send was the one behaving differently, and any client that follows the spec and reads result.status or result.artifacts just gets nothing back.

Fixed _handle_send_message to serialize task_or_message directly, same as the get/cancel handlers do.

While testing this I ran into something the original issue didn't mention: the client side transport (JsonRpcTransport.send_message) parses the result straight into a SendMessageResponse object, which only knows about the task/message fields. Once the server stopped wrapping the response, this started throwing a ParseError because fields like id and status don't exist on SendMessageResponse. So a client and server from this same SDK would break talking to each other after the fix.

I extended the client transport to handle this, then kept finding more shapes once I ran it against the ITK cross-SDK suite, which spins up real Go and Python peer implementations instead of mocks. Turns out real peers send the payload in at least three different shapes: the old wrapped {"task": {...}}, the new unwrapped shape this fix produces, and a spec-compliant unwrapped shape that carries a "kind": "task"/"message" discriminator on every Task, Message, and Part in the tree, including status.message, history entries, and message parts, none of which our protobuf-generated types declare. The client now checks for the wrapped shape first, then reads or infers the kind, strips it recursively from the whole payload, and parses into Task or Message accordingly, wrapping the result back into a SendMessageResponse so nothing downstream in BaseClient has to change. The v0.3 compat transport already does something similar for its own wire format, this just applies the same idea to the 1.0 JSON-RPC path.

One thing I want to flag instead of quietly leaving out: two ITK scenarios still fail after all of this, Star Topology Full JSONRPC & GRPC and Push Notification Test JSONRPC & GRPC, both specifically on the leg where current calls go_v10 over unary JSON-RPC. I went through the full raw logs and there is no exception anywhere, current's response to that call just comes back short. The same call to the same peer completes fully over grpc and http_json in the same run. Best explanation I have is that grpc always runs in streaming mode in itk/main.py so it waits for the terminal event, go_v10's REST handler evidently blocks until the task is done, but its JSON-RPC handler returns before the task completes. That is not a response shape bug, it looks like an async lifecycle behavior specific to go_v10's JSON-RPC handler, or a gap in the ITK test agent not polling after a non-terminal response, and either way it lives outside this repo. I do not think it should block this PR since what #1192 asked for is fixed and tested, but wanted to leave a clear trail in case a maintainer wants to chase it further or file it against a2a-go or a2a-itk.

**Changed files:**

src/a2a/server/routes/jsonrpc_dispatcher.py, the actual fix
src/a2a/client/transports/jsonrpc.py, client side parsing update covering the legacy wrapper, the kind discriminator, and the recursive strip
tests/client/transports/test_jsonrpc_client.py, new tests for the wrapped shape and for kind-discriminated payloads, both flat and nested
tests/server/test_integration.py
tests/integration/test_client_server_integration.py
tests/client/test_auth_interceptor.py
tests/integration/test_tenant.py

those last four just had mocks or assertions updated to match the new unwrapped shape.

**Test plan:**

ran the full suite, 1795 passed, nothing broken
ruff check and ruff format --check both clean
manually confirmed the round trip test that broke when I only fixed the server side now passes with both sides updated
ran against the ITK cross-SDK suite, went from 8 failing scenarios to 2, the remaining 2 are the go_v10 JSON-RPC timing issue described above, unrelated to response shape.

### DIFF
--- a/src/a2a/client/transports/jsonrpc.py
+++ b/src/a2a/client/transports/jsonrpc.py
@@ -81,13 +81,22 @@ class JsonRpcTransport(ClientTransport):
         json_rpc_response = JSONRPC20Response(**response_data)
         if json_rpc_response.error:
             raise self._create_jsonrpc_error(json_rpc_response.error)
-        # The server returns the Task or Message directly (not wrapped in
-        # the streaming SendMessageResponse oneof), so try each in turn and
-        # re-wrap for callers that expect a SendMessageResponse.
+        result = json_rpc_response.result
+        # Servers that still nest the payload under the streaming
+        # SendMessageResponse oneof (older SDKs, other language
+        # implementations) send {"task": {...}} or {"message": {...}}.
+        # Unwrap explicitly for those before falling back to the direct
+        # Task/Message shape this SDK's own server now returns.
+        if isinstance(result, dict) and 'task' in result:
+            task = json_format.ParseDict(result['task'], Task())
+            return SendMessageResponse(task=task)
+        if isinstance(result, dict) and 'message' in result:
+            message = json_format.ParseDict(result['message'], Message())
+            return SendMessageResponse(message=message)
         try:
-            task = json_format.ParseDict(json_rpc_response.result, Task())
+            task = json_format.ParseDict(result, Task())
         except json_format.ParseError:
-            message = json_format.ParseDict(json_rpc_response.result, Message())
+            message = json_format.ParseDict(result, Message())
             return SendMessageResponse(message=message)
         return SendMessageResponse(task=task)
 
--- a/src/a2a/client/transports/jsonrpc.py
+++ b/src/a2a/client/transports/jsonrpc.py
@@ -85,18 +85,34 @@ class JsonRpcTransport(ClientTransport):
         # Servers that still nest the payload under the streaming
         # SendMessageResponse oneof (older SDKs, other language
         # implementations) send {"task": {...}} or {"message": {...}}.
-        # Unwrap explicitly for those before falling back to the direct
-        # Task/Message shape this SDK's own server now returns.
         if isinstance(result, dict) and 'task' in result:
             task = json_format.ParseDict(result['task'], Task())
             return SendMessageResponse(task=task)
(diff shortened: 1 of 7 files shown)


scourtney-godaddy:
I tested this branch against four production A2A servers under webmesh.ai (agent, impact, dnsdoc, and seo). 

All four return bare Task payloads ({id, contextId, status, artifacts}) with no kind field. The branch parsed every response into a Task without error. Stock a2a-python 1.1.2, by contrast, fails on the same payloads with ParseError: SendMessageResponse has no field named "id".

richardsonlima:
Thanks for testing this against real servers, that’s exactly the validation I wanted before merging.  It resolves without kind because of a small check I added: Task always has id, Message always has messageId, so I use whichever one shows up to infer the shape. In your case that means it hits the Task branch right away, no fallback needed. That ParseError on 1.1.2 also confirms the bug from #1192, the old client expects the result already wrapped in task or message, so a bare object breaks it. One thing I still want to check: whether any of those servers ever return a bare Message. If you hit that case, let me know if it resolves the same way through messageId.

richardsonlima:
Hey @scourtney-godaddy, just bumping this since it's been a few days quiet after the last comment.

Quick status check: CI's green, the whole suite still passes (1795 tests, 93% coverage), and I added regression tests covering the exact case from #1192 , the bare Task payload with no `kind` field that the production servers were sending, so it's not just something that got manually confirmed once, it's actually pinned down in the test suite now. Old peers that still wrap the response in `SendMessageResponse` keep working too.

Let me know if there's anything you want me to change or add before merging, happy to take another pass at it.

PR #1198, opened 2026-08-23
docs(samples): add Agent Card signing sample

### LINKED ISSUE #1197: [Feat]: Add a runnable sample for Agent Card signing and verification

The SDK ships Agent Card signing support in `a2a.utils.signing`
(`create_agent_card_signer` / `create_signature_verifier`), and both
`A2ACardResolver.get_agent_card` and `ClientFactory.create_from_url` accept a
`signature_verifier`. But `samples/` only contains `hello_world_agent.py` and
`cli.py`, so there is no end-to-end example showing how the two halves fit
together.

That leaves a few things easy to get wrong, and all of them fail in ways that
are hard to diagnose:

- **The verifier is a synchronous callable.** The natural implementation of a
  `key_provider` — fetch the JWKS from the signature's `jku` with a sync HTTP
  client — does blocking I/O inside async code. Against an in-process server it
  deadlocks, and the symptom is a bare `InvalidSignaturesError: No valid
  signature found`, which points at the crypto rather than at the I/O.
- **`create_agent_card_signer` mutates the card it signs** (it appends to
  `card.signatures`). Wiring it directly into
  `create_agent_card_routes(card_modifier=...)` means the served card
  accumulates a new signature on every request.
- **`key_provider` failures need to be `PyJWTError` subclasses.** The verifier
  only catches `PyJWTError` while iterating signatures, so a provider raising
  `ValueError` aborts the whole card instead of moving on to the next
  signature.
- **Trusting the card's own `jku` defeats the purpose.** The `jku` header lives
  in the untrusted card, so a verifier that fetches keys from whatever URL the
  card names will happily verify a forged card against the attacker's own key.
  The key source has to be pinned out of band, and nothing in the API signature
  makes that obvious.

A runnable sample under `samples/` covering both halves of card signing:

- **Server**: sign the Agent Card with an ES256 key via
  `create_agent_card_signer`, serve it through
  `create_agent_card_routes(card_modifier=...)`, and publish the matching
  public key as a JWKS document at `/.well-known/jwks.json` — the URL the
  signature's `jku` points to.
- **Client**: verify the fetched card with `create_signature_verifier` passed
  to `A2ACardResolver.get_agent_card(signature_verifier=...)`, resolving keys
  by `kid` from an allowlist of pinned JWKS URLs, with the accepted algorithms
  pinned too.
- **Negative cases**, so the security property is visible rather than asserted:
  a card whose transport URL was rewritten in transit, a card with its
  signature stripped, and a genuine card whose `jku` the client does not trust.

Plus a smoke test in the style of `tests/integration/test_samples_smoke.py` and
a `samples/README.md` section.

Documenting the flow in prose only (e.g. in the SDK docs). The unit tests in
`tests/utils/test_signing.py` already cover the crypto, but they use symmetric
keys and an inline `key_provider`, so they do not surface the async, mutation
or key-pinning pitfalls above — those only show up once a real server and
client are wired together.

### Additional context

_No response_

### Code of Conduct

orangecola:
There was no runnable example for `a2a.utils.signing`, so this adds one
covering both halves of Agent Card signing.

`samples/agent_card_signing.py`:

- **Server**: generates an ES256 key, signs the Agent Card with
  `create_agent_card_signer`, serves it via
  `create_agent_card_routes(card_modifier=...)`, and publishes the public key
  as a JWKS document at `/.well-known/jwks.json` — the URL the signature's
  `jku` points to.
- **Client**: verifies the card with `create_signature_verifier` passed to
  `A2ACardResolver.get_agent_card(signature_verifier=...)`, resolving keys by
  `kid` from an allowlist of pinned JWKS URLs, with algorithms pinned to
  `ES256`.
- **Modes**: `demo` (default, runs both in-process), `serve`, `verify --url`.

The default `demo` mode verifies the genuine card and then shows the three
rejections a verifier exists for:

```
1. The card served by the agent
   verified card for agent: Signed Card Agent
   protected header: {'alg': 'ES256', 'jku': '.../.well-known/jwks.json', 'kid': 'sample-card-key-1', 'typ': 'JOSE'}

2. A card whose transport URL was rewritten in transit
   rejected as expected: No valid signature found

3. A card with its signature stripped off
   rejected as expected: AgentCard has no signatures to verify.

4. The genuine card, checked by a client that pins other keys
   rejected as expected: No valid signature found
```

Two non-obvious details are called out in comments, since they bite anyone
copying the pattern:

- The verifier is a **synchronous** callable, so fetching the JWKS inside it
  with a sync HTTP client blocks the event loop (it deadlocks against an
  in-process server and reports `No valid signature found`). The key provider
  does pure in-memory lookups over JWKS prefetched asynchronously.
- `create_agent_card_signer` **mutates** the card it signs, so a per-request
  `card_modifier` would accumulate signatures. The helper deep-copies first and
  the card is signed once at startup.

The key provider also raises `PyJWKError` rather than a plain exception, so the
verifier treats one unusable signature as unverifiable and moves to the next
instead of aborting a card that also carries a good signature.

Also adds `tests/integration/test_agent_card_signing_smoke.py` (runs the demo
on a free port, asserts one acceptance and three rejections) and a
`samples/README.md` section.

No library code changed — samples, tests and docs only.

## Testing

- `uv run pytest tests/utils/test_signing.py tests/integration/test_agent_card_signing_smoke.py tests/client/test_card_resolver.py` → 55 passed
- `uv run ruff format --check` and `uv run ruff check` → clean
- `uv run ty check` → same 4 pre-existing diagnostics as `main` (samples are outside `[tool.ty.src]`)

## Checklist

Fixes #1197 🦕


mykytanetipa (changes requested):
Presubmits need to be green before review can be granted. 

Only failure is Check Spelling: it flags SECP at samples/agent_card_signing.py. Since check-spelling splits on digits, add a bare SECP to .github/actions/spelling/allow.txt (the existing SECP256R1 won't match).

mykytanetipa (approved):
LGTM

mykytanetipa:
Please resolve conflicts in .github/actions/spelling/allow.txt

PR #1209, opened 2026-08-25
docs(server): say what the evict-on-full check actually tests

astrogilda:
Follow-up to #1137, documentation only. No behavior change: the diff touches two docstrings, one comment and one log format string. Ruff's linter and formatter are both clean on the two files, and the 73 tests in the event-queue and active-task modules pass.

#1137 added eviction so one sink whose queue has filled cannot wedge the dispatcher for every other subscriber. The behavior is right and the words around it are not. The guard is `if sink._evict_on_full and sink.queue.full():`, which tests queue fullness at delivery time and says nothing at all about the consumer. The docstring and the warning both described it as an abandoned subscriber whose consumer had stopped draining, which is a claim the check never makes. A consumer that is alive and reading steadily, merely `max_queue_size` events behind the dispatcher, satisfies the same predicate and is evicted the same way.

That gap costs an operator real time. The warning as written sends someone looking for a dead consumer, and what they will find is a live one that is slow, with nothing in the message saying that is expected. So the warning now says the queue was full at delivery time and that the sink is therefore at least max_queue_size events behind, and it states outright that this does not mean the consumer stopped reading.

Alongside it, the delivery docstring gains one paragraph saying what the test is and what it is not, the evict_on_full parameter doc names the queue bound rather than abandonment, and the comment at the subscriber tap site says the same thing. Abandonment stays where it is genuinely the subject: the teardown docstring still describes a producer wedged closing an abandoned sink, and the regression test still calls its never-drained sink abandoned, because in both of those cases that is what is happening.


src/a2a/server/events/event_queue_v2.py
```diff
@@ -99,9 +103,12 @@ async def _deliver_to_sink(
...
             logger.warning(
-                'Evicting event queue sink %r: its queue is full and the '
-                'consumer is not draining it. Closing the sink so dispatch '
-                'to the remaining sinks continues.',
+                'Evicting event queue sink %r: its queue was full at delivery '
+                'time, so this sink is at least max_queue_size events behind '
+                'the dispatcher. Closing the sink so dispatch to the remaining '
+                'sinks continues. This does not mean the consumer stopped '
+                'reading: a consumer that is still reading but lagging by more '
+                'than the bound is evicted the same way.',
```
mykytanetipa: it's too much prose for a per-eviction log line that can fire repeatedly under load. 

I'd suggest to keep it as short as possible, smth like this: 
```
'Evicting event queue sink %r: queue full at delivery time '
'(>= max_queue_size events behind); closing it so dispatch continues.'
```

src/a2a/server/agent_execution/active_task.py
```diff
@@ -633,10 +633,12 @@ async def subscribe(
...
-        # Subscriber sinks belong to remote consumers that can be abandoned
-        # (e.g. a non-blocking send whose HTTP response already returned).
-        # evict_on_full keeps one undrained subscriber from wedging dispatch
-        # for every other subscriber and, transitively, the producer.
+        # Subscriber sinks belong to remote consumers that can fall behind
+        # the dispatcher, either because the consumer went away (e.g. a
+        # non-blocking send whose HTTP response already returned) or because
+        # it is merely slow. evict_on_full keeps one sink whose queue has
+        # filled from wedging dispatch for every other subscriber and,
+        # transitively, the producer.
```
mykytanetipa: this comment feels redundant, the eviction logic is already explained in _deliver_to_sink() and tap(), just some 1 liner with evict_on_full=True mention should be enogh

src/a2a/server/events/event_queue_v2.py
```diff
@@ -72,25 +72,29 @@ async def _deliver_to_sink(
...
+        this dispatcher's ``gather`` until the sink drains: if it never does,
+        the incoming queue then fills, producers wedge in ``enqueue_event``,
+        and the task's event flow never recovers. Sinks tapped with
+        ``evict_on_full=True`` are therefore treated as broadcast observers:
+        when full, the sink is force-closed and detached so dispatch to the
+        remaining sinks continues and blocked producers recover. Sinks with
+        ``evict_on_full=False`` (including the default sink) keep the blocking
+        put, which is the documented flow-control contract.
+
+        The eviction test is queue fullness at delivery time and nothing else.
```
mykytanetipa: eviction condition?

mykytanetipa (changes requested):

src/a2a/server/events/event_queue_v2.py
astrogilda: Taken verbatim, thank you. It fires per eviction and can repeat under load, and a paragraph is the wrong shape for that.

src/a2a/server/agent_execution/active_task.py
astrogilda: Agreed, cut to one line naming the flag and pointing at _deliver_to_sink() for the mechanism.

src/a2a/server/events/event_queue_v2.py
astrogilda: Renamed to eviction condition, and the paragraph now states it once instead of restating it for the departed consumer and the slow one separately.

mykytanetipa (approved):

PR #1212, opened 2026-08-27
fix: omit artifacts from list tasks responses

### LINKED ISSUE #1211: [Bug]: ListTasks breaks the `includeArtifacts=false` spec "MUST omit" requirement
Specification [(3.1.4)](https://a2a-protocol.org/latest/specification/#314-list-tasks)

> When `includeArtifacts` is false (the default), the artifacts field MUST be omitted entirely from each Task object in the response. The field should not be present as an empty array or null value.

Actual behavior:
```python
from google.protobuf.json_format import MessageToDict

from a2a.types import ListTasksResponse, Task, TaskState, TaskStatus

task = Task(
    id="task-1",
    context_id="context-1",
    status=TaskStatus(state=TaskState.TASK_STATE_WORKING),
)
task.ClearField("artifacts")

response = ListTasksResponse(tasks=[task])
data = MessageToDict(
    response,
    preserving_proto_field_name=False,
    always_print_fields_with_no_presence=True,
)

print(data["tasks"][0]["artifacts"]) 
```

Returns `[]`.

### Code of Conduct

dawNotPoi:
# Description

The A2A specification requires each Task in a ListTasks response to omit the `artifacts` field when `includeArtifacts` is false or unset. The protobuf response already excludes artifact content, but JSON serialization with default fields enabled emitted `artifacts: []`.

This change centralizes ListTasks wire serialization for REST and JSON-RPC. It preserves the existing pagination fields while removing the `artifacts` key unless the request explicitly enables it. When enabled, the field remains present with its actual content, including an empty array for tasks without artifacts.

Regression tests cover both the default behavior and `includeArtifacts=true` across both protocol bindings.

- [x] Make the Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [x] Ensure the tests and linter pass (`bash scripts/format.sh` / project lint checks).
- [x] Appropriate docs were updated (not required; this aligns serialization with the existing protocol contract).

Validation:

- Route suites: 40 passed
- Full suite: 1974 passed, 90 skipped, 3 xfailed, 1 xpassed
- Coverage suite: 93% total coverage (88% required)
- `./scripts/lint.sh` (Ruff, format, and ty)

Fixes #1211 🦕

### DIFF
--- a/src/a2a/server/routes/common.py
+++ b/src/a2a/server/routes/common.py
@@ -1,6 +1,10 @@
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any
 
+from google.protobuf.json_format import MessageToDict
+
+from a2a.types.a2a_pb2 import ListTasksResponse
+
 
 if TYPE_CHECKING:
     from starlette.authentication import BaseUser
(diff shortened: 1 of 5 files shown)


mykytanetipa (approved):
LGTM

PR #1216, opened 2026-08-27
fix: flush the last SSE event when the stream ends without a blank line

### LINKED ISSUE #1213: [Bug]: SSE client drops the last event when the stream ends without a blank line
### What happened?

`parse_sse_stream` in `src/a2a/client/transports/http_helpers.py` only yields an event when it sees a blank line. If the peer closes the connection after `data: {...}` with no trailing blank line, leftover `payload_chunks` are thrown away.

WHATWG SSE says a pending event must be dispatched when the connection closes. `JsonRpcTransport` and `RestTransport` both use this parser, so a completed Task/Message can vanish and the client sits there as if the stream is still open.

**Repro**
1. Feed `parse_sse_stream` a response whose last lines are `data: {"task":{...}}` with no following empty line.
2. Exhaust the generator.

Observed: zero events.
Expected: one `("message", "{...}")` event.

Existing tests always send a trailing `\n`, so this path is untested.

### Relevant log output

n/a. Unit-reproducible in `tests/client/transports/test_http_helpers.py`.

### Code of Conduct

anxkhn:
# Description

- [x] Follow the CONTRIBUTING Guide
- [x] Conventional Commits title
- [x] Tests and linter pass
- [x] Docs updated if necessary

Fixes #1213

## Summary

`parse_sse_stream` only yields on a blank line. If the peer closes after `data: {...}` with no trailing blank line, leftover payload is dropped.

WHATWG SSE requires dispatching a pending event when the connection closes. JsonRpcTransport and RestTransport both use this parser, so a completed Task or Message can vanish and the client waits as if the stream is still open.

After the `aiter_lines` loop, if `payload_chunks` is non-empty, the pending event is now yielded the same way a blank line would. Existing tests already send a trailing newline and stay green.

## Testing

Added `test_parse_sse_stream_flushes_last_event_without_blank_line` in `tests/client/transports/test_http_helpers.py`.

```
uv run pytest tests/client/transports/test_http_helpers.py -q
```

4 passed.

### DIFF
--- a/src/a2a/client/transports/http_helpers.py
+++ b/src/a2a/client/transports/http_helpers.py
@@ -102,6 +102,9 @@ async def parse_sse_stream(
         elif key == 'data':
             payload_chunks.append(val)
 
+    if payload_chunks:
+        yield event_name, '\n'.join(payload_chunks)
+
 
 async def send_http_stream_request(
     httpx_client: httpx.AsyncClient,
--- a/tests/client/transports/test_http_helpers.py
+++ b/tests/client/transports/test_http_helpers.py
@@ -39,6 +39,18 @@ async def test_parse_sse_stream_edge_cases():
 
 
 @pytest.mark.asyncio
+async def test_parse_sse_stream_flushes_last_event_without_blank_line():
+    async def mock_aiter_lines():
+        yield 'data: {"task":{"id":"t1"}}\n'
+
+    response = httpx.Response(200)
+    response.aiter_lines = mock_aiter_lines  # type: ignore
+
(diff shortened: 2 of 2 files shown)


mykytanetipa (changes requested):
@anxkhn The rationale is inverted. WHATWG SSE (9.2.6, https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) requires the opposite: "Once the end of the file is reached, any pending data must be discarded. (If the file ends in the middle of an event, before the final empty line, the incomplete event is not dispatched.)"

So the parser is already correct, the client should discard an unterminated trailing event.

PR #1218, opened 2026-08-27
fix: copy in-memory push configs so callers cannot mutate stored webhooks

### LINKED ISSUE #1215: [Bug]: In-memory push store keeps the caller object so later mutations rewrite stored webhooks
### What happened?

`InMemoryPushNotificationConfigStore.set_info` appends the caller proto and `get_info` / `get_info_for_dispatch` return those same objects.

After create/get, changing `url`/`token`/`id` on the request or response proto silently changes the stored webhook. The next `send_notification` POSTs to the mutated URL.

`DatabasePushNotificationConfigStore` already `CopyFrom`s. `InMemoryTaskStore` wraps `CopyingTaskStoreAdapter` for the same reason. The JS SDK had this class of bug and cloned on save.

**Repro**
```python
cfg = TaskPushNotificationConfig(url="http://a.example/cb")
await store.set_info("t1", cfg, ctx)
cfg.url = "http://evil.example/cb"
got = await store.get_info("t1", ctx)
```

Observed: `got[0].url == "http://evil.example/cb"`
Expected: stored copy still `"http://a.example/cb"`

### Relevant log output

n/a.

### Code of Conduct

anxkhn:
# Description

- [x] Follow the CONTRIBUTING Guide
- [x] Conventional Commits title
- [x] Tests and linter pass
- [x] Docs updated if necessary

Fixes #1215

## Summary

`InMemoryPushNotificationConfigStore.set_info` appended the caller proto. `get_info` and `get_info_for_dispatch` returned those same objects.

After create or get, changing url, token, or id on the request or response proto silently changed the stored webhook. The next notification POST went to the mutated URL.

The SQL store already CopyFroms. InMemoryTaskStore wraps CopyingTaskStoreAdapter for the same reason. The JS SDK cloned on save for this class of bug.

set_info now CopyFroms into a new proto before assign-id and append. get_info and get_info_for_dispatch return copies. Default id assignment stays on the stored copy, not the caller.

## Testing

Added `test_set_info_copies_caller_and_returned_config`.

```
uv run pytest tests/server/tasks/test_inmemory_push_notifications.py -q
```

21 passed.

### DIFF
--- a/src/a2a/server/tasks/inmemory_push_notification_config_store.py
+++ b/src/a2a/server/tasks/inmemory_push_notification_config_store.py
@@ -12,6 +12,14 @@ from a2a.types.a2a_pb2 import TaskPushNotificationConfig
 logger = logging.getLogger(__name__)
 
 
+def _copy_config(
+    notification_config: TaskPushNotificationConfig,
+) -> TaskPushNotificationConfig:
+    stored = TaskPushNotificationConfig()
+    stored.CopyFrom(notification_config)
+    return stored
+
+
 class InMemoryPushNotificationConfigStore(PushNotificationConfigStore):
     """In-memory implementation of PushNotificationConfigStore interface.
 
@@ -48,16 +56,18 @@ class InMemoryPushNotificationConfigStore(PushNotificationConfigStore):
             if task_id not in owner_infos:
                 owner_infos[task_id] = []
 
-            if not notification_config.id:
-                notification_config.id = task_id
+            stored = TaskPushNotificationConfig()
+            stored.CopyFrom(notification_config)
+            if not stored.id:
+                stored.id = task_id
 
             # Remove existing config with the same ID
             for config in owner_infos[task_id]:
-                if config.id == notification_config.id:
+                if config.id == stored.id:
                     owner_infos[task_id].remove(config)
                     break
 
-            owner_infos[task_id].append(notification_config)
+            owner_infos[task_id].append(stored)
             logger.debug(
                 'Push notification config for task %s with config id %s for owner %s saved/updated.',
                 task_id,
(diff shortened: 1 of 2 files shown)


src/a2a/server/tasks/inmemory_push_notification_config_store.py
```diff
@@ -48,16 +56,18 @@ async def set_info(
             if task_id not in owner_infos:
                 owner_infos[task_id] = []
 
-            if not notification_config.id:
-                notification_config.id = task_id
```
mykytanetipa, line 52: this leads to spec violation for on_create_task_push_notification_config (3.1.7: "Outputs: - TaskPushNotificationConfig: Created configuration with assigned ID")

on_create_task_push_notification_config is expected to set id for TaskPushNotificationConfig response if not provided by caller.

src/a2a/server/tasks/inmemory_push_notification_config_store.py
mykytanetipa, line 74: stored.id? notification_config.id can now be empty

src/a2a/server/tasks/inmemory_push_notification_config_store.py
```diff
@@ -48,16 +56,18 @@ async def set_info(
...
+            stored = TaskPushNotificationConfig()
+            stored.CopyFrom(notification_config)
```
mykytanetipa, line 60: Consider using _copy_config for consistency

mykytanetipa (changes requested):

ConnorMoss02:
Simulated this locally against `#1236`'s tests. On its own it makes the in-memory store return a config with no `id`, because the copy means the id lands on the store's object instead of the one the handler returns. All four of those tests fail without `#1236` and pass with it.

PR #1220, opened 2026-08-29
fix(server): avoid out-of-range datetime task ordering

### LINKED ISSUE #1219: [Task]: datetime.min fallback in DatabaseTaskStore.list() is outside MySQL's documented DATETIME range
### What happened?
`DatabaseTaskStore.list()` uses `datetime.min` (year 0001) as the `COALESCE` fallback for NULL timestamps in its `ORDER BY`:
https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/database_task_store.py#L248-L256
```python
# Use coalesce to treat NULL timestamps as datetime.min,
# which sort last in descending order
stmt = base_stmt.order_by(
    func.coalesce(
        timestamp_col,
        datetime.min.replace(tzinfo=timezone.utc),
    ).desc(),
    self.task_model.id.desc(),
)
```
Per the [MySQL `DATETIME` documentation](https://dev.mysql.com/doc/refman/9.7/en/datetime.html), the supported range is `'1000-01-01 00:00:00'` to `'9999-12-31 23:59:59'`. Year 0001 is outside that range.
Raised by @lkk7 in review: [comment](https://github.com/a2aproject/a2a-python/pull/709#discussion_r3864326866)

**To be clear about severity: this is mostly a spec-conformance issue, not a replicable bug.** I could not make it fail, but still worth fixing since the MySQL [spec](https://dev.mysql.com/doc/refman/9.7/en/datetime.html) states "For the DATE and DATETIME range descriptions, “supported” means that although earlier values might work, there is no guarantee."

### Prior attempts that did NOT work — please don't reintroduce
This line has already been changed a couple of times. Recording so the next fix does not cycle back:
**`timestamp_col.desc().nulls_last()` — BROKEN on MySQL.** Removed in #651 (`fix: use MySQL compatible syntax in list tasks`) because MySQL has no `NULLS LAST` syntax.

### Relevant log output

```shell
MySQL version: 8.0.46
sql_mode:      ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION

Saved 5 tasks (2 with NULL last_updated)

Raw rows as stored in MySQL (ordered by the SDK expression):
  id     last_updated                 coalesce fallback
  t-A    2025-01-03 00:00:00          2025-01-03 00:00:00
  t-E    2025-01-02 00:00:00          2025-01-02 00:00:00
  t-C    2025-01-01 00:00:00          2025-01-01 00:00:00
  t-D    None                         0001-01-01 00:00:00
  t-B    None                         0001-01-01 00:00:00


COALESCE output JSON type: STRING

full list order: ['t-A', 't-E', 't-C', 't-D', 't-B']
  page 1: ['t-A', 't-E']
  page 2: ['t-C', 't-D']
  page 3: ['t-B']
paginated: ['t-A', 't-E', 't-C', 't-D', 't-B']

PASS: list() works on MySQL with NULL timestamps.
```

### Code of Conduct

Jokasa7:
# Description

`DatabaseTaskStore.list()` currently makes NULL timestamps sort last by
coalescing them to `datetime.min` (year 0001). MySQL documents year 1000 as the
lower bound of its supported `DATETIME` range. Current MySQL versions may still
accept year 0001, so this is a portability and specification-conformance risk
rather than a consistently reproducible runtime failure.

This change replaces the datetime sentinel with an explicit ordering key:

1. a `CASE` expression ranks non-NULL timestamps before NULL timestamps;
2. non-NULL timestamps remain in descending order; and
3. task IDs remain the descending tie-breaker, including for NULL timestamps.

This preserves the existing list and cursor-pagination order without binding an
out-of-range datetime. It also avoids `NULLS LAST`, which MySQL does not support
and which was previously removed in #651.

The regression test observes SQLAlchemy's compiled execution parameters and
fails if `list()` binds a datetime before year 1000. With the test applied to the
base implementation, the SQLite case failed by capturing the year-0001
sentinel. With this change, the same test passes on SQLite, PostgreSQL 15, and
MySQL 8.0, alongside the existing ordering and pagination matrix.

## Validation

- `./scripts/lint.sh` — Ruff check/format passed; `ty check` passed
- `./scripts/run_db_tests.sh -k test_list_tasks` — 36 passed across SQLite,
  PostgreSQL 15, and MySQL 8.0
- `uv run pytest tests/integration/cross_version/client_server/test_client_server.py -q`
  — 4 passed
- `uv run pytest` — 1971 passed, 92 skipped, 3 xfailed, 1 xpassed
- `uv run pytest --cov=src --cov-report=term-missing` — same test result,
  93% total coverage
- `git diff --check` — passed

## Checklist

- [x] Followed the `CONTRIBUTING` guide and repository AI workflow documents.
- [x] Used a Conventional Commit title.
- [x] Ran the required linter, formatter, type checker, full tests, and coverage.
- [x] Added a regression test for the changed runtime behavior.
- [x] No user-facing documentation change is required; ordering semantics are
      unchanged.

Fixes #1219 🦕

AI assistance was used for issue and code-path research and for drafting. I
reviewed the final diff and ran every validation command listed above.


mykytanetipa (approved):
LGTM

PR #1229, opened 2026-09-04
fix(server): surface producer errors after failed tasks

### LINKED ISSUE #1228: blocking on_message_send treats a FAILED Task as a successful result
## Summary

Follow-up from [@astrogilda's review](https://github.com/a2aproject/a2a-python/pull/1191#pullrequestreview-3120125585) of #1191 / #1175.

Blocking `DefaultRequestHandlerV2.on_message_send` returns successfully when it sees a `Task` in any terminal state, including `FAILED`:

```294:303:src/a2a/server/request_handlers/default_request_handler_v2.py
            if isinstance(event, Task) and (
                params.configuration.return_immediately
                or event.status.state
                in (TERMINAL_TASK_STATES | INTERRUPTED_TASK_STATES)
            ):
                self._validate_task_id_match(task_id, event.id)
                result = event
                break
```

A producer crash that also emits `TaskStatusUpdateEvent(FAILED)` (with `replace_status_update_with_task=True`) therefore becomes a **successful** blocking `message/send` instead of an `A2AClientError`. Scenarios 9 and 12 encode the current contract: the producer exception must surface.

That is why #1191 persists `FAILED` + push on producer-failure but does **not** enqueue a FAILED status event on the subscriber stream. After #1191, a streaming client that survives the exception still cannot distinguish "the producer crashed" from "the connection dropped" without a `get_task()` poll.

## Proposal

Treat a `FAILED` terminal `Task` as an error on the blocking path (or keep reading until the producer exception arrives), so a FAILED status event can be published to live streams without converting a crash into success. Until then, #1175's producer-failure half stays polling-dependent.

## Related

- #1175
- #1191

dawNotPoi:
# Description

Blocking `message/send` previously stopped consuming as soon as a FAILED Task replacement arrived. If the producer then raised, the subscriber never observed that exception and the request returned the failed Task as a successful RPC result.

This keeps a FAILED Task as the fallback response but continues consuming until the current request completes, the stream closes, or the producer error arrives. A clean agent-declared failure still returns its Task; a producer crash now surfaces its original exception. Other terminal states and `return_immediately` retain their existing behavior.

- [x] Make the Pull Request title follow the Conventional Commits specification.
- [x] Ensure the tests and linter pass.
- [x] Appropriate docs were updated (not necessary for this internal behavior fix).

## Test plan

- `./scripts/lint.sh`
- `uv run pytest` — 1,987 passed, 90 skipped, 3 xfailed, 1 xpassed
- `uv run pytest --cov=src --cov-report=term-missing` — 93% total coverage
- Focused request-handler regression matrix — 7 passed

Fixes #1228 🦕

## AI assistance

Codex was used for implementation, testing, and independent code review. This Draft remains subject to human review.

### DIFF
--- a/src/a2a/server/request_handlers/default_request_handler_v2.py
+++ b/src/a2a/server/request_handlers/default_request_handler_v2.py
@@ -38,6 +38,7 @@ from a2a.types.a2a_pb2 import (
     SubscribeToTaskRequest,
     Task,
     TaskPushNotificationConfig,
+    TaskState,
 )
 from a2a.utils.errors import (
     ExtendedAgentCardNotConfiguredError,
@@ -298,9 +299,16 @@ class DefaultRequestHandlerV2(RequestHandler):
             ):
                 self._validate_task_id_match(task_id, event.id)
                 result = event
-                # DO break here as it's "return_immediately".
-                # AgentExecutor will continue to run in the background.
-                break
+                # A FAILED task may be followed by a producer exception. Keep
+                # the task as the fallback result, but let the subscription
+                # surface that exception or finish the current request.
+                if (
+                    params.configuration.return_immediately
+                    or event.status.state != TaskState.TASK_STATE_FAILED
+                ):
+                    # AgentExecutor will continue to run in the background
+                    # when return_immediately is set.
+                    break
 
             if isinstance(event, Message):
                 result = event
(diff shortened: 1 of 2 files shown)


mykytanetipa (approved):
LGTM

PR #1230, opened 2026-09-07
feat: implement ACTS SUT behaviors for the ITK agent with YAML config

JakubWorek:
# Description

SUT for ACTS conformance tests

BEGIN_COMMIT_OVERRIDE
ci: implement ACTS SUT behaviors for the ITK agent with YAML config (#1230)
END_COMMIT_OVERRIDE

### DIFF
--- a/itk/main.py
+++ b/itk/main.py
@@ -461,6 +461,30 @@ class V10AgentExecutor(AgentExecutor):
         await task_updater.update_status(TaskState.TASK_STATE_CANCELED)
 
 
+def _capabilities() -> AgentCapabilities:
+    """What this agent advertises — everything, unless asked for less.
+
+    Four ACTS tests assert that an agent *without* a capability answers
+    `UnsupportedOperationError`, so their preconditions require the card not
+    to advertise it and they can never run against a fully capable agent. The
+    ACTS runner starts a second SUT with `ITK_ACTS_REDUCED_CAPABILITIES` set
+    to reach them, and the SDK already gates those operations on this card, so
+    publishing less is all it takes to refuse them.
+    """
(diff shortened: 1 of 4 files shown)


itk/main.py
```diff
@@ -495,10 +509,25 @@ async def main_async(http_port: int, grpc_port: int) -> None:
...
         version='1.0.0',
-        capabilities=AgentCapabilities(streaming=True),
+        # ACTS evaluates a test's `preconditions` against this card and skips
+        # when they are unmet (ACTS §12.5), so anything the agent really does
+        # has to be advertised or the matching tests silently never run.
+        capabilities=AgentCapabilities(
+            streaming=True,
+            push_notifications=True,
+            extended_agent_card=True,
+        ),
```
mykytanetipa: all capabilities are true unconditionally, but there are some tests that require certain capability to be false, for examples CORE-CAP-001, they will be skipped right now, correct?

maybe we can make the agent card configurable per test?

JakubWorek: added dynamic switching, acts now only needs double run to satisfy all tests precondition: one with all capabilities turned on and one with all off, relevant commit in acts runner: https://github.com/a2aproject/a2a-itk/pull/64/changes/dafd4469e2149640736232204fffedc7780856ac

mykytanetipa: but what if single test defines smth like streaming = True + push notifications = False it should be more than 2 runs in this case, right? 

e.g CLIENT-AUTH-001

mykytanetipa (approved):

itk/main.py
JakubWorek: this capabilities are not in the preconditions section, rather expected outcome on the wire that agent will return and tested one will parse

PR #1233, opened 2026-09-09
fix: compare in-memory task timestamps numerically

AndreyGatsuk:
# Description

Fix timestamp filtering and descending ordering in `InMemoryTaskStore.list()`. Comparing `ToJsonString()` values lexicographically reverses chronological order when fractional-second precision differs. For example, `2025-01-01T00:00:00.123Z` incorrectly passes a filter whose threshold is `2025-01-01T00:00:00.123000001Z`, despite being one nanosecond earlier. The same comparison also produces incorrectly ordered pages.

Compare integer nanoseconds using `ToNanoseconds()` for both filtering and sorting. Preserve the inclusive filter boundary, descending ID tie-breaker, and placement of tasks without timestamps after timestamped tasks.

[A2A specification §5.6.1, Timestamps](https://a2a-protocol.org/latest/specification/#561-timestamps) states: “Millisecond precision SHOULD be used where available”. It also specifies `google.protobuf.Timestamp`, whose representation contains seconds and nanoseconds, and permits omitted fractional seconds. Numeric comparison handles these precisions without discarding sub-millisecond information.

Add 15 regression cases covering seconds, milliseconds, microseconds, nanoseconds, equal timestamps, second boundaries, pagination, ID tie-breaking, absent timestamps/status, and timestamps before/at the Unix epoch. Before the fix, 11 cases fail and four pass; after the fix, all 35 in-memory task store tests pass.

Validation on Python 3.10 with PostgreSQL 15 and MySQL 8:

- `./scripts/lint.sh`: Ruff check/format and ty passed (three existing unused-ignore warnings outside this change).
- `uv run pytest`: 2,092 passed, 3 xfailed, 1 xpassed. The xfail/xpass cases are already marked for #869.
- `uv run pytest --cov=src --cov-report=term-missing --cov-fail-under=88`: same test results, 93.17% coverage.
- JSCPD: 2.61% duplication, below the 3% threshold.
- Built the wheel and ran `tests.install_smoke` in six clean environments: base, http-server, fastapi, grpc, telemetry, and sql all passed.

Local transport checks used proxy exclusions for loopback addresses. Cross-version `uv --with` environments were prepared before the full runs to keep dependency downloads outside server-startup timeouts.

- [x] Follow the CONTRIBUTING guide: feature branch based on current upstream main.
- [x] Conventional Commit title: `fix: compare in-memory task timestamps numerically`.
- [x] Tests and linters pass.
- [x] Protocol reference included above; no public API changes.

### DIFF
--- a/src/a2a/server/tasks/inmemory_task_store.py
+++ b/src/a2a/server/tasks/inmemory_task_store.py
@@ -93,8 +93,8 @@ class _InMemoryTaskStoreImpl(TaskStore):
                 task for task in tasks if task.status.state == params.status
             ]
         if params.HasField('status_timestamp_after'):
-            last_updated_after_iso = (
-                params.status_timestamp_after.ToJsonString()
+            last_updated_after_ns = (
+                params.status_timestamp_after.ToNanoseconds()
             )
             tasks = [
                 task
@@ -102,8 +102,8 @@ class _InMemoryTaskStoreImpl(TaskStore):
                 if (
                     task.HasField('status')
                     and task.status.HasField('timestamp')
-                    and task.status.timestamp.ToJsonString()
-                    >= last_updated_after_iso
+                    and task.status.timestamp.ToNanoseconds()
+                    >= last_updated_after_ns
                 )
             ]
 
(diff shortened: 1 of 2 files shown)


ConnorMoss02:
The `>=` matches the proto, which says tasks with a status timestamp "greater than or equal to this value" (`a2a.proto:696`). It looks like the SDKs disagree on that boundary: `a2a-go` is inclusive too (`inmemory.go:245`, via `!Before()`), but `a2a-js` uses `>` at `store.ts:91`, so a task whose timestamp exactly equals the filter is returned here and dropped there.

mykytanetipa (approved):
LGTM

PR #1243, opened 2026-09-16
ci: add A2A protocol authentication support and integrate ACTS conformance testing CI workflow

JakubWorek:
# Description

Followup after https://github.com/a2aproject/a2a-python/pull/1230

Required changes to auth testing with updated corpus.
Ci workflows for PR and nightly ACTS validation - mirrors ITK pipelines.
PR workflow is not blocking until sdk would be conformant.


itk/main.py
```diff
@@ -461,6 +472,170 @@ async def cancel(
...
+#: `SEC-AUTH-002` and `SEC-EXTCARD-002`, so a fixture has to recognise both to
+#: answer 200 / 403 / 401 as those tests require.
+ACTS_VALID_TOKEN = 'itk-valid-token'  # noqa: S105
+ACTS_INSUFFICIENT_TOKEN = 'itk-insufficient-token'  # noqa: S105
+ACTS_SECURITY_SCHEME = 'bearerAuth'
+
+#: Where the REST binding serves the extended card, relative to its mount.
+#: Matched as a suffix because `create_rest_routes` also mounts every route
+#: under `/{tenant}`, so the same operation answers on two paths.
+EXTENDED_CARD_PATH = '/extendedAgentCard'
```
mykytanetipa, line 486: why only REST validates this? no JSON-RPC, gRPC auth support?

mykytanetipa (approved):

itk/main.py
JakubWorek: SEC-EXTCARD-001/002/004 are all transport: [rest] with raw GET /extendedAgentCard, because a raw JSON-RPC/gRPC step would be byte-identical to any other unauthenticated call on the RPC endpoint. The fixture's REST-only guard matches what ACTS can verify.

PR #1259, opened 2026-09-20
fix(client): ignore the leading BOM in SSE streams

YaoxinHuang:
# Description

- [x] Follow the [CONTRIBUTING Guide](https://github.com/a2aproject/a2a-python/blob/main/CONTRIBUTING.md).
- [x] Use a Conventional Commits title.
- [x] Ensure the tests and linter pass.
- [x] Appropriate docs updated if necessary; no API change.

An SSE stream beginning with a UTF-8 BOM currently loses its first `data` line, or treats an initial `event: error` as a normal message. Ignore exactly one leading BOM, as required by the [SSE parsing rules](https://html.spec.whatwg.org/multipage/server-sent-events.html#interpreting-an-event-stream), while preserving BOM characters elsewhere.

Added byte-response regressions for initial data/error fields, repeated BOMs, and BOM characters inside or later in the stream. Transport tests, the unit suite, Ruff and type checks pass locally. Full-suite verification remains incomplete on Windows because cross-version integration subprocesses time out and their cleanup uses `os.killpg`, which is unavailable on Windows; database-service cases were skipped without configured services.


src/a2a/client/transports/http_helpers.py
```diff
@@ -72,9 +72,13 @@ async def parse_sse_stream(
     """
     event_name = 'message'
     payload_chunks: list[str] = []
+    first_line = True
 
     async for line in response.aiter_lines():
         raw_line = line.rstrip('\r\n')
+        if first_line:
+            raw_line = raw_line.removeprefix('\ufeff')
```
ishymko, line 80: nit: how about 

```suggestion
            raw_line = raw_line.removeprefix('\N{BYTE ORDER MARK}')
```

to use [`\N{name}`](https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences) syntax for Named Unicode characters?

UPD: likely this will also eliminate spelling check update for `ufeff` :)

ishymko (commented):
Thank you! Just one nit to consider for readability.

